PR #59, opened 2017-07-12
Add mirrorer

deanwilson:
This needs to be tested without an ELB being passed to the terraform module as mirrorer doesn't need inbound connections.

EDIT: I (Laura) have refactored this and tested it. It seems to work OK.


terraform/projects/govuk-security-groups/mirrorer.tf
```diff
@@ -0,0 +1,21 @@
...
+# === Variables:
+# stackname - string
+#
+# === Outputs:
+# sg_mirrorer_id
+
+resource "aws_security_group" "mirrorer" {
+  name        = "${var.stackname}_mirrorer_access"
+  vpc_id      = "${data.terraform_remote_state.govuk_vpc.vpc_id}"
+  description = "Access to the mirrorer host from its ELB"
```
SamLR: This description is in-accurate

terraform/projects/govuk-security-groups/outputs.tf
```diff
@@ -0,0 +1,79 @@
+#
+# Outputs of the security group project
+#
```
SamLR: The changes for mirrorer need to be moved into `infra-security-groups`

surminus:
@SamLR thanks, updated!

PR #79, opened 2017-07-17
Add backup machine.

deanwilson:
This will need to be redone after the refactor.


SamLR:
I've updated this to the new system and it seems to work ok.

terraform/projects/app-backup/main.tf
```diff
@@ -0,0 +1,148 @@
...
+}
+
+variable "ssh_public_key" {
+  type        = "string"
+  description = "Default public key material"
+}
+
+variable "backup_subnet" {
+  type        = "string"
+  description = "Name of the subnet to place the Graphite instance 1 and EBS volume"
```
surminus: Description is incorrect

PR #87, opened 2017-07-19
Add an ADR for the internal DNS

deanwilson:


afda16:
We have another 2 DNS docs ...

deanwilson:
We've discussed it and we'll add this rather than rewrite the existing as it's quite focused. There will be another document showing an overview that will link to the relevant decisions.

PR #90, opened 2017-07-19
Add the backend machines to the terraform configs

deanwilson:


afda16:
`terraform/projects/app-backend/issam.backend` should be `terraform/projects/app-backend/integration.issam.backend`

deanwilson:
I've renamed the backend file, thanks.

PR #92, opened 2017-07-19
Add publishing api machines

deanwilson:


terraform/projects/app-publishing-api/issam.backend
```diff
@@ -0,0 +1,4 @@
+bucket  = "govuk-terraform-steppingstone-integration"
```
SamLR: As discussed I'll clear these bringing up the issam stack

PR #116, opened 2017-07-20
Update DNS infrastructure ADR

afda16:
Update ADR to reflect the difference between a zone and a domain. We have
internal and external domains for each stack, external zones for each stack
(with a root zone delegating the stack subdomain to those zones), but a single root
internal zone, because we can't do delegation with internal domains.

Add another image to make the implementation easier to understand.


doc/architecture/decisions/0015-dns-infrastructure.md
```diff
@@ -21,23 +21,27 @@ to a different stack, execute green-blue application deployments, etc.
...
+instance, machines are using the Puppet service in `puppet.<internalrootdomain>` that is a 
+CNAME of `puppet.deana.<internalrootdomain>`. At some point, a new Puppet stack is provisioned to test a new version,
+and when it has been tested we switch the CNAME to the new stack, so `puppet.<internalrootdomain>`
+resolves to `puppet.newstack.<internalrootdomain>`. The same approach can be used for application stacks, so
 green-blue deployments can be implemented in this infrastructure.
 
 The root domain DNS records don't need to be managed by Terraform. We can implement lightweight scripts inserted
 in the Deployment process to switch active versions between stacks.
 
+We can't delegate internal subdomains overlapping domain names, so there will only be an internal DNS zone per
```
SamLR: Can you add an example to explain "overlapping domain names", please?

SamLR: Also I think the second part may be clearer as "so there will only be one internal DNS zone per VPC, which is shared between stacks"

PR #119, opened 2017-07-20
Add apt 1

SamLR:
This adds the apt box


terraform/projects/infra-security-groups/outputs.tf
```diff
@@ -68,6 +80,7 @@ output "sg_calculators-frontend_id" {
 
 output "sg_content-store_elb_id" {
   value = "${aws_security_group.content-store_elb.id}"
+
```
surminus: Adding an erroneous line

PR #122, opened 2017-07-21
Add machines for search

issyl0:


issyl0:
Thanks. Is the Puppet expected to fail?

PR #124, opened 2017-07-21
Add exception handler

SamLR:


afda16:
I updated the attach-volume snippet in https://github.com/alphagov/govuk-puppet/pull/6143 to work with the device name, that should make the snippet generic

terraform/projects/app-exception-handler/main.tf
```diff
@@ -0,0 +1,210 @@
...
+  type    = "A"
+
+  alias {
+    name                   = "${aws_elb.exception_handler_internal_elb.dns_name}"
+    zone_id                = "${aws_elb.exception_handler_internal_elb.zone_id}"
+    evaluate_target_health = true
+  }
+}
+
+module "exception_handler-1" {
```
surminus: this should just be the puppet class and with a hyphen (copying previous apps), so "exception-handler".

PR #125, opened 2017-07-21
Update volume snippet and logs-elasticsearch

afda16:
Update the attach-volume snippet to use a device tag so we can
attach several volumes to an instance and continue doing the
LVM management with Puppet.

The EBS TF resource needs to include a 'Device' tag so we can attach
the volume to the right device during the instance initialisation.

Puppet already manages the devices and LVM configuration, but we'll
need to update the lvm module in Puppet to use the puppetlabs version
instead of our fork, as it fixes several bugs that prevented a disk
already formatted from working properly when a new instance is running
Puppet the first time.


afda16:
The Puppet 'lvm' module has been updated manually on delana.puppetmaster to use the latest puppetlabs-lvm.

Check https://github.com/alphagov/govuk-puppet/pull/6143

PR #132, opened 2017-07-25
Remove un-needed whitespace and double quotes

SamLR:
Many tfvars files have a leading empty line or variable names surrounded
by double quotes, these aren't needed so remove them.


issyl0:
❤️

PR #133, opened 2017-07-25
Add terraform tests to CI

surminus:
Adding these tests ensure our code is formatted correctly and has correct syntax.

We need to specify the node to ensure we get the correct version of Terraform.

Depends on https://github.com/alphagov/govuk-puppet/pull/6155

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@
 
 REPOSITORY = 'govuk-aws'
 
-node {
+node ("terraform") {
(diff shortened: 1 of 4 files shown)


issyl0:
Isn't this what the pre-commit hooks are for?

surminus:
Yes, but it's also a good safety net to ensure nothing slips through.

surminus:
The test has already found an issue from someone else's work: https://ci.integration.publishing.service.gov.uk/job/govuk-aws/job/add-terraform-tests-to-ci/2/console

PR #134, opened 2017-07-26
[DISCUSS] Rough sketch to removing old nodes

surminus:
This allows the Puppetmaster to query what instances are running in the estate (and other things that we should discuss).

It is possible to query the nodes in puppetdb using `govuk_node_list`, although this does require a manual DNS entry in `/etc/hosts` (in the future this could be done properly).

The idea is to compare what is in puppetdb and what the API states in running in the stack, and to remove a node if it doesn't report back a match from the API.

I do not think this should be the long term way to remove nodes, but it is a quick and dirty way to complete the task while we're testing.

I started thinking about this as I made changes to Icinga configuration. Puppet gets extremely upset with Icinga if nodes do not exist, and so I needed to deactivate some old ones.

This is the script I wrote to run every now and then to delete old ones:
```
#!/bin/bash
#
# Simple script to deactivate old nodes in PuppetDB
#
for node in $(govuk_node_list); do
  if [[ $(aws ec2 --region=eu-west-1 describe-instances --filters "Name=tag:aws_stackname,Values=delana" | jq -r ' .Reservations[] | .Instances[] | .PrivateDnsName' |grep $node) == "" ]]; then
    govuk_node_clean $node
  fi
done
```

In practice this could be run every 30 minutes which would ensure nodes are taken out of monitoring etc, but I think there are better ways to do this in Production using lifecycle events.

### DIFF
--- a/terraform/projects/app-puppetmaster/additional_policy.json
+++ b/terraform/projects/app-puppetmaster/additional_policy.json
@@ -1,0 +1,15 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Sid": "Stmt1499854881000",
+            "Effect": "Allow",
+            "Action": [
+                "ec2:DescribeInstances"
+            ],
+            "Resource": [
+                "*"
+            ]
+        }
+    ]
+}
--- a/terraform/projects/app-puppetmaster/main.tf
+++ b/terraform/projects/app-puppetmaster/main.tf
@@ -140,6 +140,20 @@ module "puppetmaster" {
   instance_elb_ids              = ["${aws_elb.puppetmaster_bootstrap_elb.id}", "${aws_elb.puppetmaster_internal_elb.id}"]
 }
 
+# TODO: This will allow Puppetmaster to describe instances so that
+# we can run a script as an adhoc task that will remove nodes from
+# puppetdb that do not exist
+resource "aws_iam_policy" "puppetmaster_iam_policy" {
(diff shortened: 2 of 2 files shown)


SamLR:
I think we can probably (fairly easily) make a simple cloudwatch/lambda script to do this for us. [This blog](https://aws.amazon.com/blogs/apn/automatically-delete-terminated-instances-in-chef-server-with-aws-lambda-and-cloudwatch-events/) covers a general solution, it doesn't feel like this should be too hard to set up

surminus:
Yup, that would be good for adding new nodes, too.

afda16:
Isn't 'DescribeInstances' part of the default policy for all nodes?

surminus:
Yeah, I did this before we made it the default.

surminus:
This will no longer work as an adhoc script to run due to https://github.com/alphagov/govuk-puppet/pull/6200. We will have to write the specific query for puppetdb into an adhoc script.

surminus:
Updated the script so this will work for now:
```
#!/bin/bash
#
# Simple script to deactivate old nodes in PuppetDB
#
for node in $(govuk_node_list -C "base"); do
  if [[ $(govuk_node_list |grep $node) == "" ]]; then
    govuk_node_clean $node
  fi
done
```

This needs to run on the puppetmaster, and "puppetdb.cluster" needs to exist in `/etc/hosts` for `localhost`.

surminus:
Closing this because other work has been done to automate this process.

PR #143, opened 2017-07-28
Add ElasticSearch machines

issyl0:


terraform/projects/app-elasticsearch/additional_policy.json
```diff
@@ -0,0 +1,19 @@
...
+    "Statement": [
+        {
+            "Sid": "Stmt1499854881000",
+            "Effect": "Allow",
+            "Action": [
+                "ec2:AttachVolume",
+                "ec2:DetachVolume",
+                "ec2:DescribeVolumeStatus",
+                "ec2:DescribeVolumes",
+                "ec2:DescribeInstances"
```
surminus: This is now a default policy, sorry!

terraform/projects/app-elasticsearch/main.tf
```diff
@@ -0,0 +1,248 @@
...
+}
+
+variable "elasticsearch_2_subnet" {
+  type        = "string"
+  description = "Name of the subnet to place the  Elasticsearch 2 and EBS volume"
+}
+
+variable "elasticsearch_3_subnet" {
+  type        = "string"
+  description = "Name of the subnet to place the Graphite instance 3 and EBS volume"
```
surminus: This is incorrectly described (probably in logs-elasticsearch, too)

terraform/projects/app-elasticsearch/additional_policy.json
issyl0: Removed!

terraform/projects/app-elasticsearch/main.tf
issyl0: I think we need to go through everything - there must be so much that's mistakenly copy/pasted.

issyl0: As an example: https://github.com/alphagov/govuk-aws/blob/d9db0d750bf194b8f2be2f3864889ab40eb41494/terraform/projects/app-elasticsearch/main.tf#L3, and that's also in `logs-elasticsearch` 'cause I copied that directory.

PR #144, opened 2017-07-28
Add JSON check tool

surminus:
Check that JSON is valid using `jq`. This is a lightweight way to ensure that any IAM policies that are being submitted are valid JSON which will hopefully save a little time or pain.


surminus:
Jenkins doesn't like this, will come back to it.

PR #150, opened 2017-08-01
Add rabbitmq

surminus:
This is WIP as we're looking at doing some magic with Puppet to automatically create a cluster in an ASG.


afda16:
We have the cluster working, we might need to open more ports on the ELB but for now this should be enough

PR #156, opened 2017-08-02
Create the logs_cdn machine

issyl0:


afda16:
I think the ELB security group needs to allow access from Fastly IPs to rsyslog ports, as configured in govuk-provisioning, instead of 443

terraform/projects/app-logs_cdn/main.tf
```diff
@@ -0,0 +1,117 @@
...
+  instance_security_group_ids   = ["${data.terraform_remote_state.infra_security_groups.sg_logs-cdn_id}", "${data.terraform_remote_state.infra_security_groups.sg_management_id}"]
+  instance_type                 = "t2.micro"
+  create_instance_key           = true
+  instance_key_name             = "${var.stackname}-logs-cdn"
+  instance_public_key           = "${var.ssh_public_key}"
+  instance_additional_user_data = "${join("\n", null_resource.user_data.*.triggers.snippet)}"
+  instance_elb_ids              = ["${aws_elb.logs-cdn_external_elb.id}"]
+  asg_max_size                  = "2"
+  asg_min_size                  = "2"
+  asg_desired_capacity          = "2"
```
surminus: we should only ever have one of these

surminus:
this should mount a persistent volume; it's important we keep a history of our CDN logs

issyl0:
@surminus Yep, I was going to look into that now. Dean told me to copy Bouncer's config because it had access from Fastly and external IPs in it, hence why this was a bit wrong for the `cdn-logs` case.

terraform/projects/app-logs-cdn/main.tf
```diff
@@ -0,0 +1,138 @@
...
+}
+
+resource "aws_elb" "logs-cdn_external_elb" {
+  name            = "${var.stackname}-logs-cdn"
+  subnets         = ["${data.terraform_remote_state.infra_networking.public_subnet_ids}"]
+  security_groups = ["${data.terraform_remote_state.infra_security_groups.sg_offsite_ssh_id}"]
+  internal        = "false"
+
+  listener {
+    instance_port     = "443"
```
afda16: need to change this to the same port range that we are allowing from the security groups: 6514-6516

terraform/projects/app-logs-cdn/main.tf
```diff
@@ -0,0 +1,138 @@
...
+  size              = 500
+  type              = "gp2"
+
+  tags {
+    Name            = "${var.stackname}-logs-cdn"
+    Project         = "${var.stackname}"
+    aws_stackname   = "${var.stackname}"
+    aws_environment = "${var.aws_environment}"
+    aws_migration   = "logs_cdn"
+    aws_hostname    = "logs-cdn"
```
afda16: It needs to be the same hostname that we set in the instance tag

surminus:
The IAM policy needs to be attached.

issyl0:
@surminus I added an IAM policy. 

When I know I've done everything correctly, I'll squash the commits.

terraform/projects/app-logs-cdn/main.tf
```diff
@@ -0,0 +1,163 @@
...
+  type        = "string"
+  description = "AWS Environment"
+}
+
+variable "ssh_public_key" {
+  type        = "string"
+  description = "logs-cdn default public key material"
+}
+
+variable "logs_cdn_subnet" {
```
surminus: this and the outputs need to be put at the headers at the top of this file

PR #163, opened 2017-08-03
Add delana backends

deanwilson:


terraform/projects/app-backend-lb/integration.delana.backend
```diff
@@ -0,0 +1,4 @@
+bucket  = "govuk-terraform-steppingstone-integration"
+key     = "issam/app-backend-lb.tfstate"
```
afda16: Update the key to delana

PR #167, opened 2017-08-07
Change the required Terraform version from 0.9.10 to 0.10

deanwilson:
We're attempting to stay current.

### DIFF
--- a/tools/terraform-validate.sh
+++ b/tools/terraform-validate.sh
@@ -6,5 +6,8 @@ for file in "$@"; do
 done \
   | sort | uniq | \
     while read -r dir; do
+      rm -rf ${dir}/.terraform;
+      rm -f ${dir}/terraform.tfstate.backup;
+      terraform init "${dir}";
       terraform validate "${dir}";
     done
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,3 +15,9 @@
       entry: tools/json-check.sh
       language: script
       files: \.json$
+    - id: terraform-version-check
+      name: ===> Checking consistent versions of Terraform
+      entry: tools/terraform-version-check.rb
+      language: ruby
+      files: \.tf$
+
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,10 +18,6 @@ node ("terraform") {
       sh "find . -name '*.tf' |xargs tools/terraform-format.sh"
     }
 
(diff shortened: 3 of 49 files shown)


issyl0:
I think this can come out of WIP now - I fixed the pre-commit `terraform validate` hook.

tools/terraform-validate.sh
```diff
@@ -6,5 +6,6 @@ for file in "$@"; do
 done \
   | sort | uniq | \
     while read -r dir; do
+      terraform init "${dir}";
```
surminus: We should purge any previous terraform config, so:
```
rm -rf ${dir}/.terraform
rm -f ${dir}/terraform.tfstate.backup
```

surminus: Also, does an init require exporting of credentials?

If so, then this may break Jenkins tests.

issyl0: As discussed, I removed the pre-commit hook trigger from the Jenkinsfile (context in commit message).

surminus:
I wonder if we should now update to 0.10.4 🤔

issyl0:
Did we decide what to do with this?

surminus:
We should push to the newest version and test.

issyl0:
I've updated to `0.10.6` and included the new provider versions for Fastly and AWS. Tested with a `plan` on a couple of projects, and things seem to work. This is ready for a re-review.

issyl0:
I've updated to `0.10.6` and included the new provider versions for Fastly and AWS. Tested with a `plan` on a couple of projects, and things seem to work. This is ready for a re-review.

surminus:
I would almost be tempted to set up some tests for these that check if one module is out of sync with the others. 

I've seen a fair few copy-pasta fails, and a test seems like a good thing here.

issyl0:
@surminus I added a test for the Terraform version. Would you like providers too?

surminus:
Great, that seems to do the job! I already got some failures.

Would be good to add to the Jenkinsfile or pre-commit hooks. 

I'm still a little wary of not having `terraform validate` being put into CI, but I think the need to upgrade to a later version is greater.

surminus:
@issyl0 if you're happy to rebase and add your tests to CI or pre-commit, then I'd be happy to approve and merge before we get even more out of date

issyl0:
We've got 0.10.7 now, so I'll do this dance again and rebase and then we can merge.

PR #168, opened 2017-08-07
Update app-api-postgres

surminus:
The only database that is currently on api-postgresql database machines is Stagecraft (performance platform). We should combine our PostgreSQL instances so we only have to manage and maintain a single set.

Also update the name slightly to "postgresql" so it is more reflective of the names we use across our Puppet code base (rather than the slang "postgres").


doc/architecture/decisions/0020-merge-api-postgresql-instance-into-main-postgresql-instance.md
```diff
@@ -0,0 +1,29 @@
...
+This is able to run on the main PostgreSQL instance as we do not have the split
+of "vDCs" that we have in our current hosting provider.
+
+Work is ongoing to migrate the Performance Platform away from our infrastructure
+and to The Government PaaS, so the long term plan is not have this database in
+our infrastructure at all.
+
+## Decision
+
+We should merge the API PostgreSQl instance into the main PostgreSQL instance.
```
deanwilson: PostgreSQl should be PostgreSQL

doc/architecture/decisions/0020-merge-api-postgresql-instance-into-main-postgresql-instance.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Consequences
+
+The app will need to be updated to allow speaking to the correct database.
```
deanwilson: "The application configuration will need to be updated to connect to the correct database."

terraform/projects/app-postgresql/main.tf
```diff
@@ -69,36 +69,36 @@ module "api-postgres-primary_rds_instance" {
   allocated_storage  = "30"
   instance_class     = "db.t1.micro"
```
deanwilson: Can we bump this to a `db.m4.large` please? 8GB would be better in integration.

terraform/projects/app-postgresql/main.tf
```diff
@@ -69,36 +69,36 @@ module "api-postgres-primary_rds_instance" {
...
   multi_az           = "${var.multi_az}"
```
deanwilson: Can we bump this to a db.m4.large please? 8GB would be better in integration.

PR #169, opened 2017-08-07
Add an external ELB to the app-backend-lb

deanwilson:


terraform/projects/app-backend-lb/main.tf
```diff
@@ -86,6 +86,48 @@ resource "aws_route53_record" "backend-lb_service_record" {
...
+  cross_zone_load_balancing   = true
+  idle_timeout                = 400
+  connection_draining         = true
+  connection_draining_timeout = 400
+
+  tags = "${map("Name", "${var.stackname}-backend-lb_external", "Project", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "backend_lb")}"
+}
+
+resource "aws_route53_record" "backend-lb_external_service_record" {
+  zone_id = "${data.terraform_remote_state.infra_stack_dns_zones.internal_zone_id}"
```
afda16: If it's external, it should use external_zone_id and external_domain_name

terraform/projects/infra-security-groups/backend-lb.tf
```diff
@@ -34,6 +35,19 @@ resource "aws_security_group_rule" "allow_backend-lb_elb_in" {
...
+  type      = "ingress"
+  from_port = 443
+  to_port   = 443
+  protocol  = "tcp"
+
+  # Which security group is the rule assigned to
+  security_group_id = "${aws_security_group.backend-lb.id}"
+
+  # Which security group can use this rule
+  source_security_group_id = "${aws_security_group.backend-lb_external_elb.id}"
```
afda16: If it's external, shouldn't we allowed from office Ips or something like that?

PR #182, opened 2017-08-08
Purge terraform dir on init

surminus:
This ensures that you don't accidentally copy local statefiles from other stacks.

There's an argument that a purge and init should happen for every plan and apply for safety.


tools/build-terraform-project.sh
```diff
@@ -103,6 +103,7 @@ cd "$PROJECT_DIR"
 
 # Actually run the command
 if [[ $CMD == "init" ]]; then
+  rm -rf .terraform && \
```
SamLR: Would it be worth clearing out any 'terraform.tfstate.backup' files as well?

surminus: Good point

PR #184, opened 2017-08-08
Add default mongo subnet settings for the integration stacks

deanwilson:


terraform/data/app-router-backend/integration/blue.tfvars
```diff
@@ -1,3 +1,3 @@
-router_backend_1_subnet = "govuk_private_a"
-router_backend_2_subnet = "govuk_private_b"
-router_backend_3_subnet = "govuk_private_c"
+router-backend_1_subnet = "govuk_private_a"
```
surminus: are these required if they're in common?

PR #197, opened 2017-08-10
Add artefact bucket

SamLR:
This adds a project to build an artefact bucket. This is intended to
be written to by CI in Carrenza and read from by the deploy machines.
This means that we will not need to give AWS deploy machines access to
Carrenza (it would also simplify any bootstrapping process).

### DIFF
--- a/terraform/projects/app-cache/main.tf
+++ b/terraform/projects/app-cache/main.tf
@@ -67,11 +67,29 @@ variable "asg_desired_capacity" {
   default     = "2"
 }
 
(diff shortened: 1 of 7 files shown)


surminus:
Should we call this `infra-artefact-bucket` to keep it consistent with our naming scheme?

Are we going to have a bucket per stack?

SamLR:
I used govuk as a prefix to match our other stuff we currently have but I no strong preference either way. I think we probably only want one bucket per environment. If we wanted to use deploys for e.g. blue/green from it I think using the artefact tag is going to be preferable.

PR #210, opened 2017-08-15
Add certs to publishing-api

surminus:
This acts like a backend machine, so needs certificates.

Currently only listening on HTTP so LB will show as down, so apps are unable to speak to it.


deanwilson:
`    target              = "TCP:443"` on the ELB needs to be `80` now.

surminus:
@deanwilson updated

PR #213, opened 2017-08-16
[DISCUSS] Allow router-backend to cache

surminus:
router-api needs to communicate with the cache instances to reload routes.

This needs discussion because router needs to be able reload routes on each router instance, and if it goes through a load balancer then it will only do it on a single instance.

I suspect we may need to recode how we set the environment variable for "ROUTER_NODES" to look up router instances based upon tags.


surminus:
We need a way to populate the environment variable rather than use a LB. Closing.

PR #215, opened 2017-08-16
[WIP] Assign an external ELB to backend

deanwilson:
This is the first part of removing the backend lb host tier


afda16:
Already implemented in https://github.com/alphagov/govuk-aws/pull/269

PR #216, opened 2017-08-16
First draft of the new lb removal plans

deanwilson:


surminus:
A consequence to note will be moving the GOVUK-request-ID header thing, as this won't be possible to emulate on an ELB.

boffbowsh:
This makes sense to me. We need to ensure that we move the generation of the request id header to within the applications though.

surminus:
I'm already in the midst of removing frontend-lb, and the request ID header is set on the cache machines I think so no biggie here.

boffbowsh:
👍

PR #220, opened 2017-08-17
Add an external cache load balancer and allow https traffic to it.

deanwilson:


surminus:
We should also add a CNAME for `www-origin` like we have done [here](https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/app-backend-lb/main.tf#L152-L159)

terraform/data/app-cache/integration/common.tfvars
```diff
@@ -0,0 +1,6 @@
+"app_service_records" = [
+  "assets-origin",
+  "draft-origin"
```
surminus: missing a comma

deanwilson: The last entry in the existing data files does not have a comma. So it's either not allowed or I'm being consistent. Pick an excuse ;)

surminus: That's not the last entry though

PR #222, opened 2017-08-17
Allow ssh from carrenza integration

SamLR:
This allows SSH from the carrenza integration environment for the
purpose of syncing data


terraform/data/common/integration/common.tfvars
```diff
@@ -18,6 +18,21 @@ office_ips = [
   "213.86.153.237/32",
 ]
 
+carrenza_integration_ips = [
+  "31.210.245.98/32",
```
surminus: you could probably just use 31.210.245.96/28

PR #226, opened 2017-08-18
Add SGs to enable access to graphite:443 from monitoring

afda16:
Monitoring needs access to graphite:443 internally


afda16:
We need the internal certificate for the Graphite API to work, otherwise the client gets certificate name error

PR #228, opened 2017-08-18
Add an ADR about splitting out the data dir

SamLR:


doc/architecture/decisions/0023-use-separate-data-repository.md
```diff
@@ -0,0 +1,26 @@
...
+Date: 2017-08-18
+
+## Status
+
+Accepted
+
+## Context
+
+We have a large amount of data which is needed to run the Terraform projects.
+
```
deanwilson: Can we add a sentence or two about why we want data separate from code and how creds being public is "bad, M'kay" please? The context is a little thin atm.

PR #241, opened 2017-08-30
Add flow logs for all failed vpc connections

deanwilson:


terraform/projects/infra-flow-logs/main.tf
```diff
@@ -0,0 +1,104 @@
...
+        "Service": "vpc-flow-logs.amazonaws.com"
+      },
+      "Action": "sts:AssumeRole"
+    }
+  ]
+}
+EOF
+}
+
+resource "aws_iam_role_policy" "vpc_flow_logs_policy" {
```
SamLR: Can this use a [policy doc data resource](https://www.terraform.io/docs/providers/aws/d/iam_policy_document.html) so there's some amount of verification before it hits AWS?

afda16:
Does it have to be a different project? We could put it with infra-vpc, with a variable to enable logs if selected

afda16:
I updated the policy creation to use a json file, same as the rest of the code. I still think this code should be part of the infra-vpc project

afda16:
I moved the code to `infra-vpc` and removed the `infra-flow-logs` project

PR #250, opened 2017-08-31
Set a default for parameter_group_name

SamLR:
There needs to be a default otherwise terraform errors.


surminus:
Feels weird having a default of blank, but I'm not opposed if other people aren't bothered.

SamLR:
Yea it's not nice but it means that you don't have to provide a parameter group so *shrug*

PR #257, opened 2017-09-03
Allow further backend services access

surminus:
The whitehall machines and calculators-frontend needed access to some backend services. This opens up the security groups to allow them to do so.

https://trello.com/c/DS5stwo8/730-fix-frontend-flows

### DIFF
--- a/terraform/projects/infra-security-groups/mongo.tf
+++ b/terraform/projects/infra-security-groups/mongo.tf
@@ -80,6 +80,17 @@ resource "aws_security_group_rule" "allow_frontend_to_mongo_elb" {
   source_security_group_id = "${aws_security_group.frontend.id}"
 }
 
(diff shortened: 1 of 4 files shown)


terraform/projects/infra-security-groups/mongo.tf
```diff
@@ -80,6 +80,17 @@ resource "aws_security_group_rule" "allow_frontend_to_mongo_elb" {
   source_security_group_id = "${aws_security_group.frontend.id}"
 }
 
+resource "aws_security_group_rule" "allow_calculators-frontend_to_mongo_elb" {
```
tijmenb: Out of interest, what's the thing on calculators-frontend that needs mongo?

surminus: licencefinder apparently

PR #258, opened 2017-09-04
Update docs

surminus:
Add two sets of docs and an ADR.


doc/guides/certificate-validation.md
```diff
@@ -0,0 +1,37 @@
+# Certificate Validation
```
afda16: This looks OK for a general guide, but if someone joins the team, that person won't be able to understand from this guide which certificates are following this. I still think we should document somewhere (maybe in a private space) our specific cases.  For instance, in which account a domain has been configured, which email address it's using, for how long we are creating certificates, etc

doc/guides/certificate-validation.md
```diff
@@ -0,0 +1,37 @@
...
+
+Create a new "Rule Set" in the "Email Receiving" section of SES.
+
+Give the rule a name, add the recipient we just created, and add the action to forward the email to an S3 bucket. If you select the S3 bucket option, it will give you the option to choose which bucket to use, or it will also offer to create the bucket on your behalf.
+
+**Be sure to enable the rule set as "active" when it has been created**
+
+### Pick up the email
+
+When the bucket and rule set has been created, send the certiicate validation email. It should appear in the bucket; open up the email, and click the link inside to validate the certificate.
```
afda16: "certificate"

doc/guides/certificate-validation.md
surminus: I agree that we need to store this information somewhere; that might be a suitable use of our wiki page or the ops manual. I'm not sure it's appropriate to have that info in the repo. 

This was mostly a braindump to help me remember what we did; we should automate it in the future.

PR #259, opened 2017-09-04
Add ADR for elasticache

surminus:
https://trello.com/c/VNID5ncy/676-adr-for-using-elasticache

### DIFF
--- a/doc/architecture/decisions/0025-use-elasticache-for-redis.md
+++ b/doc/architecture/decisions/0025-use-elasticache-for-redis.md
@@ -1,0 +1,26 @@
+# 25. Use Elasticache for Redis
+
+Date: 2017-09-04
+
+## Status
+
+Accepted
+
+## Context
+
+Traditionally we provisioned our own redis machines in a non-clustered state.
+
+AWS provide Elasticache which has a Redis engine, and can be configured to be clustered.
+
+We should consider provisioning Elasticache to replace our provisioned Redis instances.
+
+## Decision
+
+We are using Elasticache instead of provisioning our own Redis instances.
+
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0025-use-elasticache-for-redis.md
```diff
@@ -0,0 +1,24 @@
...
+
+We should consider provisioning Elasticache to replace our provisioned Redis instances.
+
+## Decision
+
+We are using Elasticache instead of provisioning our own Redis instances.
+
+## Consequences
+
+We have less Puppet code to manage. There is no risk of lock-in because the client side
```
SamLR: Are there any known limitations on the AWS version of Elasticache? E.g. config options we have less control over, lack of an actual super user (such as having to use `rds_superuser` for RDS which isn't an actual superuser).

surminus: I'll defer to @deanwilson and @afda16 who set this up. I don't have enough experience with Elasticache.

deanwilson: Yes,  there are some options we can't use under Elasticache. Mostly these are things we wouldn't want in an AWS environment but if we hit any that are actual issues for us we can amend the ADR and move back to a server/container.

SamLR: Thought so, might be worth making a note of this in the ADR? (sorry @surminus)

surminus: This has been updated.

PR #260, opened 2017-09-05
Add variable to override default instance AMI

afda16:
Currently the node_group module uses the `aws_ami` data source to determine the newest AMI ID
published by Ubuntu within a major version and use that in our launch configurations. This parameter
could be overriden in each project to set a specific image name, and this is going to be the needed
to avoid having to recreate the launch_configuration every time we get a new image patch.

This commit adds a new variable to all the projects to make sure we set the same image everywhere.
The default image version should be set in the common data file.

### DIFF
--- a/terraform/projects/app-api-lb/main.tf
+++ b/terraform/projects/app-api-lb/main.tf
@@ -8,6 +8,7 @@
 # stackname
 # aws_environment
 # ssh_public_key
+# instance_ami_filter_name
 # elb_certname
 #
 # === Outputs:
@@ -32,6 +33,12 @@ variable "aws_environment" {
 variable "ssh_public_key" {
   type        = "string"
   description = "Default public key material"
+}
+
+variable "instance_ami_filter_name" {
+  type        = "string"
+  description = "Name to use to find AMI images"
+  default     = ""
 }
 
 variable "app_service_records" {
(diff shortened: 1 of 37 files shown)


terraform/modules/aws/node_group/main.tf
```diff
@@ -65,7 +66,13 @@ variable "instance_security_group_ids" {
 
 variable "instance_ami_filter_name" {
   type        = "string"
-  description = "Name to use to find AMI images for the instance"
```
SamLR: Is there a reason for providing a default variable with a default value and an override variable?

afda16: It is just to help controlling a final default value that is going to be set independently of how the override parameter is used. Because we always want to pass a override value to avoid issues with patch versions, we can expect the projects to always set a default one (in which case we can leave node_group unmodified), or we can no trust the projects and make sure we always apply a non-empty default value in node_group with the two parameters and coalesce function. Either option I think is fine, it just depends on how flexible and simple we want to be and how much we want to trust things

SamLR: I think the simpler option is probably the one we want to take

afda16:
I updated the commit to leave the node_group as it was. It's expected a "default" variable in the data directory to set the image name.

PR #261, opened 2017-09-05
Add a clustering option to redis module

surminus:
We are seeing errors with publishing-api which seem to be related with redis.

Update the module so you can choose either clustered or non-clustered. If non-clustered is chosen, then a master with a read replica is created on redis 3.2.


surminus:
I have tested this as [successfully resolving errors with publishing-api](https://deploy.blue.integration.govuk.digital/job/Deploy_App/260/), but I am unsure if the terraform code for creating the DNS CNAME record works for the non-clustered option so would appreciate a particular check and test for this resource.

PR #262, opened 2017-09-05
Allow router-api to speak to router

surminus:
Router API needs to be able to talk to the router to reload routes.

### DIFF
--- a/terraform/projects/infra-security-groups/cache.tf
+++ b/terraform/projects/infra-security-groups/cache.tf
@@ -47,6 +47,20 @@ resource "aws_security_group_rule" "allow_cache_external_elb_in" {
   source_security_group_id = "${aws_security_group.cache_external_elb.id}"
 }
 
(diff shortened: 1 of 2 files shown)


terraform/projects/infra-security-groups/cache.tf
```diff
@@ -47,6 +47,20 @@ resource "aws_security_group_rule" "allow_cache_external_elb_in" {
...
 
+# Allow the router-backend instances to reload router routes
+resource "aws_security_group_rule" "allow_cache_from_router-backend" {
+  type      = "ingress"
+  from_port = 3055
+  to_port   = 3055
+  protocol  = "tcp"
+
+  # Which security group is the rule assigned to
+  security_group_id = "${aws_security_group.cache.id}"
```
afda16: Is router-backend connecting directly with the cache instances?

afda16: ... Got it

PR #263, opened 2017-09-05
Add Asset Master

surminus:
We're using EFS as the datastore, this changes the functionality a little from how things work in our current environment. I will write an ADR to explain this, but we have a central EFS share that all machines connect to, rather than relying on the asset-master to create the share.

Security groups reflect that each machine should mount the central EFS share over port 111 on TCP/UDP.

https://trello.com/c/YGAQ2KDD/732-aws-migration-asset-master-asset-slave


terraform/projects/app-asset-master/main.tf
```diff
@@ -0,0 +1,92 @@
...
+  instance_elb_ids              = []
+  asg_max_size                  = "1"
+  asg_min_size                  = "1"
+  asg_desired_capacity          = "1"
+  root_block_device_volume_size = "30"
+}
+
+# Outputs
+# --------------------------------------------------------------
+output "efs_mount_target_dns_names" {
```
deanwilson: Should `"aws_route53_record" "assets_service_record"` also be an output or is that DNS record unneeded?

surminus: It should probably be an output, but I don't think it's required by anything.

surminus: @deanwilson updated.

PR #269, opened 2017-09-12
Add external ELB to backend

surminus:
Adding an external ELB to backend, along with all the service records that originally existed on backend-lb, means that we can direct traffic directly to the backend instances and remove the backend-lb tier.

Services speak to backend both internally and externally, so an internal and external ELB are both required.

https://trello.com/c/goRJMaPL/712-remove-aws-backend-lb-machines


terraform/projects/app-backend/main.tf
```diff
@@ -49,15 +60,72 @@ provider "aws" {
...
+data "aws_acm_certificate" "elb_internal_cert" {
+  domain   = "${var.elb_internal_certname}"
   statuses = ["ISSUED"]
 }
 
-resource "aws_elb" "backend_elb" {
-  name            = "${var.stackname}-backend"
+resource "aws_elb" "backend_elb_external" {
+  name            = "${var.stackname}-backend-external"
+  subnets         = ["${data.terraform_remote_state.infra_networking.private_subnet_ids}"]
```
SamLR: I think this needs to be associated with the public subnets, not the private ones.

PR #271, opened 2017-09-12
Add a transition_db_admin RDS instance

issyl0:


issyl0:
This only sets up the primary machine, not the slave that Bouncer will eventually read from.

terraform/projects/app-transition-db-admin/main.tf
```diff
@@ -0,0 +1,114 @@
...
+  connection_draining_timeout = 400
+
+  tags = "${map("Name", "${var.stackname}-transition-db-admin", "Project", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "transition_db_admin")}"
+}
+
+module "transition-db-admin" {
+  source                        = "../../modules/aws/node_group"
+  name                          = "${var.stackname}-transition-db-admin"
+  vpc_id                        = "${data.terraform_remote_state.infra_vpc.vpc_id}"
+  default_tags                  = "${map("Project", var.stackname, "aws_stackname", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "db_admin", "aws_hostname", "transition-db-admin-1")}"
```
surminus: The `aws_migration` value should be `transition_db_admin` to match the Puppet node class.

terraform/projects/app-transition-db-admin/main.tf
```diff
@@ -0,0 +1,114 @@
...
+  instance_elb_ids              = ["${aws_elb.transition-db-admin_elb.id}"]
+  asg_max_size                  = "1"
+  asg_min_size                  = "1"
+  asg_desired_capacity          = "1"
+  root_block_device_volume_size = "64"
+}
+
+resource "aws_route53_record" "transition_db_admin_service_record" {
+  zone_id = "${data.terraform_remote_state.infra_stack_dns_zones.internal_zone_id}"
+  name    = "transition_db-admin.${data.terraform_remote_state.infra_stack_dns_zones.internal_domain_name}"
```
surminus: I think this should be `transition-db-admin` as we never use underscores in host names

terraform/projects/app-transition-db-admin/main.tf
issyl0: I would have been surprised if I hadn't missed one. 👍

issyl0:
I forgot the security groups, so I updated the commit to include them.

PR #274, opened 2017-09-13
Run build tool from anywhere

surminus:
This commit adds a new argument (`GOVUK_AWS_REPO`) which you can use to specify the location of this repo. It exchanges relative paths for absolute paths, which gives the user the option of running the script from anywhere.

I've submitted this commit because I'd prefer to have the script in `/usr/local/bin` and run from whatever directory I'm currently working from.

I've also changed the name because I love a tool with a snappy name.

### DIFF
--- a/tools/govuk-tf
+++ b/tools/govuk-tf
@@ -7,13 +7,14 @@
 
 set -e
 
-while getopts "c:d:e:p:s:h" option
+while getopts "c:d:e:p:s:r:h" option
 do
   case $option in
     c ) CMD=$OPTARG ;;
     d ) DATA_DIR=$OPTARG ;;
     e ) ENVIRONMENT=$OPTARG ;;
     p ) PROJECT=$OPTARG ;;
+    r ) GOVUK_AWS_REPO=$OPTARG ;;
     s ) STACKNAME=$OPTARG ;;
     h ) HELP=1 ;;
   esac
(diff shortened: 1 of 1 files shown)


issyl0:
I'd prefer to keep this as it is, as I have a reference point then for what each of the option names are (I can tab-complete directory names beforehand (stacks and apps). And running `source .blue.integration.env`, as I learnt yesterday, is also in the root directory, so it makes sense to keep it running from the same place. And you always know which terraform you're running, if you explicitly have to go into `govuk-aws` to run it.

surminus:
You can still run it from the root of the repo. If you don't specify `GOVUK_AWS_REPO` then it assumes the current directory is the root. It just means we open it up to work in different ways. I have some pretty expansive shell completion and I find having to go to the repo root or opening up a new tab to run commands more intrusive to the way I work.

I think we're all using `chtf`, which I think is based upon an env var in your current shell rather than directory location (which is what `tfenv` does), so `terraform` versions shouldn't be affected.

People should test and let me know if we see any issues.

afda16:
Update `usage()` to include new parameter, I like the changes

surminus:
Updated with usage.

surminus:
I replaced my usage of this tool with https://github.com/surminus/terragov so I'm not bothered about merging this, shall I close?

surminus:
Closing due to lack of response.

PR #278, opened 2017-09-13
In the db-admin docs, specify a destination for an `scp` command

issyl0:


doc/guides/db-admin-database-restore-process.md
```diff
@@ -30,7 +30,7 @@ it's key to make everything work.
    (`-A`), and copy the file to the dbadmin machine. You'll need to find the
    `db_admin` machine, so use `govuk_node_list` for this.
 
-        scp -C publishing_api_integration.dump `govuk_node_list -c db_admin`
+        scp -C publishing_api_integration.dump `govuk_node_list -c db_admin`:.
```
surminus: Actually, don't think you need to have the `.` on the end.

issyl0: I didn't know that. Should I change all other occurrences of `:.` in these docs to just `:` then?

surminus: Nah, both are fine.

PR #287, opened 2017-09-19
Increase default disk size to 20GB

SamLR:
This will bring the default inline with carrenza


SamLR:
I think it's probably best doing that as we go

PR #289, opened 2017-09-19
Add external and internal ELBs to content stores

afda16:
Update content-store and draft-content-store to use an internal
and external ELB in both cases. Use dedicated security groups
for each type of ELB.

### DIFF
--- a/terraform/projects/infra-security-groups/draft-content-store.tf
+++ b/terraform/projects/infra-security-groups/draft-content-store.tf
@@ -21,7 +21,7 @@ resource "aws_security_group" "draft-content-store" {
   }
 }
 
-resource "aws_security_group_rule" "allow_draft-content-store_elb_in" {
+resource "aws_security_group_rule" "allow_draft-content-store_internal_elb_in" {
   type      = "ingress"
   from_port = 80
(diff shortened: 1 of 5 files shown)


terraform/projects/infra-security-groups/draft-content-store.tf
```diff
@@ -31,25 +31,80 @@ resource "aws_security_group_rule" "allow_draft-content-store_elb_in" {
...
+
+# Content Store should be available externally
+resource "aws_security_group_rule" "allow_public_to_draft-content-store_external_elb" {
+  type      = "ingress"
+  from_port = 443
+  to_port   = 443
+  protocol  = "tcp"
+
+  security_group_id = "${aws_security_group.draft-content-store_external_elb.id}"
+  cidr_blocks       = ["0.0.0.0/0", "${var.office_ips}"]
```
surminus: I think content-store itself is fine to be available externally, but AFAIK draft-content-store should not be available to the public, so should be limited (perhaps to office IPs only?)

surminus:
One small comment, but looks fine

afda16:
@surminus updated, thanks!

PR #290, opened 2017-09-19
Remove logs-elasticsearch, logs-redis and logging

surminus:
We're migrating to do our logging on The GDS Way (tm) sanctioned ELK SaaS provider. We're now sending all our logs to this provider so we can switch off the associated instances.

Before merging, we need to update Puppet code: https://github.com/alphagov/govuk-puppet/pull/6512


surminus:
Since the following have been merged we should be able to remove these machines from AWS:

https://github.com/alphagov/govuk-puppet/pull/6528
https://github.com/alphagov/govuk-puppet/pull/6512

PR #291, opened 2017-09-20
Make instance type a variable for mapit

SamLR:
Start using a variable to configure instance type for Mapit. This will
likely need to be done eventually (to allow larger instances in
production/staging) and this means that the two ASGs will both use the
same instance type.

### DIFF
--- a/terraform/projects/app-mapit/main.tf
+++ b/terraform/projects/app-mapit/main.tf
@@ -57,6 +57,12 @@ variable "elb_internal_certname" {
   description = "The ACM cert domain name to find the ARN of"
 }
 
+variable "instance_type" {
+  type        = "string"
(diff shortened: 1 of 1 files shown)


terraform/projects/app-mapit/main.tf
```diff
@@ -57,6 +57,12 @@ variable "elb_internal_certname" {
   description = "The ACM cert domain name to find the ARN of"
 }
 
+variable "instance_type" {
+  type        = "string"
+  description = "The type of EC2 instance to use for both ASGs."
+  default     = "t2.micro"
```
surminus: Can we make the default medium?

SamLR: of course

PR #299, opened 2017-09-22
Apt update ELB config and add certs

afda16:
Update the ELB configuration with internal and external certificates
and instance port/protocols


surminus:
Probably need security group changes, too.

afda16:
@surminus updated

PR #302, opened 2017-09-22
Add more disk to app servers

surminus:


surminus:
Default was raised, closing.

PR #303, opened 2017-09-25
Use private EIPs rather than ELBs on mongo

surminus:
We're seeing an issue where applications using the mongo cluster are unable to connect to the cluster despite it being available.

We believe it may be related to the dynamic nature of ELBs, and the mongo-ruby-driver currently being unable to handle DNS updates without a reload of the application.

This commit removes the load balancers, creates an ENI and an EIP, and attaches the ENI to the instance on boot using userdata.

As per the documentation, an IP associated with an ENI should follow the interface to any attached instance:

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html

We can assign a DNS record to the IP address so the Puppet code underneath does not have to change.

https://trello.com/c/luBryWgW/748-change-the-way-mongo-in-aws-works

### DIFF
--- a/terraform/projects/app-mongo/additional_policy.json
+++ b/terraform/projects/app-mongo/additional_policy.json
@@ -1,0 +1,16 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Sid": "Stmt1499854881000",
+            "Effect": "Allow",
+            "Action": [
+                "ec2:DescribeNetworkInterfaces",
+                "ec2:AttachNetworkInterface"
+            ],
+            "Resource": [
+                "*"
+            ]
+        }
+    ]
+}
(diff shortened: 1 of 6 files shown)


terraform/userdata/30-associate-eni
```diff
@@ -0,0 +1,29 @@
...
+echo "[$(date '+%H:%M:%S %d-%m-%Y')] START SNIPPET: associate-eni"
+
+# The environment variables INSTANCE_ID and REGION are set
+# in the base snippet
+#
+F_MIGRATION=$(facter aws_migration)
+F_STACKNAME=$(facter aws_stackname)
+F_HOSTNAME=$(facter aws_hostname)
+
+echo "[$(date '+%H:%M:%S %d-%m-%Y')] attach-volume: find network interface ID ..."
```
afda16: Log message still says volume

PR #304, opened 2017-09-25
Test Prometheus

afda16:
Add some security rules and test instance to start installing
Prometheus. The internal ELB rules aren't properly set yet, this
configuration is just a placeholder.

We'll remove this project after the firebreak


surminus:
I'm merging so I can deploy my changes without removing these ones.

PR #309, opened 2017-09-29
[WIP] Create frontend swarm cluster

surminus:
This creates the "docker-frontend" project, which includes:

An ASG comprised of 2 workers
An ASG comprised of a single manager

They are joined to the same security group that allows full cluster communications.

The idea is that a working Docker Swarm cluster could entirely replace our frontend application instances (currently frontend,
calculators-frontend and content-store).

If a machine goes away, then it should hopefully be able to rejoin the cluster and continue running the services it requires in that environment.

https://trello.com/c/MdGM72w9/750-docker-swarm


surminus:
Firebreak over!

PR #310, opened 2017-10-02
[WIP] Initial Puppetmaster 5 provisioning

afda16:
To start testing Puppet5 we are going to provision a new 'green' stack
for the app-puppetmaster project. Userdata and puppetserver bootstrap
scripts need to be updated to reflect the new version requirements.


deanwilson:
This work is not currently prioritised so we're closing the PR but leaving the branch for when we can eventually return to it.

PR #321, opened 2017-10-19
Update RDS module and add replica

surminus:

Please see detail in commits.

This updates the module with some sane defaults, and assigns a replica for app-postgresql.

### DIFF
--- a/terraform/modules/aws/rds_instance/main.tf
+++ b/terraform/modules/aws/rds_instance/main.tf
@@ -109,6 +109,30 @@ variable "parameter_group_name" {
   default     = ""
 }
 
+variable "skip_final_snapshot" {
+  type        = "string"
+  description = "Set to false to create a final snapshot when the cluster is deleted."
+  default     = "true"
+}
+
+variable "maintenance_window" {
+  type        = "string"
+  description = "The window to perform maintenance in."
+  default     = "Mon:01:00-Mon:03:00"
+}
+
+variable "backup_retention_period" {
+  type        = "string"
+  description = "The days to retain backups for."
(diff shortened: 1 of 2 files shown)


surminus:
I've added two more commits which change the way in which the module works.

I'm also aware that I've created the replica in a different way that was envisioned; I think that we it was envisioned to have a separate project for the replica rather than rolling it into the same project.

I don't think this makes sense. A database cluster should be an atomic thing, and having to manage them separately seems counter intuitive to me, but I am willing to be convinced otherwise.

terraform/modules/aws/rds_instance/main.tf
```diff
@@ -209,17 +209,33 @@ resource "aws_db_instance" "db_instance" {
 #
 
 output "rds_instance_id" {
-  value = "${var.create_replicate_source_db > 0 ? join("", aws_db_instance.db_instance_replica.*.id) : aws_db_instance.db_instance.id}"
+  value = "${aws_db_instance.db_instance.id}"
+}
+
+output "rds_replica_id" {
+  value = "${aws_db_instance.db_instance_replica.id}"
```
afda16: does this work when the module is applied on the master? I remember having to add the conditional because otherwise the resource does not exist and the output value cacn't be evaluated

surminus: I ran it against a module that doesn't use a replica, and it didn't change anything related to DNS records or the outputs. 

I think the resource will always exist, albeit with a `count` of `0` which means that nothing is outputted. I did not receive any errors with an empty output.

PR #327, opened 2017-10-20
Add AWS alarms Elasticache module

afda16:
Add module to monitor ElastiCache metrics. For now it only includes
host alarms.

### DIFF
--- a/terraform/modules/aws/alarms/elasticache/main.tf
+++ b/terraform/modules/aws/alarms/elasticache/main.tf
@@ -1,0 +1,95 @@
+/**
+* ## Module: aws::alarms::elasticache
+*
+* This module creates the following CloudWatch alarms in the
+* AWS/ElastiCache namespace:
+*
+*   - CPUUtilization greater than or equal to threshold
+*   - FreeableMemory less than threshold
+*
+* All metrics are measured during a period of 60 seconds and evaluated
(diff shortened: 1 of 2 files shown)


terraform/modules/aws/alarms/elasticache/main.tf
```diff
@@ -0,0 +1,95 @@
...
+}
+
+variable "cache_node_id" {
+  type        = "string"
+  description = "The ID of the cache node that we want to monitor."
+}
+
+variable "cpuutilization_threshold" {
+  type        = "string"
+  description = "The value against which the CPUUtilization metric is compared. Defaults to 80."
```
deanwilson: Remove default from docs as it's given in the code.

terraform/modules/aws/alarms/elasticache/main.tf
```diff
@@ -0,0 +1,95 @@
...
+*
+*/
+variable "name_prefix" {
+  type        = "string"
+  description = "The alarm name prefix."
+}
+
+variable "alarm_actions" {
+  type        = "list"
+  description = "The list of actions to execute when this alarm transitions into an ALARM state from any other state. Each action is specified as an Amazon Resource Number (ARN)."
```
deanwilson: "from any other state" isn't needed

PR #330, opened 2017-10-23
Add an RDS replica to Transition PostgreSQL database

issyl0:
- Bouncer uses the database secondary (here, standby) to read its data, so we set one up.


issyl0:
This seems to `plan` correctly.

surminus:
This seems to create a full new RDS instance rather than a secondary. See [here](https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/app-postgresql/main.tf#L74-L83) for setting up a standby.

issyl0:
I did the thing in the link first, following your PR to add replicas to `app-postgresql`, and then wondered why there were so many more options and thought I'd done it wrong so started again. I'll do it again again. 😪

issyl0:
@surminus I just re-did this. :-)

PR #334, opened 2017-10-24
Add a project for a database backups bucket

timmow:
This contains a bucket and an iam policy allowing to write to this
bucket. Once this is applied the iam policy will be attached to instance
roles to allow them to write to the bucket.


surminus:
One thing to note is that we've changed the format of our Terraform files. Check the format in other projects, and then [terraform-docs](https://github.com/segmentio/terraform-docs) to create README.md in the project.

timmow:
This has been updated to include the attachment of the profile to the instance roles that need to write to the bucket

PR #343, opened 2017-10-27
Cloudwatch logs exporter

afda16:
Add resources, module and implementation to export CloudWatch log groups to S3 via Lambda-KinesisFirehose

The `infra-aws-logging` project continues providing the basic resources to manage infrastructure logs in our platform, including IAM roles and policies to process CloudWatch logs.

The `cloudwatch_log_exporter` module allows us to provide a CloudWatch log group and Lambda function to export those logs to S3. This module has been implemented in infra-vpc to start exporting the VPC flow logs.

https://trello.com/c/J5zD19nv/803-send-rds-logs-to-s3


surminus:
I think this looks good, but can we squash the first two commits? The first one adds a bunch of stuff, which then gets deleted/moved in the second and so coming back to reading the two might be confusing.

PR #344, opened 2017-10-31
Add lifecycle rules to database backups

surminus:
Add lifecycle rules to our database backups. In line with what we currently do, keep backups for 3 months, but move them to lower storage costs after 30 and 60 days respectively.

With mongodb, keep "regular" backups to 7 days so we have point in time backups that go back 7 days. Keep "daily" backups for 90 days.

### DIFF
--- a/terraform/projects/infra-database-backups-bucket/main.tf
+++ b/terraform/projects/infra-database-backups-bucket/main.tf
@@ -36,4 +36,70 @@ resource "aws_s3_bucket" "database_backups" {
     Name            = "govuk-${var.aws_environment}-database-backups"
     aws_environment = "${var.aws_environment}"
   }
+
+  lifecycle_rule {
+    prefix  = "mysql/"
+    enabled = true
+
+    transition {
+      storage_class = "STANDARD_IA"
+      days          = 30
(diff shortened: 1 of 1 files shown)


terraform/projects/infra-database-backups-bucket/main.tf
```diff
@@ -36,4 +36,70 @@ resource "aws_s3_bucket" "database_backups" {
...
+    prefix  = "postgres/"
+    enabled = true
+
+    transition {
+      storage_class = "STANDARD_IA"
+      days          = 30
+    }
+
+    transition {
+      storage_class = "GLACIER"
```
timmow: I wonder if its worth sending to Glacier - from [the docs](http://docs.aws.amazon.com/AmazonS3/latest/dev/lifecycle-transition-general-considerations.html)

> Number of days you plan to keep objects archived – Amazon Glacier is a long-term archival solution. Deleting data that is archived to Amazon Glacier is free if the objects you delete are archived for three months or longer. If you delete or overwrite an object within three months of archiving it, Amazon S3 charges a prorated early deletion fee.

PR #348, opened 2017-11-03
Prepare to enable backups on Rummager ElasticSearch machines

issyl0:
- Add three new variables and use them as tags for `backups_enabled`, so
  we can enable backups on only one out of three machines.


terraform/projects/app-rummager-elasticsearch/main.tf
```diff
@@ -50,6 +50,24 @@ variable "cluster_name" {
   description = "Name of the Elasticsearch cluster to use for discovery"
 }
 
+variable "rummager_elasticsearch_1_backups_enabled" {
+  type        = "string"
+  description = "Whether or not this machine takes the ES backups"
+  default     = "false"
```
afda16: https://www.terraform.io/docs/configuration/variables.html#booleans

Have you tried passing the variable with `-var` ? Maybe we should just use "0" and "1"

issyl0: Now I know why the values appear as `0` or `1` when I look for the tags!

PR #349, opened 2017-11-06
Allow rummager elasticsearch to write to backups bucket

surminus:
Attach the policy that would allow each instance to write snapshots to
the database backups S3 bucket.

Related to https://github.com/alphagov/govuk-aws/pull/348

https://trello.com/c/wl3p8Y9D/787-investigate-elasticsearch-backups

### DIFF
--- a/terraform/projects/app-db-admin/main.tf
+++ b/terraform/projects/app-db-admin/main.tf
@@ -32,7 +32,7 @@ variable "instance_ami_filter_name" {
   default     = ""
 }
 
-variable "remote_state_infra_database_backups_bucket_stack" {
+variable "remote_state_infra_database_backups_bucket_key_stack" {
   type        = "string"
   description = "Override stackname path to infra_database_backups_bucket remote state"
   default     = ""
@@ -139,7 +139,7 @@ data "terraform_remote_state" "infra_database_backups_bucket" {
 
   config {
     bucket = "${var.remote_state_bucket}"
-    key    = "${coalesce(var.remote_state_infra_database_backups_bucket_stack, var.stackname)}/infra-database-backups-bucket.tfstate"
+    key    = "${coalesce(var.remote_state_infra_database_backups_bucket_key_stack, var.stackname)}/infra-database-backups-bucket.tfstate"
     region = "eu-west-1"
   }
 }
--- a/terraform/projects/app-db-admin/main.tf
+++ b/terraform/projects/app-db-admin/main.tf
@@ -32,6 +32,12 @@ variable "instance_ami_filter_name" {
   default     = ""
 }
 
+variable "remote_state_infra_database_backups_bucket_stack" {
+  type        = "string"
+  description = "Override stackname path to infra_database_backups_bucket remote state"
+  default     = ""
+}
+
 # Resources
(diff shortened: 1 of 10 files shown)


terraform/projects/app-rummager-elasticsearch/main.tf
```diff
@@ -248,6 +248,31 @@ module "alarms-elb-rummager-elasticsearch-internal" {
   healthyhostcount_threshold     = "1"
 }
 
+data "terraform_remote_state" "infra_database_backups_bucket" {
+  backend = "s3"
+
+  config {
+    bucket = "${var.remote_state_bucket}"
+    key    = "${coalesce(var.remote_state_infra_vpc_key_stack, var.stackname)}/infra-database-backups-bucket.tfstate"
```
afda16: It looks a little bit like a hack, to use other remote_state variable to refer to the backups stack. It breaks the way we've been doing remote states

surminus: I think this was a mistake rather than intentional, and I pinched it from where it had been set previously. I will update all of them.

surminus: Oh, maybe it was intentional because that variable isn't set. Interesting.

terraform/projects/app-rummager-elasticsearch/main.tf
```diff
@@ -248,6 +254,22 @@ module "alarms-elb-rummager-elasticsearch-internal" {
...
+  config {
+    bucket = "${var.remote_state_bucket}"
+    key    = "${coalesce(var.remote_state_infra_database_backups_bucket_stack, var.stackname)}/infra-database-backups-bucket.tfstate"
+    region = "eu-west-1"
+  }
+}
+
+resource "aws_iam_role_policy_attachment" "rummager-elasticsearch-1_database_backups_iam_role_policy_attachment" {
+  count      = 3
+  role       = "${element(list(module.rummager-elasticsearch-1.instance_iam_role_name, module.rummager-elasticsearch-2.instance_iam_role_name, module.rummager-elasticsearch-3.instance_iam_role_name))}"
```
afda16: `element` needs 2 arguments, it's missing `count.index`

https://www.terraform.io/docs/configuration/interpolation.html#element-list-index-

surminus: Thanks, updated!

terraform/projects/app-db-admin/main.tf
```diff
@@ -32,6 +32,12 @@ variable "instance_ami_filter_name" {
   default     = ""
 }
 
+variable "remote_state_infra_database_backups_bucket_stack" {
```
afda16: I didn't realise this before ... `remote_state_infra_artefact_bucket_stack ` and `remote_state_infra_database_backups_bucket_stack`

I personally would call everything `_key_stack` because it refers to the bucket key, same as the rest of remote state variables. We could change it later

surminus: I'll update it now as pushing this stuff down the line is just going to get worse and won't take me a moment. Good spot.

PR #352, opened 2017-11-06
Rds logs to s3

afda16:


terraform/projects/app-transition-postgresql/main.tf
```diff
@@ -105,6 +110,24 @@ module "alarms-rds-transition-postgresql-standby" {
   replicalag_threshold = "120"
 }
 
+module "transition-postgresql-primary_log_exporter" {
+  source                       = "../../modules/aws/rds_log_exporter"
+  rds_instance_id              = "${module.transition-postgresql-primary_rds_instance.rds_instance_id}"
+  s3_logging_bucket_name       = "${data.terraform_remote_state.infra_aws_logging.aws_logging_bucket_id}"
+  lambda_filename              = "${path.module}/../../lambda/RDSLogsToS3/RDSLogsToS3.zip"
```
surminus: Why do we specify the `lambda_filename` here rather than specifying in the module itself? Is there a case of using a different lambda function for different uses of RDS?

surminus:
I've left a question (rather than a change) but otherwise it looks OK to me. Not approving as I'd prefer a secondary reviewer to have a look.

terraform/projects/app-transition-postgresql/main.tf
timmow: I was looking at this and wondered the same thing

afda16: I think the module should be generic, no knowledge about local paths. In this case the lambda function is always going to be the same, it's similar to the way in which it's done in cloudwath_log_exporter.

I agree it's not nice, If instead of a filename we were using S3 then we could make it nicer.

terraform/projects/app-transition-postgresql/main.tf
```diff
@@ -105,6 +110,24 @@ module "alarms-rds-transition-postgresql-standby" {
...
+  lambda_role_arn              = "${data.terraform_remote_state.infra_aws_logging.lambda_rds_logs_to_s3_role_arn}"
```
timmow: Similarly for the role - I guess this might be to do with passing data between modules?

afda16: Yes, the role resources are managed in infra-aws-logging and reused in the module. Other option would be to create the role/policies inside the module itself, we would have a role/policy for each lambda function created, instead of reapplying the same one

terraform/projects/app-transition-postgresql/main.tf
timmow: Agreed the module shouldn't know about local file paths outside of the module - I wonder but dont feel strongly if the lambda file should be encapsulated within the module, rather than being in a shared area for lambda files

terraform/projects/app-transition-postgresql/main.tf
timmow: Yes I see, better to have one role rather than one per lambda created I think

terraform/projects/app-transition-postgresql/main.tf
surminus: I guess it depends what the use of making it generic is. If we want to allow other people to use it to quickly ship RDS logs, then shipping the lambda function along with the module makes a certain amount of sense, but if we're using modules more like frameworks than all-purpose pieces of code, then keeping it separate makes sense (and could almost argue to have an even more generic function).

Either way, I do not feel that strongly about it, so happy to approve. I think it may be worth writing a styleguide for the way that we write and use our modules.

afda16: We can have a wider discussion about how we want to manage lambda, maybe? I don't like the filename parameter but would like to have a common approach for Lambda

surminus: Sure, seems like a good idea. I have approved, anyway!

PR #356, opened 2017-11-07
Add MySQL read replica

surminus:
Add a MySQL read replica. In our Puppet code this was previously called "mysql-slave", so take this opportunity to rename to "replica".

### DIFF
--- a/terraform/projects/app-mysql/main.tf
+++ b/terraform/projects/app-mysql/main.tf
@@ -46,6 +46,7 @@ provider "aws" {
   version = "1.0.0"
 }
 
+# MySQL Primary instance
 module "mysql_primary_rds_instance" {
   source               = "../../modules/aws/rds_instance"
   name                 = "${var.stackname}-mysql-primary"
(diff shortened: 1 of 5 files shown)


terraform/projects/app-mysql/main.tf
```diff
@@ -99,6 +100,41 @@ module "mysql_primary_log_exporter" {
...
+
+resource "aws_route53_record" "replica_service_record" {
+  zone_id = "${data.terraform_remote_state.infra_stack_dns_zones.internal_zone_id}"
+  name    = "mysql-replica.${data.terraform_remote_state.infra_stack_dns_zones.internal_domain_name}"
+  type    = "CNAME"
+  ttl     = 300
+  records = ["${module.mysql_replica_rds_instance.rds_replica_address}"]
+}
+
+module "alarms-rds-mysql-replica" {
```
afda16: It would be good to enable the replicaLag alarm, same as https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/app-postgresql/main.tf#L110

PR #357, opened 2017-11-07
Update docs

surminus:
I wrote some bash that is a script we can use to occasionally update docs when
we get the urge.

### DIFF
--- a/terraform/modules/aws/node_group/README.md
+++ b/terraform/modules/aws/node_group/README.md
@@ -41,5 +41,6 @@ policies to in other modules.
 
 | Name | Description |
 |------|-------------|
+| autoscaling_group_name | The name of the node auto scaling group. |
 | instance_iam_role_name | Node IAM Role Name. Use with aws_iam_role_policy_attachment to attach specific policies to the node role |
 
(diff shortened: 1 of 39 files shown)


terraform/modules/aws/network/nat/README.md
```diff
@@ -0,0 +1,20 @@
...
+| Name | Description | Type | Default | Required |
+|------|-------------|:----:|:-----:|:-----:|
+| subnet_ids | List of public subnet IDs where you want to create a NAT Gateway | list | - | yes |
+| subnet_ids_length | Length of subnet_ids variable | string | - | yes |
+
+## Outputs
+
+| Name | Description |
+|------|-------------|
+| nat_gateway_ids | Outputs -------------------------------------------------------------- |
```
afda16: I updated the docs for the network modules in https://github.com/alphagov/govuk-aws/pull/358 , I didn't realise this change.

Some modules need to include a blank line before the first `output` resource to avoid that "Outputs -------------------------------------------------------------- " description, that is getting from the comment right before `output`. The VPC module was also missing output descriptions.

We could update this PR or remove the network modules from here and use my changes instead.

afda16:
I rebased and fixed one of the docs

PR #366, opened 2017-11-20
Allow internal hosts to connect to alert on port 443

deanwilson:
We have a number of hosts running the `check_icinga` plugin. These are
currently failing to connect. This changeset should permit that access.


terraform/projects/infra-security-groups/monitoring.tf
```diff
@@ -97,6 +110,16 @@ resource "aws_security_group_rule" "allow_management_to_monitoring_internal_elb_
   source_security_group_id = "${aws_security_group.management.id}"
 }
 
+resource "aws_security_group_rule" "allow_management_to_monitoring_internal_elb_https" {
+  type      = "ingress"
+  from_port = 80
```
surminus: I think this needs to be `443` as well.

PR #368, opened 2017-11-20
Add AWS lb module

afda16:
Add AWS lb module to create application and network Load Balancers.
It requires aws module > 1.2.0

### DIFF
--- a/terraform/modules/aws/lb/main.tf
+++ b/terraform/modules/aws/lb/main.tf
@@ -1,0 +1,214 @@
+/**
+* ## Modules: aws/lb
+*
+* This module creates a Load Balancer resource, with associated
+* listeners and target groups.
+*
+* Listeners and target groups are defined with map variables,
+* where the key is the name of the resource, and the value is
+* the port and protocol of the resource, specified as `PROTOCOL:PORT`.
+*
+* Listeners are associated to target groups with the `listener_target_groups`
+* map, that uses the name of the resources.
+*
+* When using a listener on HTTPS, we can specify the certificate with
+* the `listener_certificates` variable, where the value is the domain name
+* of the certificate as registered on AWS. The value is used by an
+* `aws_acm_certificate` data resource to find the certificate ARN.
+*
+*```
+* listeners = {
+*   "myapp-http-80"   = "HTTP:80"
+*   "myapp-https-443" = "HTTPS:443"
+* }
+*
+* target_groups = {
+*   "http-80" = "HTTP:80"
+* }
+*
+* listener_target_groups = {
+*   "myapp-http-80"   = "http-80"
+*   "myapp-https-443" = "http-80"
+* }
+*
+* listener_certificates = {
+*   "myapp-https-443" = "mydomain.com"
+* }
+*
+*```
+*/
+
+variable "default_tags" {
+  type        = "map"
+  description = "Additional resource tags"
+  default     = {}
+}
+
+variable "vpc_id" {
+  type        = "string"
+  description = "The ID of the VPC in which the target groups are created."
+}
+
+variable "load_balancer_type" {
+  type        = "string"
+  description = "The type of load balancer to create. Possible values are application or network. The default value is application."
+  default     = "application"
+}
+
+variable "access_logs_bucket_name" {
+  type        = "string"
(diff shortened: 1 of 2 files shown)


terraform/modules/aws/lb/main.tf
```diff
@@ -0,0 +1,198 @@
...
+}
+
+resource "aws_lb" "lb" {
+  name               = "${var.name}"
+  internal           = "${var.internal}"
+  security_groups    = ["${var.security_groups}"]
+  subnets            = ["${var.subnets}"]
+  load_balancer_type = "${var.load_balancer_type}"
+
+  enable_deletion_protection = false
```
surminus: Should this be a variable? Is there any case where we might want to enable this?

terraform/modules/aws/lb/main.tf
```diff
@@ -0,0 +1,198 @@
...
+
+  access_logs {
```
surminus: Awesome.

terraform/modules/aws/lb/main.tf
```diff
@@ -0,0 +1,198 @@
...
+  count             = "${length(keys(var.listeners))}"
+  load_balancer_arn = "${aws_lb.lb.arn}"
+  port              = "${element(split(":", lookup(var.listeners, element(keys(var.listeners), count.index))), 1)}"
+  protocol          = "${element(split(":", lookup(var.listeners, element(keys(var.listeners), count.index))), 0)}"
+  ssl_policy        = "ELBSecurityPolicy-2015-05"
+  certificate_arn   = "${lookup(local.listener_certificate_arns, element(keys(var.listeners), count.index), "")}"
+
+  default_action {
+    target_group_arn = "${matchkeys(values(local.target_groups_arns), keys(local.target_groups_arns), lookup(var.listener_target_groups, element(keys(var.listeners), count.index)))}"
+    type             = "forward"
```
surminus: From the docs: "The type of routing action. The only valid value is forward."

It really upsets me this has to be specified.

terraform/modules/aws/lb/main.tf
```diff
@@ -0,0 +1,198 @@
...
+  }
+}
+
+resource "aws_lb_target_group" "tg" {
+  count                = "${length(keys(var.target_groups))}"
+  name                 = "${element(keys(var.target_groups), count.index)}"
+  port                 = "${element(split(":", lookup(var.target_groups, element(keys(var.target_groups), count.index))), 1)}"
+  protocol             = "${element(split(":", lookup(var.target_groups, element(keys(var.target_groups), count.index))), 0)}"
+  vpc_id               = "${var.vpc_id}"
+  deregistration_delay = 120
```
surminus: I feel like this is something that we may wish to change later on, so may be worth having a variable with a sensible default.

terraform/modules/aws/lb/main.tf
```diff
@@ -0,0 +1,198 @@
...
+
+  health_check {
```
surminus: I think most elements of the health check we would probably want to change depending on the type of application. I personally would set interval to a lower threshold for applications, so it may help in the long run to set variables with sensible defaults.

surminus:
I think this all makes sense, with comments mostly on how configurable we wish to make it.

I wonder if it could be extended by adding default alarms, or did we wish to keep these separate?

terraform/modules/aws/lb/main.tf
afda16: enable_deletion_protection is false by default, setting it to true prevents TF from removing the LB so probably we don't want to do that. I could remove it from the code

terraform/modules/aws/lb/main.tf
afda16: updated

terraform/modules/aws/lb/main.tf
afda16: updated

terraform/modules/aws/lb/main.tf
surminus: If it's false by default I think we should remove to avoid confusion.

afda16: updated

PR #369, opened 2017-11-20
Update lb modules and add apt example

afda16:
Update app-apt to use the lb module. Classic LBs are
migrated to ALBs.

https://trello.com/c/bKqwIbOi/840-refactor-elbs-to-lbs

### DIFF
--- a/terraform/modules/aws/lb/main.tf
+++ b/terraform/modules/aws/lb/main.tf
@@ -2,39 +2,17 @@
 * ## Modules: aws/lb
 *
 * This module creates a Load Balancer resource, with associated
-* listeners and target groups.
+* listeners and default target groups.
 *
-* Listeners and target groups are defined with map variables,
-* where the key is the name of the resource, and the value is
-* the port and protocol of the resource, specified as `PROTOCOL:PORT`.
-*
-* Listeners are associated to target groups with the `listener_target_groups`
-* map, that uses the name of the resources.
-*
-* When using a listener on HTTPS, we can specify the certificate with
-* the `listener_certificates` variable, where the value is the domain name
-* of the certificate as registered on AWS. The value is used by an
-* `aws_acm_certificate` data resource to find the certificate ARN.
+* The listeners and default actions are configured in the `listener_action`
+* map. The keys are the listeners PROTOCOL:PORT parameters, and the values
+* are the PROTOCOL:PORT parameters of the default target group of that listener.
 *
 *```
-* listeners = {
-*   "myapp-http-80"   = "HTTP:80"
-*   "myapp-https-443" = "HTTPS:443"
+* listener_action = {
+*   "HTTP:80"   = "HTTP:8080"
+*   "HTTPS:443" = "HTTP:8080"
 * }
-*
-* target_groups = {
-*   "http-80" = "HTTP:80"
-* }
-*
-* listener_target_groups = {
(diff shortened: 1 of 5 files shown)


terraform/projects/app-apt/main.tf
```diff
@@ -111,55 +78,38 @@ resource "aws_route53_record" "apt_external_service_record" {
...
+module "apt_internal_lb" {
+  source                  = "../../modules/aws/lb"
+  name                    = "${var.stackname}-apt-internal"
+  internal                = true
+  vpc_id                  = "${data.terraform_remote_state.infra_vpc.vpc_id}"
+  access_logs_bucket_name = "${data.terraform_remote_state.infra_aws_logging.aws_logging_bucket_id}"
+  listeners               = "${map("apt-http-80", "HTTP:80", "apt-https-443", "HTTPS:443")}"
+  listener_certificates   = "${map("apt-https-443", var.elb_internal_certname)}"
+  target_groups           = "${map("apt-http-80", "HTTP:80")}"
+  listener_target_groups  = "${map("apt-http-80", "apt-http-80", "apt-https-443", "apt-http-80")}"
```
surminus: I found this a little confusing to read, but I think I understand what's going on:

```
listener_target_groups {
  "apt-http-80" => "apt-http-80",
  "apt-https-443" => "apt-http-80"
}

I wonder if it would make it more readable by naming what the resource is? Something like:

`apt-http-80-listener`
`apt-http-80-target`

**Comment by surminus** [comment]
One comment but otherwise looks good.

File: terraform/projects/app-apt/main.tf (hunk 1 above)
Comment (afda16) [reply terraform/projects/app-apt/main.tf]: yes that sounds good, I'll update

**Comment by afda16** [comment]
Rebased and updated

**Comment by afda16** [comment]
WIP: trying to add routing rules

File: terraform/modules/aws/lb_listener_rules/main.tf [hunk 2]
```diff
@@ -0,0 +1,143 @@
...
+#--------------------------------------------------------------
+
+locals {
+  hosts = "${zipmap(formatlist("%s:%s", "host-header", keys(var.rules_host)), values(var.rules_host))}"
+  paths = "${zipmap(formatlist("%s:%s", "path-pattern", keys(var.rules_path)), values(var.rules_path))}"
+  rules = "${merge(var.rules_host_and_path, local.hosts, local.paths)}"
+}
+
+locals {
+  target_groups = "${distinct(values(local.rules))}"
```
surminus: Why is this outside the `locals` block above?

afda16: No specific reason, just keep the block next to the place where it's used so it's easy to understand what it's doing, that's what it looked to me

terraform/modules/aws/lb/main.tf
```diff
@@ -60,24 +33,27 @@ variable "access_logs_bucket_name" {
...
-  description = "A map of Load Balancer Listener certificate domain names."
+  description = "A map of Load Balancer Listener and default target group action, both specified as PROTOCOL:PORT."
 }
 
-variable "target_groups" {
-  type        = "map"
-  description = "A map of Load Balancer Target group resources."
+variable "listener_certificate_domain_name" {
+  type        = "string"
+  description = "HTTPS Listener certificate domain names"
```
surminus: This should be `name` singular.

PR #371, opened 2017-11-21
Add ASG notifications to node_group

afda16:
Add AutoScaling Group notifications to the node_group module.
This is enabled by default, and requires the SNS alerts topic ARN. We
need to update all the app-* projects with the `asg_notification_topic_arn`
parameter before merging this change.


surminus:
Will we update all the node_group's with `asg_notification_topic_arn` in this PR?

afda16:
yes probably better to do it in the same one, I'll add it

PR #375, opened 2017-11-22
Add infra security

afda16:


surminus:
What other things might go into this project in the future? I'm wondering if something like `infra-security` is a plainer, less prescriptive name than `harden`?

afda16:
Updated to infra-security

PR #376, opened 2017-11-23
Allow the monitoring machines access to Bouncer

issyl0:
- This "allow_management_to" block is true for every other project, and
  we do monitor Bouncer, so we should let the monitoring machines
  monitor it.


issyl0:
And TIL that management ≠ monitoring. I'll fix this tomorrow. 😴

PR #380, opened 2017-11-27
Add variable to disable Puppetmaster bootstrap ELB

surminus:
When browsing the puppetmaster manifest I thought it might easy enough to add in a variable which defaults to having no bootstrap ELB, but upon deployment we can set the var to true.

The story (referenced below) suggests that we should entirely extract the bootstrap ELB into it's own module. Personally this feels like overkill, and Terraform should be able to handle creating/destroy an ELB without issue. However, I am not aware of the context when the story was created so there may be something I'm not aware of.

https://trello.com/c/yDjFSzOX/652-aws-migration-refactor-puppetmaster-bootstrap


afda16:
I think the "aws_security_group_rule" "puppetmaster_from_elb_in_22"  resource needs the variable too, we only need that sg to enable ssh access from the puppetmaster elb during bootstrap

surminus:
Good point. I should probably appropriately update the bootstrapping docs, too.

PR #382, opened 2017-11-27
Allow deploy instance to write to artefact bucket

surminus:
Allow the deploy instance to write to the artefact bucket so it can make a copy of artefacts into their version of the "deploy-to-environment" branch. We use this branch to make assumptions about the version of the app that is deployed, and will use it to automatically deploy apps when nodes come up.

Router is pretty key to run on the cache instances, so we need to allow this to happen.

Relates to https://github.com/alphagov/govuk-app-deployment/commit/636b1ac253a71d80cd6a6af1b02afe0adbe1fe3a

See the [PR here](https://github.com/alphagov/govuk-app-deployment/pull/211).

Successful write [here](https://deploy.blue.integration.govuk.digital/job/Deploy_App/705/console), and successful subsequent deploy [here](https://deploy.blue.integration.govuk.digital/job/Deploy_App/706/console)


surminus:
I may look at updating the policy for this so that it is no longer public read.

I may also change the name of the project to `infra-artefacts` as being specific over what it is slightly annoys me.

afda16:
Should the bucket be part of the app-deploy project? Does the bucket need to be there before applying the app-* projects?

surminus:
No, it should be a separate entity because we'll probably only ever have a single artefact bucket. CI should only write to a single bucket, and we shouldn't have one per stack.

afda16:
Makes sense 👍

PR #391, opened 2017-12-01
Add a user data script to deploy apps

timmow:
Only run this script if the file exists, which should only be on app
servers

### DIFF
--- a/terraform/userdata/30-deploy-apps
+++ b/terraform/userdata/30-deploy-apps
@@ -1,0 +1,7 @@
+#
+# Snippet: deploy-apps
+#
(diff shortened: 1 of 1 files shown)


terraform/userdata/30-deploy-apps
```diff
@@ -0,0 +1,7 @@
+#
+# Snippet: deploy-apps
+#
+echo "[$(date '+%H:%M:%S %d-%m-%Y')] START SNIPPET: deploy-apps"
+test -f /usr/local/bin/govuk_sync_apps && /usr/local/bin/./govuk_sync_apps
```
deanwilson: What's the extra `/./` for the in the command execution?

timmow:
Good point, i've updated this now

PR #395, opened 2017-12-04
Lb fixes

afda16:


surminus:
> health_check needs both path and matcher. Matcher needs to include 404 because vhosts are always service specific and sometimes we are still providing a 404. We'll need to review this.

We should be able to find this out by using [what is set in Puppet](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/app.pp#L223).

PR #412, opened 2017-12-12
Add root DNS records for app services

afda16:
Add `.integration.govuk-internal.digital` CNAMES to blue stack for
app records.


surminus:
Can we go through this later? From my reading of it, if we deploy a project on a specific stack, it's going to update the top level domain (`.integration.govuk.digital`) to point to the thing that's just been deployed. Am I understanding this correctly, and if so, do we want it as part of the project or an independent thing?

I am also worried you asked me this question yesterday and I said the wrong thing from my limited understanding 😕

PR #413, opened 2017-12-13
Require the passwords to meet a minimum criteria

deanwilson:
This helps stop people using simple passwords and complies with the CIS
security docs (and scanners). We previously had everything apart from upper and special characters set so this is a little tighter.

This could have a param for each of those but I don't think we need that at this time.


surminus:
We're planning on using the central account to manage users in the future, so not sure if this is needed. See the [guide](https://docs.publishing.service.gov.uk/manual/user-management-in-aws.html).

deanwilson:
It'll still force things like the root user, which will have to be present, to have a decent password set. So less useful but a decent fall back.

PR #417, opened 2017-12-14
Use single keypair

deanwilson:


afda16:
Can you update the READMEs? There is a script in `tools` to help updating all the projects

PR #418, opened 2017-12-15
Add missing assets to internal DNS and make logs_cdn conditional

afda16:
The assets internal DNS entry is missing.

Logs_cdn is not deployed right now, so the ASG-LB attachment needs to be
conditional.


surminus:
If other things aren't deployed, will we see similar behaviour? I wonder if it's worth adding the `count` function across each attachment? 

I would argue we should be able to deploy this project even when a small amount of apps have been deployed.

PR #419, opened 2017-12-18
Add tool to update hosts

surminus:
I have written this with future proofing in mind, for new environments and if we rebuild the stack. It does a lookup of the current assigned DNS record and adds it to your hosts file.

### DIFF
--- a/tools/migration-hosts.sh
+++ b/tools/migration-hosts.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+#
+# A helper script to insert or remove migration domains from /etc/hosts
+DOMAINS=(
+alert
+assets-origin
+backend
+bouncer
+collections-publisher
+contacts-admin
+content-performance-manager
+content-tagger
+deploy
+docs
+grafana
+graphite
+hmrc-manuals-api
+imminence
+jumpbox
+kibana
+local-links-manager
+logs-cdn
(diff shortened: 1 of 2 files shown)


tools/migration-hosts.sh
```diff
@@ -0,0 +1,80 @@
...
+    IP=$(dig +short ${subdomain}.${LOOKUP} |tail -n1)
+    if [[ $IP != "" ]]; then
+      echo "${IP} ${subdomain}.${ENV}.publishing.service.gov.uk" | sudo tee -a /etc/hosts
+    else
+      echo "WARNING: No results for ${subdomain}"
+    fi
+  done
+elif  [[ $ACTION == 'del' ]]; then
+  for subdomain in ${DOMAINS[*]}; do
+    sudo sed -i .bak /${subdomain}.${ENV}.publishing.service.gov.uk/d /etc/hosts
```
timmow: Not that its really a problem, but this will create a new backup file for every iteration of the loop, that gets overwritten except the last iteration, which will leave a file containing all the hosts except the last one - think you could change it to `sed -i ''` to avoid this

surminus: That's a good point.

timmow: A file like your normal hosts file with one additional host in it

PR #425, opened 2017-12-28
Environment deployment

surminus:
Adds some configuration files to use with Terragov to deploy an
environment.

We do not have to merge this, but it's useful to have the branch when thinking about deployment order.


surminus:
Bah, my code is full of bugs. Ignore.

PR #432, opened 2018-01-02
Update Security Group rule names

surminus:
Update the names of security group rules so they match the convention detailed in the [ADR](https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0032-security-group-naming.md).

This also adds the tool to CI to ensure this is enforced.

https://trello.com/c/sO5UkqnX/896-audit-sg-rule-names

### DIFF
--- a/terraform/projects/infra-security-groups/monitoring.tf
+++ b/terraform/projects/infra-security-groups/monitoring.tf
@@ -21,7 +21,7 @@ resource "aws_security_group" "monitoring" {
   }
 }
 
-resource "aws_security_group_rule" "allow_monitoring_external_elb_in" {
+resource "aws_security_group_rule" "monitoring_ingress_monitoring-external-elb_http" {
   type      = "ingress"
   from_port = 80
   to_port   = 80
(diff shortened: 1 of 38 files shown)


terraform/projects/infra-security-groups/monitoring.tf
```diff
@@ -110,7 +109,8 @@ resource "aws_security_group_rule" "allow_management_to_monitoring_internal_elb_
   source_security_group_id = "${aws_security_group.management.id}"
 }
 
-resource "aws_security_group_rule" "allow_management_to_monitoring_internal_elb_https" {
+# TODO: Audit (is the rule correct?)
+resource "aws_security_group_rule" "monitoring-internal-elb_ingress_management_https" {
   type      = "ingress"
   from_port = 80
   to_port   = 443
```
afda16: I think this is 80 (in the monitoring-internal-elb resource the instance port is 80)

afda16:
I think we need to standardise the elb SG names, because we use +-internal/external randomly , I guess as we kept adding LBs. That would require changes in the app projects too, so probably something to update later.

PR #436, opened 2018-01-03
Use splat syntax to access attributes with counts

issyl0:
```
Warning: output "internal_root_zone_id": must use splat syntax to
access aws_route53_zone.internal_zone attribute "zone_id", because it
has "count" set; use aws_route53_zone.internal_zone.*.zone_id to obtain
a list of the attributes across all instances.
```

### DIFF
--- a/terraform/projects/infra-root-dns-zones/main.tf
+++ b/terraform/projects/infra-root-dns-zones/main.tf
@@ -94,21 +94,21 @@ resource "aws_route53_zone" "external_zone" {
 # --------------------------------------------------------------
 
 output "internal_root_zone_id" {
-  value       = "${aws_route53_zone.internal_zone.zone_id}"
+  value       = "${join("", aws_route53_zone.internal_zone.*.zone_id)}"
   description = "Route53 Internal Root Zone ID"
 }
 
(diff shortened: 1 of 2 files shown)


terraform/projects/infra-root-dns-zones/main.tf
```diff
@@ -94,21 +94,21 @@ resource "aws_route53_zone" "external_zone" {
 # --------------------------------------------------------------
 
 output "internal_root_zone_id" {
-  value       = "${aws_route53_zone.internal_zone.zone_id}"
+  value       = "${aws_route53_zone.internal_zone.*.zone_id}"
```
afda16: the internal and external zones are created only when 'create_*_zone' is set to true, so there could be a case when the resource is not created and the output would error. Same as the rds_module, I think it's better to do something like this in the outputs

```
  value = "${join("", aws_route53_zone.internal_zone.*.zone_id)}"
```

issyl0: Should I do this for every change here?

issyl0: @afda16 Fixed, I hope! Thanks.

PR #442, opened 2018-01-04
Add cloudtrail

afda16:
CloudTrail settings and alarms


surminus:
I think this all looks sensible, but can we add detail on why we think we need these alarms, even if they're based on security recommendations? It might be useful to also link to resources if possible.

I'm wary that we're adding a lot of code with very little detail and a lot of assumed knowledge, mostly around 8109dc5.

afda16:
Added more info to the commit message

PR #444, opened 2018-01-08
Resize graphite machine type

afda16:
Update machine type to m5.xlarge. It should be cheaper, we tried to
resize it before but it didn't work because of disk labels. Trying again
after the 'Device' tag was added to the EBS volume.


surminus:
Related PR https://github.com/alphagov/govuk-aws/pull/428

issyl0:
Oh. I thought that was covered by Ana's comment "Trying again
after the 'Device' tag was added to the EBS volume." Sorry.

surminus:
My fault, I was only half paying attention. Sorry.

surminus:
I found the related commit: https://github.com/alphagov/govuk-aws/commit/87d07031e3b1017df1cddbee282291d22276797c - if we can include this in the commit message for context that would be good.

PR #451, opened 2018-01-09
Allow access from the Deploy Jenkins to Graphite

issyl0:


surminus:
Graphite only currently listens on https: https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/app-graphite/main.tf#L156

Ideally we should get the notification to happen over https.

issyl0:
Presumably that would involve changing Carrenza too, given the port is hardcoded in https://github.com/alphagov/govuk-app-deployment/blob/master/recipes/notify.rb#L76. Unless we do another envvar for `GRAPHITE_PORT`.

terraform/projects/infra-security-groups/graphite.tf
```diff
@@ -49,6 +49,19 @@ resource "aws_security_group_rule" "graphite_ingress_graphite-internal-elb_http"
   source_security_group_id = "${aws_security_group.graphite_internal_elb.id}"
 }
 
+resource "aws_security_group_rule" "graphite-internal-elb_ingress_deploy_http" {
```
afda16: Can we put this rule next to 

```
resource "aws_security_group_rule" "graphite-internal-elb_ingress_monitoring_https" {
```

It think it's easier to read if they are group by the source

surminus:
Yeah, that's not ideal. We either do an env var, or we need to get the ELB to listen on tcp/80. 

I have no strong opinion either way.

issyl0:
My preference is an envvar, given making this ELB listen on port 80 would be inconsistent with the other ELBs. I'll do that.

issyl0:
Envvar PR: https://github.com/alphagov/govuk-puppet/pull/7049.

issyl0:
@afda16 I moved the rule down to be with the `monitoring` one.

@surminus I updated the port to be 443, ahead of the Puppet and app-deployment PRs.

issyl0:
I did a `plan` and an `apply`, so merging.

PR #455, opened 2018-01-15
Create backend files for the staging environment

issyl0:


surminus:
I'm publicly disclaiming my dislike of the `govuk-terraform-steppingstone-staging` name (but don't really care in all seriousness).

Are we setting `eu-west-2` to build in London specifically?

issyl0:
Oh! I meant Ireland, eu-west-2. I'll update them.

issyl0:
I dislike the name too, but that's what we picked for integration, and `govuk-terraform-staging` is very broad. Though, `steppingstone` for what...? 🤔

surminus:
It's passive aggressive naming because we were told it was "Project Stepping Stone" until we migrate GOV.UK to the PaaS, which is why I'd prefer to call it something dull. 

`govuk-terraform-state-staging` is probably OK.

issyl0:
There's already a bucket `govuk-terraform-state-integration`, so I'll go with that naming for staging too.

PR #456, opened 2018-01-16
Allow reads to database buckets from the office

surminus:
Developers traditionally synced database data to the dev VM by SSHing into backup-1 and fetching backups. In AWS, there is no backup-1, but we put backups into an S3 bucket.

This allows access from the office to fetch backups.

https://trello.com/c/FES1jMky/898-ensure-the-dev-vm-can-retrieve-backups-from-the-s3-bucket


terraform/projects/infra-database-backups-bucket/main.tf
```diff
@@ -34,6 +34,18 @@ variable "remote_state_infra_monitoring_key_stack" {
...
+variable "allow_read_from_office" {
+  type        = "string"
+  description = "Set to true to allow read access to the bucket from the office."
+  default     = false
+}
+
+variable "office_ips" {
+  type        = "list"
+  description = "A list of the office IPs"
+  #default     = []
```
issyl0: This comment probably doesn't need to be here? `office_ips` is defined in https://github.com/alphagov/govuk-aws-data/blob/5716300dfe66dcada0efdcd344bbae5ca9cda4c1/data/common/integration/common.tfvars#L11-L18 so should never be blank?

surminus: Yep, you're right.

surminus:
@issyl0 updated.

issyl0:
CI is complaining on the `terraform fmt` step, about misaligned `=` and lacking trailing commas.

PR #469, opened 2018-01-29
Update env provisioning notes

surminus:
This updates the guide to give a semblance of order rather than just "deploy the rest of the stack".


surminus:
That's not an official deployment tool so it's good to note it down 😉

PR #470, opened 2018-01-29
Seperate internal services from public services

surminus:
The `infra-public-services` project contained two different purposes:

1. Create top level internal DNS records that the application level infrastructure is entirely dependent on. For example, if an application wishes to speak to the current live Redis instance, it will query the host "redis" which will resolve the currently live (green or blue) load balancer for this service.

2. Public facing load balancers that resolve the publishing domain, contain the public SSL certifcate, which are attached to the relevant live autoscaling group, and then have a external DNS record assigned to it. This is something that needs to be run when everything else is live so that we can serve the infrastructure off the publishing API.

This splits out this project into two defined projects. It ensures that the infra-internal-services can be updated and run without having to touch or create any public load balancers. This is essentially just creating or amending DNS records so should be a non-intrusive change.

### How to migrate the state

I suspect the best way to migrate the state from public-infra-services would be to just wait until a "maintenance period" and run an apply on both the projects. This would destroy and then re-create the internal DNS records, but if done at an appropriate time wouldn't be too much of a blocker to the current environment.


issyl0:
Is this mergeable?

surminus:
This requires a review from at least 2 people as it is a large change.

surminus:
Due to lack of review I am going to close this. I will keep the branch active incase someone wishes to revisit this "moving forward".

PR #471, opened 2018-01-29
[WIP] Add environment common data

surminus:
Update the build tool to allow the creation of environment common data: data which is common across all environments, which should be stored in `data/app-awesome-thing/common.tfvars`.

As a specific implementation this should *not* include secrets because these should be different for each environment, and this should be enforced.


surminus:
I will further enhance this, and then add an example to govuk-aws-data when I am finished.

surminus:
I think this can be closed 😂

PR #472, opened 2018-01-30
Add script for Jenkins deploy

surminus:
This script invokes the `build-terraform-project.sh` script and should be triggered using Jenkins.

Sops is a precursor as an install to run these scripts, and should be managed by Puppet as it is less liable to change than Terraform.

If the Terraform version is not specified using TERRAFORM_VERSION, then it will use the default installed on the system where it is being run.

If you wish to test newer versions of Terraform (quite a common occurrence at this point in the Terraform lifecycle), then you can specify a version and the script will download, verify and set the path for that version of Terraform.

https://trello.com/c/qAzTAU4p/952-build-box-spike-on-adding-tf-deploys-to-current-integration-carrenza-deploy-jenkins

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+#
+# This script invokes the build-terraform-project.sh tool to deploy the code.
+#
+if [[ ! $(which sops) ]]; then
+  echo "sops not installed, exiting"
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


surminus:
I just thought that for every job it will download Terraform, which makes it independent, but isn't great and won't be cached anywhere, so it may be better to just do it with packages anyway.

PR #473, opened 2018-01-30
Upgrade to 0.11.2

surminus:


surminus:
Maybe we should test it first.

surminus:
Tested and everything seems OK.

PR #475, opened 2018-01-31
Add poweruser role

dwhenry:
work in combination with: https://github.com/alphagov/govuk-aws-data/pull/85

https://trello.com/c/h7CNr5H5/51-dev-vm-replication


surminus:
Looks good to me, thanks for updating the README too!

PR #477, opened 2018-02-01
Give access to s3 DB backups for govuk-users

dwhenry:
This attempts to give access to the s3 bucket for `govuk-users`, my assumption is that other users who currently have access won't lose it. 

https://trello.com/c/h7CNr5H5/51-dev-vm-replication

### DIFF
--- a/terraform/projects/infra-security/main.tf
+++ b/terraform/projects/infra-security/main.tf
@@ -64,6 +64,11 @@ variable "ssh_public_key" {
   description = "The public part of an SSH keypair"
 }
 
+variable "database_backup_access_enabled" {
+  type        = "string"
+  description = "Whether or not database backup access is allowed"
+}
+
 # Resources
 # --------------------------------------------------------------
 
@@ -222,4 +227,21 @@ module "cloudtrail-alarm-root-account" {
   alarm_name                = "${var.stackname}-cloudtrail-root-event"
   alarm_description         = "Alarms when the root account is used."
   alarm_actions             = ["${aws_sns_topic.cloudtrail.arn}"]
+}
+
+# Database backup access for Dev VM
+
+resource "aws_iam_policy" "aws-database-backup_dev-read_iam_policy" {
+  count       = "${var.database_backup_access_enabled}"
+  name        = "aws-database-backup_dev-read_iam_policy"
+  path        = "/"
+  description = "Allow read access to S3 aws-database-backup bucket"
+  policy      = "${file("${path.module}/../../policies/iam_s3_aws_database_backup_read_policy.tpl")}"
+}
+
+resource "aws_iam_policy_attachment" "aws-database-backup_dev-read_iam_policy_attachment" {
+  count      = "${var.database_backup_access_enabled}"
+  name       = "aws-database-backup_dev-read_iam_policy_attachment"
+  roles      = ["${module.role_poweruser.arn}"]
+  policy_arn = "${aws_iam_policy.aws-database-backup_dev-read_iam_policy.arn}"
 }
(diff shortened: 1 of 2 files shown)


terraform/projects/infra-database-backups-bucket/main.tf
```diff
@@ -134,3 +134,25 @@ resource "aws_s3_bucket" "database_backups" {
...
+
+# Create a read user to allow ingestion of logs from the bucket
+resource "aws_iam_policy" "aws-database-backup_dev-read_iam_policy" {
+  name        = "aws-database-backup_dev-read_iam_policy"
+  path        = "/"
+  description = "Allow read access to S3 aws-database-backup bucket"
+  policy      = "${data.template_file.iam_aws_database-backup_dev_read_policy_template.rendered}"
+}
+
+resource "aws_iam_role" "aws-database-backup_dev-read_iam_role" {
```
surminus: I think this role is already created in `infra-security`, so this may not work and we'll have to do some data shuffles. I'll wait for you to come up and we can pair on it.

afda16: Already managed here: https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/infra-security/main.tf#L94

terraform/projects/infra-database-backups-bucket/main.tf
```diff
@@ -134,3 +134,25 @@ resource "aws_s3_bucket" "database_backups" {
     }
   }
 }
+
+data "template_file" "iam_aws_database_backup_dev_read_policy_template" {
```
afda16: It doesn't need to be a template, nothing to interpolate. You can do file() in the aws_iam_policy below

afda16: Actually ... You want to interpolate the environment in the name. If we don't want to grant this group access to the backups on all the environments, we need more logic

terraform/projects/infra-database-backups-bucket/main.tf
afda16: Probably better keep infra-security standalone, we can use this to avoid remote state file: https://www.terraform.io/docs/providers/aws/d/iam_role.html

dwhenry: Moved this to infra-security to avoid dependancy and make use of existing role object

dwhenry: as discussed

terraform/projects/infra-database-backups-bucket/main.tf
dwhenry: done

terraform/projects/infra-security/main.tf
```diff
@@ -223,3 +228,20 @@ module "cloudtrail-alarm-root-account" {
...
+
+resource "aws_iam_policy" "aws-database-backup_dev-read_iam_policy" {
+  count       = ${var.database_backup_access_enabled}
+  name        = "aws-database-backup_dev-read_iam_policy"
+  path        = "/"
+  description = "Allow read access to S3 aws-database-backup bucket"
+  policy      = "${file("${path.module}/../../policies/iam_s3_aws_database_backup_read_policy.tpl")}"
+}
+
+resource "aws_iam_policy_attachment" "aws-database-backup_dev-read_iam_policy_attachment" {
```
afda16: the role module supports an array of policy ARNs with the `role_policy_arns` parameter, maybe this is not needed here and you can add `aws_iam_policy.aws-database-backup_dev-read_iam_policy.arn` to the array. 

Maybe it's easier to relate things if we move the policy next to the module

afda16:
We tested the current permissions for GOVUK users, this wouldn't be needed

PR #478, opened 2018-02-02
Add new govuk wrapper tool

surminus:
This is to help developers SSH to the environment, and also remove the older govuk-ssh tool which has a proxy script which can also be removed from Puppet.

I've written it so it can be expanded for other uses in the future.

Originally we wanted to write something in Ruby as a gem and have interactions with the AWS API, but didn't get a chance to do it.


doc/guides/connecting-to-the-environment.md
```diff
@@ -19,26 +19,29 @@ SSH to the hostname:
...
-access to machines in our environments without long lists of SSH config.
+The `govukcli` is a script which can help you login to the environment.
 
 ### Install
 
-The script is [located here](tools/govuk-ssh).
-
-Copy the script to `/usr/local/bin/govuk-ssh` and make executable.
+```
+wget https://raw.githubusercontent.com/alphagov/govuk-aws/master/tools/govukcli
```
boffbowsh: I think you made a typo. This should be:

```
curl url | sudo bash
```

boffbowsh:
This is really useful for now, but I think having to set the different env for each environment might be unwieldy in the future. I think 
```bash
govukcli ssh int backend
```
would be really cool. Especially if we can have shorthand of things like int, prd, prod, stg, stag etc.

surminus:
Yeah, agree the env var might be a pain, but having to type the environment each time could also be annoying. Ideally we could do `govukcli set-context int` that reads from a config file or something. I'll have a ponder.

surminus:
I updated it (in a hurry), but the usage would be to set contexts:
```
🐣  ~ govukcli set-context integration
🐣  ~ govukcli get-context
integration
🐣  ~ govukcli ssh backend
```

This means it doesn't have to be specified each time and can be changed easily. We could probably update different SSH config for the current Staging and Production if we wanted to.

surminus:
Also to list:
```
🐣  ~ govukcli list-contexts
integration
production
staging
```

PR #479, opened 2018-02-06
[WIP] Add buckets for storing data for model training

cbaines:
In relation to the Topic Taxonomy on GOV.UK.

WIP, as I have no idea if this works, or how to test it yet.


afda16:
It's missing a backend file, generated documentation, and remote states manifest. I think the name of the project should be app-<>

More info here https://github.com/alphagov/govuk-aws/blob/master/doc/guides/getting-started.md, and the style guide linked in that document

issyl0:
Is this still needed? I suspect not?

PR #483, opened 2018-02-12
Introduced new variable asg_size

rtrinque:
New variable asg_size to store the same value for asg_max_size, asg_min_size, and asg_desired_size

### DIFF
--- a/terraform/projects/app-backend/main.tf
+++ b/terraform/projects/app-backend/main.tf
@@ -41,6 +41,12 @@ variable "app_service_records" {
   default     = []
 }
 
+variable "asg_size" {
+  type        = "string"
+  description = "The autoscaling groups desired/max/min capacity"
+  default     = "2"
+}
+
 # Resources
 # --------------------------------------------------------------
 terraform {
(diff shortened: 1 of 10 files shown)


terraform/modules/aws/node_group/main.tf
```diff
@@ -115,6 +115,12 @@ variable "instance_target_group_arns_length" {
   default     = 0
 }
 
+variable "asg_size" {
```
surminus: This `variable` block will need to live inside of each project rather than the module itself, so at the top of `terraform/projects/app-backend/main.tf`, for example. No logic inside the module is using this variable.

terraform/projects/app-backend/main.tf
```diff
@@ -191,9 +197,10 @@ module "backend" {
   instance_elb_ids_length       = "2"
   instance_elb_ids              = ["${aws_elb.backend_elb_internal.id}", "${aws_elb.backend_elb_external.id}"]
   instance_ami_filter_name      = "${var.instance_ami_filter_name}"
-  asg_max_size                  = "2"
-  asg_min_size                  = "2"
-  asg_desired_capacity          = "2"
+  asg_size                      = "2"
```
surminus: This doesn't need to be here. The module only accepts the three variables below ([ref](https://github.com/alphagov/govuk-aws/blob/master/terraform/modules/aws/node_group/README.md)), but we can still just pass in the same value as defined in the `asg_size` variable.

terraform/projects/app-backend/main.tf
```diff
@@ -41,6 +41,12 @@ variable "app_service_records" {
   default     = []
 }
 
+variable "asg_size" {
+  type        = "string"
+  description = "The autoscaling groups desired/max/min capacity"
+  default     = "1"
```
surminus: Sorry just noticed, we probably want the default to be `2` here as we do not override this variable for Integration.

surminus:
One further comment, but otherwise this is correct.

surminus:
Just need to squash the commits into one, and then this is good to go!

surminus:
Tested successfully with app-backend.

PR #486, opened 2018-02-14
Allow option to list node-types

surminus:
Someone suggested a way to list node types. In Carrenza, this is tricky, but since I built in the option to SSH straight to specific machines it's less valuable.

In AWS, we can use the tags assigned to instances, so using a bit of jq magic, we can list all the types of instances available to SSH to.

### DIFF
--- a/tools/govukcli
+++ b/tools/govukcli
@@ -55,11 +55,14 @@ function ssh_usage {
 
   Options:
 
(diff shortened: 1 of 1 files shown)


surminus:
It uses `jq` on the jumpbox within the environment, and every node has `jq` installed by default, so shouldn't be an issue.

PR #488, opened 2018-02-19
Graphite backup

rtrinque:
Creating an S3 bucket and the policy and role to receive the backups from graphite

https://trello.com/c/H9rrNIZc/987-graphite-backups-to-s3-directly-with-whisper-backup-or-similar

### DIFF
--- a/terraform/projects/app-graphite/main.tf
+++ b/terraform/projects/app-graphite/main.tf
@@ -40,6 +40,12 @@ variable "elb_internal_certname" {
   description = "The ACM cert domain name to find the ARN of"
 }
 
+variable "remote_state_infra_database_backups_bucket_key_stack" {
+  type        = "string"
+  description = "Override stackname path to infra_database_backups_bucket remote state"
+  default     = ""
+}
+
 # Resources
 # --------------------------------------------------------------
 terraform {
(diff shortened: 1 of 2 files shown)


surminus:
I wonder if we should have a separate bucket for graphite backups. We could re-use the "database-backups-bucket" (I have an issue with that name, I don't think we should specify the resource in the project name, but anyway...)

Having another separate bucket means we have an additional project to run and more code to manage. I'm not sure what the advantage of a separate bucket is when we have one that already stores our backups. I would check with @issyl0 and @stephenharker. 

I haven't looked into how graphite-backup works, but we should also consider what lifecycle settings we may want to set. 

This should be a single commit.

rtrinque:
Change of plan, I just attach the writing policy on graphite to the already existing database backups bucket now

surminus:
You will need to load the remote state for infra-database-backups-bucket project ([example](https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/app-mongo/main.tf#L280-L288)). We often do this by default in the `remote_state.tf` files, but we do not do this for that project because it's only used in a few places.

We may also want to think about the [lifecycle settings](https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/infra-database-backups-bucket/main.tf#L90-L107) as Graphite produces a lot of data, but we may need to think about how long we want to keep it for.

PR #490, opened 2018-02-20
AWS SNS Topic and Lambda Function

suthagarht:
- We have a necessity to create an AWS SNS Topic that can be used to
  communicate SNS notifications from the govuk-{environment}-artefact
buckets.

- We also need a Lambda function to synchronize buckets between
  environments.

The two points mentioned above has been covered via this commit.

Solo: @suthagarht


suthagarht:
Test

PR #494, opened 2018-02-21
Add new machines for email-alert-api

rubenarakelyan:
This commit adds new machines for email-alert-api to AWS.

We're adding these new machines since email-alert-api is using the majority of resources on the current backend machines and we'd like to give it space as well as reduce any possible impact on other backend apps. This is a result of the work to migrate all email sending to our own systems.

Trello: https://trello.com/c/oqi6Xvnu/629-move-email-alert-api-and-email-alert-service-to-dedicated-vms


surminus:
I think this looks good at first glance - does it require a public facing load balancer ie are there any external facing services? Will try and go through it in detail shortly.

rubenarakelyan:
@surminus Yes, there is a public-facing API endpoint (currently exposed as `email-alert-api-public`) - does this need some extra AWS config?

surminus:
Nope, I was checking the external load balancer was necessary!

terraform/projects/infra-security-groups/email-alert-api.tf
```diff
@@ -0,0 +1,114 @@
...
+  vpc_id      = "${data.terraform_remote_state.infra_vpc.vpc_id}"
+  description = "Access the email-alert-api ELB"
+
+  tags {
+    Name = "${var.stackname}_email-alert-api_elb_access"
+  }
+}
+
+# TODO: Audit
+resource "aws_security_group_rule" "email-alert-api-elb-internal_ingress_management_https" {
```
surminus: Do we know what instances need access to email-alert-api? It would be good not to open up everything up if we can decide at this stage what needs access.

surminus:
Other than one minor comment I think this looks good!

terraform/projects/infra-security-groups/email-alert-api.tf
rubenarakelyan: The ones I can think of are `backend` and `publishing-api` at the moment...

rubenarakelyan: Also `frontend`, and of course we have the publicly-accessible API endpoint too. Does that mean we have to allow all access?

surminus: If we can open up the internal load balancer to specific instances that would be useful, and then we have the external facing load balancer for the API endpoint.

rubenarakelyan: Which rule is it specifically that I'm looking at here - Terraform fu is weak for me!

surminus:
I think this is fine. To apply, you will need to deploy `infra-security-groups`, and then `app-email-alert-api`.

PR #498, opened 2018-02-23
Only ask for MFA on assume-role once in 12 hours

issyl0:
- Previously, every time we called "aws assume-role" on the command
  line, we'd have to get our phones (or other designated MFA solution)
  and type in a code. This would have to happen every hour, which was
  tedious for long-running builds like building an environment.
- With this change, AWS won't require an MFA token upon assuming a role
  if `sts get-session-token` with an MFA code has been successful within
  the last 12 hours (43200 seconds).
- This doesn't remove the requirement to refresh your assume-role tokens
  every hour, as that's a different AWS limitation.


issyl0:
Successfully `apply`ed all the way to production. Merging.

PR #499, opened 2018-02-27
Deploy app artefact sync

suthagarht:
# Necessity
We have an existing process where a Jenkins job called **Deploy_Apps** creates a Go language binary file under an S3 bucket for the relevant environment. For example **govuk-integration-artefact**. When we build another environment, for example **Staging**, then that binary file should be available in the environment specific bucket for the new build. For staging , it will be  **govuk-staging-artefact**. Hence, we need a mechanism to promote the binary file accross environments.

# Concept
We create a mechanism that will be aware when a new file gets created.
1. We will create an **SNS topic** in the environment where the binary file will be created.
2. We create an **S3 bucket notification event** and associate it to the bucket that will get the new file.
3. The S3 bucket event will be associated to the **SNS topic**.

We create a mechanism that can know when the new file is created and retrieve the file.
1. We create a **Lambda** function that can copy a file from another **S3** bucket.
2. We create an **SNS Subscription** that subscribes to the **SNS Topic** at the source.
3. We associate the **SNS Subscription** to the **Lambda** function. 

When a new file gets created, the following will happen.(We will use **Integration**(INT) as source and **Staging**(STG) as destination)
New file -> INT SNS Topic notified -> STG Subscription alerted -> STG Lambda function syncs the file

# Working Components

## Lambda Files
The lambda related files are stored in **terraform/lambda/ArtefactSync**.

## Terraform Resources
There are number of terraform resources that work together to complete the functionality.

### govuk-artefact-bucket-policy
This resource assigns an S3 bucket policy. The policy will restrict access and allow the AWS Integration and Staging account access to the artefact bucket.

### artefact_topic
This resource creates an AWS SNS Topic in the source environment.

### artefact_topic_policy and artefact_sns_topic_policy
**artefact_sns_topic_policy**  is the actual policy that give Lambda related permissions to interact with AWS SNS. **artefact_topic_policy** acts like a policy attachment to the SNS topic.

### artefact_bucket_notification
This resource links the **S3 Bucket Event** to the **SNS Topic**.

### artefact_topic_subscription
This resource is the **SNS Subscription** at the target location. Also, this resource will associate to the **Lambda Function** at the source location.

### artefact_lambda_function
This resource creates the Lambda function and connects to the relevent IAM role.

### govuk_artefact_lambda_role
This resource allows the Lambda function to assume the relevant IAM role.

### govuk_artefact_policy
This resource will create an IAM policy. This policy will allow logs to be written to CloudWatch, during the Lambda function's activity.

### govuk_artefact_policy_attachment
This resource is used to attach the **IAM Role** and **IAM Policy** used by the Lambda function.

### govuk_artefact_lambda_sns
This is the resource that allows the **SNS Subscription** to trigger the **Lambda Function** and cause the execution.

https://trello.com/c/xpd7o4oU/980-router-deployed-to-staging-aws-iam-s3

Solo: @suthagarht


surminus:
Without looking at what has changed, this has created quite a few resources without any context on the changes themselves. Please can you write a more detailed commit message so someone in the future has an understanding on how it works, and importantly why it has been implemented in this way.

surminus:
Great detail! Please can you update the commit message itself rather than just the PR. Commit history is long term, whereas PR comments should provide a summary of the commits themselves (obligatory [link to blog post](https://gdstechnology.blog.gov.uk/2016/05/31/how-we-use-git-at-the-government-digital-service/)). 

I think the "Necessity" and "Concept" sections are only really required for a commit message, as you detail the stages of the process. People should hopefully be able to understand the resources and how they are used by looking at the code itself. 

Otherwise excellent detail!

terraform/projects/infra-artefact-bucket/main.tf
```diff
@@ -31,6 +31,31 @@ variable "aws_environment" {
...
+  type        = "string"
+  description = "The AWS Account ID"
+}
+
+variable "aws_subscription_account_id" {
+  type        = "string"
+  description = "The AWS Account ID that will appear on the subscription"
+}
+
+variable "create_sns_topic" {
```
surminus: does this also need a `type`?

terraform/projects/infra-artefact-bucket/main.tf
```diff
@@ -118,6 +143,192 @@ resource "aws_s3_bucket" "artefact" {
...
+{
+    "Version": "2012-10-17",
+    "Id": "ArtefactBucketPolicy1519740678",
+    "Statement": [
+        {
+            "Sid": "Stmt1519740678001",
+            "Effect": "Allow",
+            "Principal": {
+                "AWS": [
+                    "arn:aws:iam::210287912431:root",
```
surminus: Please remove these and add them as variables. This is a public repository.

terraform/projects/infra-artefact-bucket/main.tf
```diff
@@ -118,6 +143,192 @@ resource "aws_s3_bucket" "artefact" {
   }
 }
 
+# AWS S3 Bucket policy for Lambda access
+resource "aws_s3_bucket_policy" "govuk-artefact-bucket-policy" {
+  bucket = "${aws_s3_bucket.artefact.id}"
+
+  policy = <<POLICY
```
surminus: Unless there is a good reason for this to be inline, we should stick with convention and use `template_file`.

terraform/projects/infra-artefact-bucket/main.tf
```diff
@@ -118,6 +143,192 @@ resource "aws_s3_bucket" "artefact" {
...
+  handler       = "main.lambda_handler"
+  runtime       = "python2.7"
+}
+
+# AWS Lambda Role
+resource "aws_iam_role" "govuk_artefact_lambda_role" {
+  count = "${var.create_sns_subscription ? 1 : 0}"
+  name  = "govuk_artefact_lambda_role"
+
+  assume_role_policy = <<EOF
```
surminus: Unless there is a good reason for this to be inline, we should stick with convention and use template_file.

terraform/projects/infra-artefact-bucket/main.tf
```diff
@@ -118,6 +143,192 @@ resource "aws_s3_bucket" "artefact" {
...
+EOF
+}
+
+# AWS SNS IAM Policy
+resource "aws_iam_policy" "govuk_artefact_policy" {
+  count       = "${var.create_sns_subscription ? 1 : 0}"
+  name        = "govuk-artefact-policy"
+  description = "Provides necessary access to the Lambda function."
+
+  policy = <<EOF
```
surminus: Unless there is a good reason for this to be inline, we should stick with convention and use `template_file`.

surminus:
Please can we also ensure we update the README as per [the styleguide](https://github.com/alphagov/govuk-aws/blob/master/doc/guides/styleguide.md).

terraform/projects/infra-artefact-bucket/main.tf
suthagarht: create_sns_topic - This is a boolean. I have set a default value as false.  I believe this would suffice. https://www.terraform.io/docs/configuration/variables.html

terraform/projects/infra-artefact-bucket/main.tf
suthagarht: Good point. Done.

terraform/projects/infra-artefact-bucket/main.tf
surminus: In Terraform there is no real Boolean type (they are converted to strings of either `0` or `1`). I think you're correct that it does work, but to have consistent code we should consider specifying the type for each variable.

terraform/projects/infra-artefact-bucket/main.tf
suthagarht: agreed, done.

terraform/projects/infra-artefact-bucket/main.tf
suthagarht: agreed, done.

terraform/projects/infra-artefact-bucket/main.tf
suthagarht: you are correct. I didn't read that attentively and used as boolean.  I have now set the variable type to string.

surminus:
I noticed you've updated the commit message, so thank you, but please can we ensure the title of the commit is descriptive? At the moment it is titled "Necessity" which doesn't help someone looking through the commit history.

If someone wanted to check the last 10 commits for example then it's useful to briefly understand the context of the commit:

```
🐮  ~/govuk/govuk-aws git:(master) git log --oneline |grep -v Merge |head
672b47e Remove email-alert-api CNAME
c9167c0 Add verbose error for no backend file
6884b1d Only ask for MFA on assume-role once in 12 hours
442bd30 Add new machines for email-alert-api
ccb9084 Add note about Puppet librarian
2ef01a3 Fix build script
a0edeb6 Increase backend-redis to cache.m3.large
ec2d8f6 Allow backend machines access to graphite
04783c2 Fix if statement syntax in the jenkins.sh that runs Terraform
83a29b3 Git clone using SSH
```

A good title could be something like "Create method to promote artefacts between envs", for example.

I would also hesitate to list all the different resource types as hopefully the code should document their functions well enough.

suthagarht:
hi @surminus . Thank you for checking. I have now included a relevant heading.

surminus:
Thanks for adding the note! It might be worth a review from someone else to get context on the change as well.

PR #500, opened 2018-02-28
Add verbose error for no backend file

surminus:
When a backend file is missing, a misleading error is shown. We should produce an error that guides the user to try the correct thing. This should produce a list of possible stacks to deploy to if the backend file does not exist for that stack in the specified environment.

### DIFF
--- a/tools/build-terraform-project.sh
+++ b/tools/build-terraform-project.sh
@@ -128,6 +128,12 @@ cd "$PROJECT_DIR"
 
 # Actually run the command
 function init() {
+  if [[ ! -f $BACKEND_FILE ]]; then
+    echo "Could not find backend file for $STACKNAME stack."
+    echo "Possible stacks to deploy in $ENVIRONMENT for $PROJECT are: "
+    find . -name "${ENVIRONMENT}.*.backend" |cut -d "." -f3
+    exit 1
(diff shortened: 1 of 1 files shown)


tools/build-terraform-project.sh
```diff
@@ -130,6 +130,12 @@ cd "$PROJECT_DIR"
 function init() {
   rm -rf .terraform && \
   rm -rf terraform.tfstate.backup && \
+  if [[ ! -f $BACKEND_FILE ]]; then
+    echo "Could not find backend file for $STACKNAME stack."
+    echo "Possible stacks to deploy in $ENVIRONMENT for $PROJECT are: "
+    find . -name "integration.*.backend" |cut -d "." -f3
```
deanwilson: Should this literal (`integration`) be there?

surminus:
I tested the function:

```
➜  ~/govuk/govuk-aws git:(add-verbose-error-for-wrong-stack) tools/build-terraform-project.sh -c plan -p infra-security-groups -e staging -s blue -d ../govuk-aws-data/data
data dir is  ../govuk-aws-data/data
Could not find backend file for blue stack.
Possible stacks to deploy in staging for infra-security-groups are:
govuk
```

PR #502, opened 2018-03-01
Add_mongo_ebs

jljalderman:
Add AWS EBS volume to mongo instances and reduce the root disk size.
This needs the corresponding change in govuk-puppet to create the mount etc.

### DIFF
--- a/terraform/projects/app-mongo/main.tf
+++ b/terraform/projects/app-mongo/main.tf
@@ -125,7 +125,26 @@ module "mongo-1" {
   instance_elb_ids              = []
   instance_ami_filter_name      = "${var.instance_ami_filter_name}"
   asg_notification_topic_arn    = "${data.terraform_remote_state.infra_monitoring.sns_topic_autoscaling_group_events_arn}"
-  root_block_device_volume_size = "150"
+  root_block_device_volume_size = "50"
+}
+
+resource "aws_volume_attachment" "external_volume_attachment" {
+  device_name = "/dev/sdf"
+  volume_id   = "${aws_ebs_volume.ebs-mongo-1.id}"
+  instance_id = "${aws_instance.mongo-1_eni.id}"
+}
+
+resource "aws_ebs_volume" "ebs-mongo-1" {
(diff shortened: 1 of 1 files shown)


issyl0:
Looks like you need to line up some `=` in your Terraform to appease the linter.

jljalderman:
done

terraform/projects/app-mongo/main.tf
```diff
@@ -205,7 +243,26 @@ module "mongo-3" {
...
+
+resource "aws_volume_attachment" "external_volume_attachment" {
+  device_name = "/dev/sdf"
+  volume_id   = "${aws_ebs_volume.ebs-mongo-3.id}"
+  instance_id = "${aws_instance.mongo-3_eni.id}"
+}
+
+resource "aws_ebs_volume" "ebs-mongo-3" {
+  availability_zone = "${lookup(data.terraform_remote_state.infra_networking.private_subnet_names_azs_map, var.mongo_3_subnet)}"
+  type              = "gp2"
```
surminus: I wonder if it's worth considering `io1` type as we're mounting a database off it? Additional cost, but better performance.

surminus:
The block to add the volumes seems different to the way we have mounted them previously in the code - we do it in a few other places ([example](https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/app-logs-cdn/main.tf#L124-L149)). We create the mount with specific tags, ensure the instance is able to be able to query those tags, and then attach and mount the volume with [userdata](https://github.com/alphagov/govuk-aws/blob/master/terraform/userdata/10-attach-volumes). We ensure it is loaded in from adding it to the runtime [here](https://github.com/alphagov/govuk-aws-data/blob/master/data/app-logs-cdn/integration/common.tfvars#L3).

terraform/projects/app-mongo/main.tf
issyl0: We have prior art for this in [app-graphite](https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/app-graphite/main.tf#L209).

jljalderman:
redoing

PR #503, opened 2018-03-01
Fix email alert api public service

boffbowsh:
Firstly, use correct name for email alert api external elb security group output. It should use dashes not underscores for `email-alert-api`.

Also, we hit the character limit on an ELB target group name. This changes the naming to be managed by TF, and moves the previously specified naming into tags.


jljalderman:
looks ok to me

PR #504, opened 2018-03-01
Add completion for govukcli

surminus:
Someone mentioned about bash completion (/cc @boffbowsh), so as a firebreak task I thought I'd add it to govukcli as it makes it easier to get to instances.

The `$lastword` variable is useful as it stops recursive completion against commands, which is why it's use in all the commands across these gross case statements. Writing in case statements does make it easier to add in further options in the future, though.

### DIFF
--- a/doc/guides/connecting-to-the-environment.md
+++ b/doc/guides/connecting-to-the-environment.md
@@ -48,3 +48,25 @@ To view current context:
 To list available contexts:
 
 `govukcli list-contexts`
+
+### Completion
+
+The script comes with tab completion to make it easier to work with.
+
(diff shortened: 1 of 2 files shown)


boffbowsh:
👍 but it needs docs on how to install it. I added it to `/usr/local/etc/bash_completion.d/` but that might be wrong

surminus:
Ah yes, I'll add some docs tomorrow. 

It can either go in a bash_completion directory or equally could be sourced in a `.bashrc`/`.zshrc` file. I'll test using the preferred homebrew method as that's what most folks use.

surminus:
Added docs.

PR #507, opened 2018-03-05
Add an AWS credentials helper to govukcli

boffbowsh:
Handles fetching AWS sessions for both gds-users and the requested
GOV.UK environment, including requesting the MFA token.

We fetch and cache a session token for gds-users first so we can assume
roles in other accounts without having to re-request an MFA token.

Requires a small documentation change in the dev docs: https://github.com/alphagov/govuk-developer-docs/pull/746

```
$ govukcli aws s3 ls
Enter MFA Token:
```

### DIFF
--- a/tools/govukcli
+++ b/tools/govukcli
@@ -11,7 +11,6 @@ CONFIG_DIR=~/.govukcli
 CONTEXT_FILE=$CONFIG_DIR/current-context
 CONTEXTS='ci integration staging production staging-aws production-aws'
 
-
 ### Help output
 function usage {
   cat << EOF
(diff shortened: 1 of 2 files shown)


surminus:
I found an error:

```➜  ~/govuk/govuk-aws git:(add-aws-helper) ✗ govukcli aws
/usr/local/bin/govukcli: line 249: [: ==: unary operator expected
Enter MFA token: ^C```

boffbowsh:
Ta, fixed

surminus:
I successfully tested this 👍 

I think it would be good to update `govukcli.completion` along with this change to keep it up to date.

My other comment would be I think we should encourage people _not_ to store their credentials on disk - I personally really dislike this, although the MFA token goes some way in ensuring this isn't abused. I have an alternative method which maybe I'll expand on when this is merged.

boffbowsh:
I've updated the bash completion for this

PR #509, opened 2018-03-05
Add shellcheck

surminus:
Check the scripts in the tools directory, including govukcli, and userdata scripts.

I'm a bit wary of merging this as it changes things that could be problematic, but I also have faith in `shellcheck` and the documentation that comes with it. It caught a few things that I was surprised at and on the whole I think it makes it a better set of scripts.

### DIFF
--- a/tools/aws-copy-puppet-setup.sh
+++ b/tools/aws-copy-puppet-setup.sh
@@ -6,9 +6,9 @@ set -ex
 while getopts "e:s:h" option
 do
   case $option in
-    h ) HELP=1 ;;
     s ) STACKNAME=$OPTARG ;;
     e ) ENVIRONMENT=$OPTARG ;;
+    h|* ) HELP=1 ;;
   esac
 done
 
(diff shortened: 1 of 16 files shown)


tools/aws-copy-puppet-setup.sh
```diff
@@ -50,7 +50,7 @@ mkdir -p /usr/share/puppet/production/releases
 mv ~/govuk-puppet /usr/share/puppet/production/releases/${RELEASENAME}
 rm -f /usr/share/puppet/production/current
 ln -s /usr/share/puppet/production/releases/${RELEASENAME} /usr/share/puppet/production/current
-mkdir -p -m 0700 /etc/puppet/gpg
+mkdir -p 0700 /etc/puppet/gpg
```
surminus: As noted this was an incorrect change.

surminus:
@deanwilson and I paired reviewing on this. It may be worth a test on userdata - this could be achieved by deploying the branch to a project, and terminating an instance and ensuring it comes up correctly.

deanwilson:
Merge away.

PR #511, opened 2018-03-06
Update DNS architecture ADR

surminus:
It needs to be amended slightly to make clear that we cannot do application specific green and blue deployments.

Also adds a clearer diagram.


issyl0:
Thanks for re-doing this! Nitpick: it would be nice to have this diagram in higher quality.

surminus:
Updated with a higher resolution image.

PR #531, opened 2018-03-22
Revert "add_mongo_ebs_volume"

schmie:
Reverts alphagov/govuk-aws#520


schmie:
Will rather re-apply terraform from prior version

PR #533, opened 2018-04-04
Increase the assume-role MFA expiry time to 8 hours (28800 seconds)

issyl0:
- AWS recently increased the assume-role session token expiry time to a
  maximum of 12 hours. This makes our sessions last for eight hours, the
  length of the working day, to avoid having to re-assume role multiple
  times a day.
- This will revert the changes I made in
  https://github.com/alphagov/govuk-aws/pull/498, and I'll have to
  change the docs again, but it makes it simpler in the long run - you
  only have to remember an `assume-role` command, not also
  `get-session-token`.

(I'll test this in the test account first.)

### DIFF
--- a/terraform/modules/aws/iam/role_user/main.tf
+++ b/terraform/modules/aws/iam/role_user/main.tf
@@ -62,15 +62,6 @@ data "aws_iam_policy_document" "assume_policy_document" {
         "true",
       ]
     }
-
-    condition {
-      test     = "NumericLessThan"
-      variable = "aws:MultiFactorAuthAge"
-
-      values = [
-        "43200",
-      ]
-    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


issyl0:
It seems like I have to wait for https://github.com/terraform-providers/terraform-provider-aws/pull/3977, upgrade the provider and then change my approach. Closing this for now.

issyl0:
That PR has been merged now, so I've implemented this in the correct way (hopefully). I'll test it in the test account and see if it works.

issyl0:
That appears to have been the correct setting. I did an `apply` in the test account and both govuk-users and govuk-administrators roles are coming up as:

screenshot 2018-04-11 13 49 23

issyl0:
Applied in all environments. Merging.

PR #538, opened 2018-04-12
Update the govukcli script to only use `aws assume-role`

issyl0:
- Now that tokens last for 8 hours
  (23dd074dbfaf4b7201d3fd11d9c948e6806cfbf2), we don't need to
  re-assume-role every hour.
- Now we don't need to do that, we can just use `aws assume-role` once
  every eight hours, we don't need the intermediary step of `aws sts
  get-session-token`.

(This requires #533 to be merged and deployed first.)
Reviewers will find the `?w=1` query string useful to ignore whitespace changes.

### DIFF
--- a/tools/govukcli
+++ b/tools/govukcli
@@ -297,35 +297,24 @@ EOF
   if ! _check_credentials; then
     unset AWS_SESSION_TOKEN
 
-    # check if GDS session is valid
-    GDS_SESSION=~/.aws/gds/session
-    test -f ${GDS_SESSION} && source ${GDS_SESSION}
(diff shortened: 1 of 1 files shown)


boffbowsh:
The reason for the get-session-token was so that you didn't need to enter your MFA token individually for integration, staging and production. I dunno how annoying that will be.

issyl0:
Ah! I thought it was purely to get around entering MFA details once an hour on `assume-role`.

We could not merge this this, or we could trial it for a bit - not much switching around environments will happen until we've got staging and production in AWS, which won't be for a few months still.

PR #541, opened 2018-04-12
Specify PosgreSQL major version only

rubenarakelyan:
This commit changes the PosgreSQL version to be the major version only since RDS minor version updates are automatically applied.


issyl0:
Should we do this for other Postgresql RDS instances (transition)? Does/Will MySQL have the same issue?

rubenarakelyan:
Yup, done that and it works with no changes planned

PR #542, opened 2018-04-13
Keep the mirrorer instance from hanging after a while

schmie:
- Ramp up instance size from t2.micro to t2.large, matching the 8GB of RAM on
  the Carrenza instance. (The mirrorer/crawler runs its own Redis and
  RabbitMQ)

Solo: @schmie


issyl0:
This is superseded by https://github.com/alphagov/govuk-aws/pull/550.

PR #543, opened 2018-04-16
Make our monitoring instances m5.large instead of t2

boffbowsh:
The monitoring instance that runs Icinga and Smokey in Integration has
begun running out of CPU credits. This is making it difficult to
diagnose problems elsewhere in the stack. Making it an m5.large costs
~$77pcm instead of ~$36pcm, and doubles the RAM.

screen shot 2018-04-16 at 09 49 19


boffbowsh:
Seems only in the last few days, but I think it's an important enough instance to not want to sorry about it for the sake of $35pcm.

PR #552, opened 2018-04-24
AWS RDS Timeout

suthagarht:
- We have identified that we can change the AWS token TTL. This will
  help terraform apply without timing out soon. However, terrafom sets some
default time outs by default.
 https://www.terraform.io/docs/providers/aws/r/db_instance.html [Please
check Timeouts]

- If this is not addressed, we get a message similar to the following.

"... aws_db_instance.db_instance_replica: timeout while waiting for
state to become 'available, storage-optimization' ..."

- This change is aimed at extending the default time-out to match the
  maximum AWS TTL that we can set (8 hours).

https://trello.com/c/KlI3mK8H/1134-postgresql-rds-integration

https://govuk.zendesk.com/agent/tickets/2629112

Solo: @suthagarht

### DIFF
--- a/terraform/modules/aws/rds_instance/main.tf
+++ b/terraform/modules/aws/rds_instance/main.tf
@@ -151,6 +151,24 @@ variable "snapshot_identifier" {
   default     = ""
 }
 
+variable "terraform_create_rds_timeout" {
+  type        = "string"
+  description = "Set the timeout time for AWS RDS creation."
+  default     = "2h"
+}
+
+variable "terraform_update_rds_timeout" {
+  type        = "string"
+  description = "Set the timeout time for AWS RDS modification."
+  default     = "2h"
+}
+
+variable "terraform_delete_rds_timeout" {
+  type        = "string"
+  description = "Set the timeout time for AWS RDS deletion."
+  default     = "2h"
+}
+
 # Resources
 # --------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


issyl0:
If we have an 8 hour Terraform RDS creation job, something has gone very wrong! I think these values should be distinct from our individual token expiry times. Maybe we could increase them to only 1h30, or something? How long is the creation of the snapshot likely to take? How often are we going to have to do these?

To make these numbers more easily editable everywhere, could we make them variables from govuk-aws-data?

suthagarht:
Hi @issyl0 ,

Thank you for checking.

Can we pass this as variables [govuk-aws] - certainly

Postgresql took between 105-80 minutes. This was just for the replica. I guess we can change this to 120 minutes. I don't know whether that would be limiting for mysql. We can try and change if necessary.

I will make the changes now.

Thank You
SK

suthagarht:
Hi @issyl0 , Variables are now provided via https://github.com/alphagov/govuk-aws-data/pull/128 . Also, they are called via the module.

issyl0:
Thanks! Could you edit the commit message (`git commit --amend` lets you edit just the message, not the files changed) to specify the new, lower, timeout, please? It still says eight hours.

suthagarht:
Hi @issyl0 , sorry. Done.

PR #553, opened 2018-04-25
Make out-of-date docs fail CI

issyl0:
- We often forget to run this script when we update files, so instead
  I'm adding it as a CI step that fails when
  changes are detected by the `terraform-docs` command.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,4 +15,3 @@
       entry: tools/terraform-validate.sh
       language: script
       files: \.tf$
-
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,6 +24,17 @@ node ("terraform") {
       ])
     }
 
+    stage("Check generated docs are up-to-date") {
+      sh "tools/update-docs.sh"
+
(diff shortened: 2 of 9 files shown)


issyl0:
Oh good, it works already:

```
+ find . -name *.tf
+ xargs tools/check-for-docs.sh
Must install terraform-docs
```

😂

tijmenb:
Oh yes! I ran into that too in https://github.com/alphagov/govuk-aws/compare/docs-fail-ci?expand=1 and gave up :(

issyl0:
The `terraform-docs` package is now installed (💦) but I'm seeing some weird stuff with the output of the script - it wants to *remove* things from READMEs. I'm not sure that's right. It was also looking at all `.tf` files, not just `main.tf` like the old script, so I edited it to only `main.tf` files, but I'm still not convinced it's working. Tempted just to use @tijmenb's branch instead!

issyl0:
I stole @tijmenb's code for this, but a rebase and removed all notion that I'd ever `cherry-pick`ed the commit - sorry Tijmen!

This is ready to go.

PR #555, opened 2018-04-26
Add data warehouse RDS instance

boffbowsh:
Creates a new RDS postgres instance to hold the Data Warehouse.

Secrets added in https://github.com/alphagov/govuk-aws-data/pull/131


boffbowsh:
Good catch on the instance type discrepancy, fixed thanks!
64GB is more than enough - the current usage is ~8GB but will grow
Should be the same AMI as for the other db_admin types

PR #560, opened 2018-05-04
Puppet Master Instance Type

suthagarht:
- We have noticed high CPU usage in the AWS staging puppet master. While
comparing with the Carrenza staging server, we noticed that we have less
CPU cores in AWS. Hence we have performed the following change.

Please check here for history:
https://trello.com/c/nr8kuCOv/1170-aws-staging-puppet-is-broken

Solo: @suthagarht

### DIFF
--- a/terraform/projects/app-puppetmaster/main.tf
+++ b/terraform/projects/app-puppetmaster/main.tf
@@ -19,6 +19,12 @@ variable "aws_environment" {
   description = "AWS environment"
 }
 
+variable "puppetmaster_instance_type" {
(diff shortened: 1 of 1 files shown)


issyl0:
Do we need 2xl in integration too? If not, it might be doing something like https://github.com/alphagov/govuk-aws/pull/550.

suthagarht:
Thank you @issyl0 , You are correct. Done. https://github.com/alphagov/govuk-aws-data/tree/puppet-master-instance-type

PR #562, opened 2018-05-16
Puppetmaster auto bootstrap

schmie:
- The old bootstrap process involves a lot of manual execution and is
  quite complicated

- Integrating formerly manual executed scripts into one and call it via
  terraform userdata snippet 20-puppetmaster

- Move required private keys to AWS SSM parameter store

- Update docs to reflect changes

Pair: @rtrinque @schmie


schmie:
Depends on facter upgrade to work: https://github.com/alphagov/govuk-puppet/pull/7561

PR #563, opened 2018-05-21
Adding support for a secondary certificate in the aws/lb module

rtrinque:
This is required because we want to be able to use the .digital and .publishing.service.gov.uk certificates on the public service load balancers to make the switch as seamless as possible


issyl0:
Could you please edit your commit message as per the [styleguide](https://github.com/alphagov/styleguides/blob/master/git.md#structure)?

rtrinque:
Commit message amended

PR #565, opened 2018-05-22
Fix SSH key parameter name

schmie:
- This was accidentally ommitted from the original pull request

- SSH key name for cloning from github.com

@schmie


suthagarht:
This relates to https://github.com/alphagov/govuk-aws/pull/562

PR #566, opened 2018-05-23
Increase mongo disk size to cope with dumps

benthorner:
https://alert.integration.publishing.service.gov.uk/cgi-bin/icinga/extinfo.cgi?type=2&host=ip-10-1-4-78.eu-west-1.compute.internal&service=low+available+disk+space

The disk on mongo integration is nearly full, which is preventing the
sync from production from working. Doubling the disk size should tie us
over for the foreseable future.

### DIFF
--- a/terraform/projects/app-mongo/main.tf
+++ b/terraform/projects/app-mongo/main.tf
@@ -131,7 +131,7 @@ module "mongo-1" {
 resource "aws_ebs_volume" "mongo-1" {
   availability_zone = "${lookup(data.terraform_remote_state.infra_networking.private_subnet_names_azs_map, var.mongo_1_subnet)}"
   type              = "io1"
-  size              = 150
+  size              = 300
   iops              = 100
 
   tags {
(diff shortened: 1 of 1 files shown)


benthorner:
@alphagov/reliability-engineering Could someone take a look at this? It's breaking the integration sync.

terraform/projects/app-mongo/main.tf
```diff
@@ -131,7 +131,7 @@ module "mongo-1" {
 resource "aws_ebs_volume" "mongo-1" {
   availability_zone = "${lookup(data.terraform_remote_state.infra_networking.private_subnet_names_azs_map, var.mongo_1_subnet)}"
   type              = "io1"
-  size              = 150
+  size              = 300
```
issyl0: This just increases the size for the volume attached to `mongo-1`, but there are three instances (further down in this file, lines 189 and 231). Is the dump pinned to just one of the instances, or could it be any of them (given instances in AWS are dynamic)?

benthorner: Good spot - I was too happy after finding the place to change it ;-)

PR #569, opened 2018-05-25
Manage Fastly service for data.gov.uk in Terraform

boffbowsh:
This is the basic definition for data.gov.uk's Fastly service. The current resource has been imported, and the definition written so that `plan` results in no changes. Soon, I'll be changing this so it has different backends for different paths. There's only a `production` backend as there's only a Production service.

This is a proof-of-concept. I'd like a little discussion on whether this is the right place for this. This is a well-structured terraform repo and framework, so it's easy to put it here. However, this isn't actually AWS so it's a bit weird.

### DIFF
--- a/terraform/projects/fastly-datagovuk/README.md
+++ b/terraform/projects/fastly-datagovuk/README.md
@@ -1,0 +1,13 @@
+## Project: fastly-datagovuk
+
+Manages the Fastly service for data.gov.uk
+
+
+## Inputs
+
+| Name | Description | Type | Default | Required |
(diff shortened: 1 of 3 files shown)


boffbowsh:
@issyl0 @stephenharker do you have any opinions on this?

boffbowsh:
Yes, this is a proof of concept. If this works well for DGU's service, we can do it for GOV.UK Fastly services. This is already tracked as an item of tech debt: https://trello.com/c/y6MIgxjp/14-cdn-configuration-is-spread-over-multiple-repositories-and-not-reproducible.

PR #571, opened 2018-06-05
AWS RDS Instance naming.

suthagarht:
- We don't specify an instance name when we create AWS RDS instances.
Hence, a default name is issued at creation. For example, it would look
like the following. 'terraform-20180410102142467900000002'

- This poses two issues.
  - It is difficult to view the AWS RDS console and determine the
instance visually, without selecting the instance and checking other
notations.
  - We identified that it is difficult to create Icninga checks with
dynamic names.

- To address the above mentioned issues, we are implementing the
`identifier` attribute of the `aws_db_instance` Terraform module.

https://trello.com/c/C7SY8edZ/1123-monitor-rds-and-elasticache-in-aws-and-alert-icinga

Solo: @suthagarht


issyl0:
This is failing on documentation not being up to date.

PR #573, opened 2018-06-11
Icinga Monitoring - AWS RDS

suthagarht:
- We have made few changes to enable AWS RDS instances to be monitored by
Icinga. This change incorporates some instructions/guidance on how to
add an instance to be monitored or to add a new matric to be monitored.

https://trello.com/c/C7SY8edZ/1123-monitor-rds-and-elasticache-in-aws-and-alert-icinga

Pair: @suthagarht @szd55gds


suthagarht:
> The check is arbitrary and uses `terraform-docs` to generate docs from the manifests. I'm not sure how many people ever read those READMEs, so it's probably not a great place for docs. If you want to write documentation, maybe put it in the `guides/` or `docs` directory?

From issy

PR #574, opened 2018-06-11
Allow Production Carrenza to talk to Integration

issyl0:
- We have a Jenkins job that runs after the data sync jobs, called
  "trigger data sync complete". This has been failing while trying to
  trigger the integration one, because access to
  `https://deploy.integration.publishing.service.gov.uk` has been
  failing. The IPs in each of the `carrenza_{environment}_ips` variables
  in govuk-aws-data look good, they're just not referenced anywhere in
  the security groups for the deploy ELB.


issyl0:
Do we want Staging IPs in this list too?

issyl0:
I removed staging from the allowed list, as the problem is currently only between Production and Integration.

issyl0:
Both a `plan` and `apply` were successful. Merging.

PR #576, opened 2018-06-12
Add ckan machine class

boffbowsh:
Adds a new class of machine for running CKAN. This requires a persistent EBS volume due to the need to run Solr on this machine. The RAM requirement is high due to the number of harvester processes running.

Requires https://github.com/alphagov/govuk-aws-data/pull/150


terraform/projects/app-ckan/main.tf
```diff
@@ -0,0 +1,287 @@
...
+  surgequeuelength_threshold     = "0"
+  healthyhostcount_threshold     = "0"
+}
+
+# Outputs
+# --------------------------------------------------------------
+
+output "ckan_elb_internal_address" {
+  value       = "${aws_elb.ckan_elb_internal.dns_name}"
+  description = "AWS' internal DNS name for the ckan ELB"
```
andrewgarner: Shall we say "CKAN" as it is an acronym "Comprehensive Knowledge Archive Network"?!

terraform/projects/app-ckan/remote_state.tf
```diff
@@ -0,0 +1,111 @@
...
+
+variable "remote_state_infra_networking_key_stack" {
+  type        = "string"
+  description = "Override infra_networking remote state path"
+  default     = ""
+}
+
+variable "remote_state_infra_security_groups_key_stack" {
+  type        = "string"
+  description = "Override infra_security_groups stackname path to infra_vpc remote state "
```
andrewgarner: Do these descriptions need trailing whitespace?

terraform/projects/app-ckan/main.tf
boffbowsh: It's lowercase as it's the name of the ELB, which is lowercase.

andrewgarner: This is a description though?

boffbowsh: Yes but it's explicitly a description of which ELB it's pointing to, not what the ELB is used for.

andrewgarner: Fair enough 👍

terraform/projects/app-ckan/remote_state.tf
boffbowsh: Fixed

PR #580, opened 2018-06-27
Allow ingress of graphite to self for grafana

schmie:
- In AWS grafana living on the same instance as graphite can no longer
  access graphite as a data source

Pair: @schmie @benthorner

### DIFF
--- a/terraform/projects/infra-security-groups/graphite.tf
+++ b/terraform/projects/infra-security-groups/graphite.tf
@@ -75,6 +75,19 @@ resource "aws_security_group_rule" "graphite_ingress_graphite-internal-elb_pickl
   source_security_group_id = "${aws_security_group.graphite_internal_elb.id}"
 }
 
+resource "aws_security_group_rule" "graphite_ingress_graphite_internal_elb_https" {
+  type      = "ingress"
+  from_port = 443
+  to_port   = 443
+  protocol  = "tcp"
+
+  # Which security group is the rule assigned to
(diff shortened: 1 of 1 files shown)


cbaines:
> In AWS grafana living on the same instance as graphite can no longer access graphite as a data source

Any idea why? I was pretty certain it was working up until a few days ago, but unfortunately I don't know what configuration was used.

I had a check earlier, and it seems that using the hostname `graphite` apparently works (which I didn't change). I guess this change might not be necessary?

benthorner:
So it looks like someone or something had changed the datasources: ES had been deleted (we had to go to Logit to create it from scratch) and Graphite was set to 'graphite.cluster', for which there's no evidence AWS integration ever supported such a domain. We found that 'graphite' isn't a simple loopback - it goes through the internal LB - so a new security rule was necessary to get it working.

cbaines:
> So it looks like someone or something had changed the datasources: ES had been deleted (we had to go to Logit to create it from scratch) and Graphite was set to 'graphite.cluster', for which there's no evidence AWS integration ever supported such a domain. We found that 'graphite' isn't a simple loopback - it goes through the internal LB - so a new security rule was necessary to get it working.

So I believe the datasources were deleted, because they are stored on the machine, and the machine was deleted.

I manually created the Graphite datasource, as I wanted to use it, but I couldn't work out the correct value, `graphite.cluster` was one of the ones I tried unsuccessfully.

benthorner:
Ah, the mist is lifting now. It took us quite a while to find something that worked!

PR #581, opened 2018-06-27
Bucket for CDN Logs

kevindew:
Off the back of https://github.com/alphagov/govuk-rfcs/pull/93 this is a PR a new bucket that can be used for storing data with fastly.

Although we already have a user and credentials for some specialised fastly logs I felt it was more fitting with the lay out of this project to create a dedicated user for that role.


terraform/projects/fastly-logs/main.tf
```diff
@@ -0,0 +1,80 @@
...
+  required_version = "= 0.11.7"
+}
+
+provider "aws" {
+  region  = "${var.aws_region}"
+  version = "1.14.0"
+}
+
+resource "aws_s3_bucket" "fastly_logs" {
+  bucket = "govuk-fastly-logs-${var.aws_environment}"
```
schmie: Lately we have used govuk-${var.aws_environment}-bucket-name as naming convention

terraform/projects/fastly-logs/main.tf
```diff
@@ -0,0 +1,80 @@
...
+
+  tags {
+    Name            = "govuk-fastly-logs-${var.aws_environment}"
+    aws_environment = "${var.aws_environment}"
```
schmie: +1

terraform/projects/fastly-logs/main.tf
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  logging {
+    target_bucket = "${data.terraform_remote_state.infra_monitoring.aws_logging_bucket_id}"
+    target_prefix = "s3/govuk-fastly-logs-${var.aws_environment}/"
+  }
```
schmie: +1

PR #582, opened 2018-06-27
Create bucket for govuk-cdn-logs-monitor data

kevindew:
As per https://github.com/alphagov/govuk-rfcs/pull/93 we're planning to
archive the govuk-cdn-logs-monitor application and this serves as a
place to store the data until we decide it is fine to delete it.


kevindew:
@schmie Thanks 👍

PR #584, opened 2018-06-27
Email Alert API Archive Project

kevindew:
Trello: https://trello.com/c/FM3MCuAl/255-investigate-email-alert-api-database-size-and-push-email-archives-to-s3

This creates an S3 bucket which can be accessed to store archive data
from Email Alert API. Currently we have a large PostgreSQL database
storing this information which is not efficient.

The intention is to store this in S3 and then access it via Athena.


kevindew:
Thank you 🙂

PR #588, opened 2018-07-05
AWS Glue Crawlers for log files

kevindew:
This sets up crawlers to run with the 3 strands of log files we have being sent from Fastly to S3: govuk-www, govuk-assets and bouncer.

It sets up a database for these 3 log files and a schema based on the format Fastly is configured to send (details of this is in the individual commits)

Finally this deletes logs from these buckets when they reach 90 days old.

### DIFF
--- a/terraform/projects/infra-fastly-logs/main.tf
+++ b/terraform/projects/infra-fastly-logs/main.tf
@@ -487,6 +487,174 @@ resource "aws_glue_catalog_table" "govuk_assets" {
   }
 }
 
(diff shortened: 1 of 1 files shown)


boffbowsh:
Should the Lambda function used to partition the files be included in this?

kevindew:
@boffbowsh No lambda required 🎉can specify it in fastly UI: `/govuk_www/year=%Y/month=%m/date=%d/`

boffbowsh:
Mind. Blown.

PR #589, opened 2018-07-09
Change field type for Email Alert API DB definition

kevindew:
Trello: https://trello.com/c/FM3MCuAl/255-investigate-email-alert-api-database-size-and-push-email-archives-to-s3

The field, subscriber_id, was incorrectly set up as a string with the expectation that this field would be a UUID. In fact it is actually an integer in Email Alert API so this updates this to match.


boffbowsh:
@kevindew I'll merge and apply this

kevindew:
@boffbowsh Thanks 👍

PR #590, opened 2018-07-11
(WIP) Puppet Master build fixes

dougneal:
Related to https://github.com/alphagov/govuk-puppet/pull/7731.

Updates/fixes to the bootstrap process which takes the stock Ubuntu AMI and runs `puppet apply` to install the Puppet Master.

### DIFF
--- a/terraform/userdata/20-puppetmaster
+++ b/terraform/userdata/20-puppetmaster
@@ -30,7 +30,7 @@ cd /var/govuk
 
 # Get git and checkout govuk-aws
 apt-get -y install git
-git clone https://github.com/alphagov/govuk-aws.git
+git clone --branch puppetmaster-build-fixes https://github.com/alphagov/govuk-aws.git
 
 # Run the bootstrap script
 /var/govuk/govuk-aws/tools/govuk-puppetmaster-bootstrap.sh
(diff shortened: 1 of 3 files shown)


terraform/userdata/20-puppetmaster
```diff
@@ -30,7 +30,7 @@ cd /var/govuk
 
 # Get git and checkout govuk-aws
 apt-get -y install git
-git clone https://github.com/alphagov/govuk-aws.git
+git clone --branch puppetmaster-build-fixes https://github.com/alphagov/govuk-aws.git
```
dougneal: Revert this fragment once https://github.com/alphagov/govuk-puppet/pull/7731 is merged

tools/govuk-puppetmaster-bootstrap.sh
```diff
@@ -1,128 +1,88 @@
...
+get_ssm_parameter 'govuk_base64_github.com_ssh_readonly' | base64 --decode > /root/.ssh/id_rsa
+chmod 400 /root/.ssh/id_rsa
+
+# Import GPG for decrypting secrets in Hiera-EYAML
+mkdir /etc/puppet/gpg
+chmod 700 /etc/puppet/gpg
+(
+  get_ssm_parameter 'govuk_base64_staging_gpg_1_of_3'
+  get_ssm_parameter 'govuk_base64_staging_gpg_2_of_3'
+  get_ssm_parameter 'govuk_base64_staging_gpg_3_of_3'
```
dougneal: We agreed to rename the parameters to remove the string `staging`, this means that the script will just use the same name in each environment but the key material will differ (given that they run in different AWS accounts).

PR #594, opened 2018-07-17
Allow pentesters to touch everything

deanwilson:
These resources are deliberatly being added out of sequence so they are
more visually disturbing


issyl0:
Hopefully we can just click “revert” on this PR when they’re done.

issyl0:
Is this everything now (management, databases (RDS, mongo, Redis))? Or is the first commit message misleading and we need to add rules to every security group file?

issyl0:
Ran a `plan` and `apply` on this, successfully.

PR #598, opened 2018-07-19
Install the AWS CLI from the Apt repo instead of from Pip

dougneal:
Add GOV.UK Apt repos in 00-base cloud-init snippet.

Install the signing key and configure (one of) the GOV.UK Apt repos (awscli).

Install awscli in 00-base with apt, and don't install it in `tools/govuk-puppetmaster-bootstrap.sh` with pip.

### DIFF
--- a/terraform/userdata/00-base
+++ b/terraform/userdata/00-base
@@ -26,6 +26,13 @@ wget -qO $TMPFILE http://apt.puppetlabs.com/puppetlabs-release-$DISTRIB_CODENAME
 dpkg -i $TMPFILE
 rm $TMPFILE
 
+echo "[$(date '+%H:%M:%S %d-%m-%Y')] base: install Gov.UK package signing key and repositories ..."
+curl --silent --show-error --fail \
+  https://apt.publishing.service.gov.uk/EC5FE1A937E3ACBB.asc | apt-key add -
+
+echo "deb [arch=amd64] http://apt.production.alphagov.co.uk/awscli $DISTRIB_CODENAME main" \
+  >> /etc/apt/sources.list.d/awscli.list
+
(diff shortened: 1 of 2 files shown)


deanwilson:
What is the reason behind the change, and the benefits from moving to the package? The PR is about the what, rather than the why.

When we rebuild environments does this create a cyclic dependency where the apt box won't have been built yet?

dougneal:
We built a Debian package of the AWS CLI as a dependency for the data sync task, as the Trusty servers in Carrenza required a much more recent AWS CLI than the distro provides. Personally I wasn't happy with installing it with Pip - my opinion is that mixing package managers is generally bad practice unless there are specific mitigations around conflicts, with items installed being namespaced and encapsulated properly. Specifically in this case, one of awscli's dependencies is PyYAML, which has native code components - it needs to be compiled against libyaml, which requires libyaml-dev to be installed, as well as the C compiler toolchain, and compiling things at install time (as root!) on production servers is also bad practice, probably worse than mixing package managers.

The Debian package for AWS that we built uses Amazon's quite nice installer to package away the AWS CLI and all of its dependencies (including the Python binaries themselves) into one self-contained tree under /opt, with a shim at `/usr/bin/aws`, and provides a clean upgrade path from the distribution's native awscli package without any risk of conflicts (e.g., we were seeing two versions of awscli installed, one old version in /usr/bin and one new version in /usr/local/bin).

As for cyclic dependencies for environment builds, this is something to keep an eye on generally and the problem would not be limited to the awscli package. The apt repo server is hosted in the production environment and all other environments point to it, but we should consider the apt repo server to really be separate/external to all environments to ensure we don't make cyclic dependency problems for ourselves.

PR #599, opened 2018-07-19
Due to no guidance as to what cmd to run to deploy Puppet Master

szd55gds:


doc/guides/environment-provisioning.md
```diff
@@ -141,6 +141,9 @@ The script requires for secrets to be available to the blue-puppetmaster role in
 - `govuk_staging_secrets_1_of_2_gpg`, First part of the GPG key. See below.
 - `govuk_staging_secrets_2_of_2_gpg`, Second part of the GPG key. At the moment the length of SecretString in the AWS SSM parameter store is limited to 4096 characters. 
 
+Sample one-line command to deploy Puppet Master (from govuk-aws)
+- `tools/build-terraform-project.sh -c plan -p app-puppetmaster -d ../govuk-aws-data/data -e production -s blue`
```
issyl0: Further up in this guide, we set the `ENVIRONMENT`, `DATA_DIR` and `STACKNAME` environment variables to avoid re-specifying everything every time - and to avoid copy/pasting commands running against the wrong environment (eg, you've specified `production` here).

Could we use the same pattern, eg just have `tools/build-terraform-project.sh -p app-puppetmaster -c plan` as in other occurrences, instead of all of the options?

issyl0:
Thanks. Could you also please edit your commit message as per our [styleguide](https://github.com/alphagov/styleguides/blob/master/git.md)? Present tense, eg "Add example command for deploying the Puppetmaster."

PR #605, opened 2018-07-31
Revert "AWS S3 Database Backup Writer"

suthagarht:
Reverts alphagov/govuk-aws#585

- We still need **Integration** to be able to write to the backup bucket.
- The new data-sync script doesn't use **prefix**. This means that S3 IAM policy rules that were applied using this commit will prevent the backup. https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_env_sync/files/govuk_env_sync.sh
- At the current moment, backups from **Integration** to the S3 backup buckets are failing.


schmie:
Will this break the access scheme for the new sync? Could we just add the old resources into the new scheme instead of just reverting everything? : )

issyl0:
I think SK said he'd fix this tomorrow. He's now off for the afternoon. If we know this is broken, and have known it's broken for a while, we're probably better off just rolling forward tomorrow rather than breaking more in-progress things today.

PR #608, opened 2018-08-09
New VPC resources for VPN to Carrenza.

stephenharker:
This creates the two gateways and a VPN for connecting the VPC to Carrenza. There are 2 new variables to add in govuk-aws-data also. This should create the VPN and then we can get the created configuration to use to update the VCloud configuration in govuk-provisioning to deploy the Carrenza side of things.

### DIFF
--- a/terraform/projects/infra-vpc/main.tf
+++ b/terraform/projects/infra-vpc/main.tf
@@ -49,6 +49,16 @@ variable "remote_state_infra_monitoring_key_stack" {
   default     = ""
 }
 
+variable "carrenza_vpn_endpoint_ip" {
+  type        = "string"
+  description = "Public IP address of the VPN gateway in Carrenza"
+}
+
+variable "carrenza_internal_net_cidr" {
+  type        = "string"
+  description = "Internal network range of the environment in Carrenza"
+}
+
 # Resources
 # --------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


terraform/projects/infra-vpc/main.tf
```diff
@@ -123,6 +133,40 @@ module "vpc_flow_log_exporter" {
   lambda_log_retention_in_days = "${var.cloudwatch_log_retention}"
 }
 
+resource "aws_customer_gateway" "carrenza_vpn_gateway" {
+  bgp_asn    = 65000
+  ip_address = "${var.carrenza_vpn_endpoint_ip}"
```
issyl0: I guess this is one of the things that needs a `govuk-aws-data` PR?

stephenharker: Yes I have made one of these now:

https://github.com/alphagov/govuk-aws-data/pull/189

But I wonder if I should add these for Integration and Production before we merge this PR..?

issyl0: Yes please, otherwise we'll be blocked on deploying these changes to those environments because `var.carrenza_vpn_endpoint_ip` won't exist. Do we need a VPN in integration though?

stephenharker: Ok I'll do that. We can create the config for Integration even if we don't end up connecting it to anything.

PR #609, opened 2018-08-09
Add RDS policy to db-admin instance

rtrinque:
We give the right to db-admin instance to create an RDS to do
database scrubbing before importing them in integration.

### DIFF
--- a/terraform/projects/app-db-admin/main.tf
+++ b/terraform/projects/app-db-admin/main.tf
@@ -157,6 +157,17 @@ resource "aws_iam_role_policy_attachment" "read_db-admin_database_backups_iam_ro
   policy_arn = "${data.terraform_remote_state.infra_database_backups_bucket.dbadmin_read_database_backups_bucket_policy_arn}"
 }
 
+resource "aws_iam_policy" "db-admin_iam_policy" {
+  name   = "${var.stackname}-db-admin-additional"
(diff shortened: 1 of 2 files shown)


terraform/projects/app-db-admin/main.tf
```diff
@@ -157,6 +157,17 @@ resource "aws_iam_role_policy_attachment" "read_db-admin_database_backups_iam_ro
   policy_arn = "${data.terraform_remote_state.infra_database_backups_bucket.dbadmin_read_database_backups_bucket_policy_arn}"
 }
 
+resource "aws_iam_policy" "db-admin_iam_policy" {
+  name   = "${var.stackname}-db-admin-additional"
+  path   = "/"
+  policy = "${file("${path.module}/additional_policy.json")}"
+}
+
+resource "aws_iam_role_policy_attachment" "db-admin_iam_role_policy_attachment" {
```
issyl0: Is it worth naming this something like `datascrubber_policy_attachment` rather than a generic `iam_role_policy_attachment`?

rtrinque: I guess, I just followed the general trend, we call most (all?) our policies like that

issyl0: OK then.

PR #610, opened 2018-08-10
Pluralize the `govuk_poweruser` role name to match the other roles

issyl0:
- This inconsistency is annoying for everyone new to assume-role, and
  our docs are even inconsistent. Instead of changing the docs, change the
  pluralization of the role name to match `govuk-administratorS` and
  `govuk-userS`, as now it's just a snowflake and always takes some time
  to debug problems with people's config because of what is essentially
  a typo.

### DIFF
--- a/terraform/projects/infra-security/main.tf
+++ b/terraform/projects/infra-security/main.tf
@@ -89,7 +89,7 @@ module "role_admin" {
 
 module "role_poweruser" {
   source           = "../../modules/aws/iam/role_user"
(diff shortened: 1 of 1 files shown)


issyl0:
This will have the side effect of needing everyone who currently assumes the PowerUser role having to change the role name (to add an `s` - hurray).

issyl0:
This has gone through successful `plan` and `apply` steps.

PR #617, opened 2018-08-20
Add new project for Content Publisher image bucket

benthorner:
https://trello.com/c/63nq4r30/138-setup-an-s3-bucket-to-store-assets

This commit adds a new infra-content-publisher project and corresponding
app user and private S3 bucket, which the user can write data into.


terraform/policies/content_publisher_s3_writer_policy.tpl
```diff
@@ -0,0 +1,28 @@
+{
+    "Version": "2018-08-20",
```
kevindew: This should be "Version": "2012-10-17", reference: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_version.html

benthorner:
@kevindew Should be ready to look at again :-).

PR #623, opened 2018-08-22
Increasing the size of the root partition of backend from 30G to 60G

szd55gds:
See Zendesk ticket #3107683

@szd55gds


szd55gds:
Requested by Ben (2nd line {publishing}).
Unable to free any large chunks of space and will slightly better than match current Carreza.

PR #625, opened 2018-08-24
Process Fastly logs with Athena daily for transition

barrucadu:
We'll need someone with admin powers to deploy this as it touches IAM, after which we can do the rest of the stuff on the card.

---

[Trello card](https://trello.com/c/dNtEfmJM/395-nothing-to-do-until-fri-meeting-move-transition-hit-stats-processing-to-s3-and-athena)


terraform/projects/infra-fastly-logs/main.tf
```diff
@@ -658,6 +696,100 @@ resource "aws_glue_catalog_table" "bouncer" {
   }
 }
 
+resource "aws_athena_named_query" "transition_logs" {
+  name     = "transition-logs-query"
+  database = "${aws_glue_catalog_database.fastly_logs.name}"
+
+  query = <<EOF
```
thomasleese: Can we put this query in a separate file maybe?

terraform/projects/infra-fastly-logs/main.tf
```diff
@@ -658,6 +696,100 @@ resource "aws_glue_catalog_table" "bouncer" {
...
+WHERE year = year(current_date - interval '1' day)
+  AND month = month(current_date - interval '1' day)
+  AND date = day(current_date - interval '1' day)
+GROUP BY cast(request_received as date), status, host, url
+ORDER BY 2 DESC
+EOF
+}
+
+resource "aws_lambda_function" "process_logs" {
+  filename      = "../../lambda/TransitionLogs/TransitionLogs.zip"
```
thomasleese: Can we get Terraform to generate this file?

barrucadu: We could, but the other lambda functions have zips checked into the repo as well

thomasleese: 😢

terraform/policies/transition_executor_policy.tpl
```diff
@@ -0,0 +1,53 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "logs:CreateLogGroup",
+        "logs:CreateLogStream",
+        "logs:PutLogEvents"
+      ],
+      "Resource": "arn:aws:logs:*:*:*",
```
deanwilson: I'm not familiar with Athena but the resource specifiers in this policy seem very broad. Is there a way to tighten them up?

terraform/projects/infra-fastly-logs/main.tf
deanwilson: https://github.com/alphagov/govuk-terraform-provisioning/pull/46#issuecomment-213586214 - Terraform maintainer :)

terraform/policies/transition_executor_policy.tpl
barrucadu: I based it on the [`AmazonAthenaFullAccess` policy](https://docs.aws.amazon.com/athena/latest/ug/access.html#amazonathenafullaccess-managed-policy), which is a generic policy, so that'll be why it's so broad.  I'll try to work out what it actually needs.

barrucadu: I don't think Athena and Glue do resource specifiers.  There's nothing about ARNs for them here: https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html and I can't find things in their own docs.  But I should be able to restrict the logs resource.

barrucadu: Although I've just noticed that other policies in this repo use the same log permissions:
```
$ git grep '"Resource": "arn:aws:logs:'
terraform/policies/govuk_artefact_policy.tpl:     "Resource": "arn:aws:logs:*:*:*"
terraform/policies/lambda_logs_to_firehose_policy.tpl:            "Resource": "arn:aws:logs:*:*:*"
terraform/policies/lambda_rds_logs_to_s3_policy.tpl:            "Resource": "arn:aws:logs:*:*:*"
terraform/policies/transition_executor_policy.tpl:      "Resource": "arn:aws:logs:*:*:*",
```

issyl0: If we do it already then I suppose follow that prior art. It'd be nice to know if there was a reason for it, back in the midsts of time. Maybe it's because they have *all* logs from a service (eg Lambda)? Which I suppose Athena does too. 🤔

barrucadu: I *think* the logs here are just for the Lambda function, but because there's no separation between the permissions the actual code of the Lambda needs and the permissions the Athena execution needs it's all a bit blurred together.

PR #632, opened 2018-09-04
Enable encryption on all EBS volumes, per environment

issyl0:
- This variable is set in `govuk-aws-data` common.tfvars per
  environment. This is because we don't want to enable this just yet in
  integration, but we do want it in staging and production because we're
  starting afresh with those environments.
- This was a recommendation from the AWS well-architected review.

https://trello.com/c/GsT3KAMg/1479-enable-encryption-at-rest-on-several-ebs-volumes


issyl0:
The related `govuk-aws-data` PR is [this one](https://github.com/alphagov/govuk-aws-data/pull/213).

PR #634, opened 2018-09-06
RDS: Default to keeping final snapshots

dougneal:
Change the default in the rds_instance module to be production-safe.


dougneal:
The change for integration is in my other branch, I'll pull it over now

issyl0:
I've never seen it done this way before (we usually [set variables in govuk-aws-data per-environment](https://github.com/alphagov/govuk-aws-data/pull/126)) - is this because it's a module? It seems more efficient than referring to another repo to get the values.

dougneal:
May as well do the same for all the databases, not just app-mysql ... don't merge yet.

PR #641, opened 2018-09-11
Change frontend instance type to t3.large

schmie:
- This was t2.medium and could not cope with an increase in worker procs
imposed by platform health.

- Switched to t3 for testing purposes as well

solo: @schmie


issyl0:
This has been applied to AWS Staging. And now Integration. Merging.

PR #670, opened 2018-10-02
Create Ubuntu Test class to test Xenial on.

stephenharker:
We need to find out how to proceed with Xenial and what Puppet changes, packaging, versions etc. need tweaking to get this to work. This would create an 'ubuntutest' VM in Integration to try this out on.


stephenharker:
Yes I basically copied and sedded the jumpbox class and tweaked a few other things to suit. So it will have direct access from the office as well.

PR #680, opened 2018-10-18
[DO NOT MERGE] Create buckets for primary and transition WAL-E backups

thomasleese:
This is based on the infra-wal-e-warehouse-bucket project. Only needed in production and staging as integration is using RDS.

This used to be configured in govuk-terraform-provisioning but we can't deploy that now, so I've created new buckets here and my plan is to configure WAL-E to use the new ones and copy the data across from the old ones.

[Trello Card](https://trello.com/c/DLAEfbPu/542-legacy-terraform-repo-missing-permissions-for-wal-e)

### DIFF
--- a/terraform/projects/infra-wal-e-primary-bucket/main.tf
+++ b/terraform/projects/infra-wal-e-primary-bucket/main.tf
@@ -1,0 +1,166 @@
+/**
+* ## Project: wal-e-primary-bucket
+*
+* This creates an s3 bucket
+*
+* wal-e-primary: The bucket that will hold database backups
+*
+*/
+
+variable "aws_region" {
+  type        = "string"
+  description = "AWS region"
+  default     = "eu-west-1"
+}
+
+variable "aws_backup_region" {
+  type        = "string"
+  description = "AWS region"
+  default     = "eu-west-2"
+}
+
+variable "aws_environment" {
+  type        = "string"
+  description = "AWS Environment"
+}
+
+variable "stackname" {
+  type        = "string"
+  description = "Stackname"
+}
+
+variable "remote_state_bucket" {
+  type        = "string"
+  description = "S3 bucket we store our terraform state in"
+}
+
+variable "remote_state_infra_monitoring_key_stack" {
+  type        = "string"
+  description = "Override stackname path to infra_monitoring remote state "
+  default     = ""
+}
+
+# Set up the backend & provider for each region
+terraform {
+  backend          "s3"             {}
+  required_version = "= 0.11.7"
+}
+
+provider "aws" {
+  region  = "${var.aws_region}"
+  version = "1.40.0"
+}
+
+provider "aws" {
+  alias   = "eu-london"
+  region  = "${var.aws_backup_region}"
+  version = "1.40.0"
+}
+
+data "terraform_remote_state" "infra_monitoring" {
+  backend = "s3"
+
+  config {
+    bucket = "${var.remote_state_bucket}"
+    key    = "${coalesce(var.remote_state_infra_monitoring_key_stack, var.stackname)}/infra-monitoring.tfstate"
+    region = "eu-west-1"
+  }
+}
+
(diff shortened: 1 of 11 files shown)


terraform/projects/infra-wal-e-transition-bucket/writer.tf
```diff
@@ -0,0 +1,56 @@
...
+  statement {
+    sid = "ReadBucketLists"
+
+    actions = [
+      "s3:ListBucket",
+      "s3:ListAllMyBuckets",
+      "s3:GetBucketLocation",
+    ]
+
+    # In theory  should work but it doesn't so use * instead
```
barrucadu: Is there a missing word between "theory" and "should"?

barrucadu:
The policy for the staging postgresql-primary WAL-E user, which works, is

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": [
                "s3:ListBucketMultipartUploads",
                "s3:ListBucketVersions",
                "s3:ListBucket",
                "s3:GetBucketLocation"
            ],
            "Resource": "arn:aws:s3:::govuk-wal-e-backups-postgresql-staging"
        },
        {
            "Sid": "VisualEditor1",
            "Effect": "Allow",
            "Action": [
                "s3:PutObject",
                "s3:GetObjectAcl",
                "s3:GetObject",
                "s3:AbortMultipartUpload",
                "s3:PutObjectAcl",
                "s3:GetObjectVersion",
                "s3:ListMultipartUploadParts"
            ],
            "Resource": "arn:aws:s3:::govuk-wal-e-backups-postgresql-staging/*"
        },
        {
            "Sid": "VisualEditor2",
            "Effect": "Allow",
            "Action": "S3:DeleteObject",
            "Resource": "arn:aws:s3:::govuk-wal-e-backups-postgresql-staging/*"
        }
    ]
}
```

I think the policies here should be based on (or just the same as) this

thomasleese:
@barrucadu I copied what we have for the warehouse database, which seems to be simply be `s3:*`.

https://github.com/alphagov/govuk-aws/pull/680/files#diff-50d70a75501daac5fec206611b8e29a1R32

barrucadu:
Yep, I just realised that - I think I misread the PR and thought you meant you'd based the policies on something in the old terraform.

This should be fine.

terraform/projects/infra-wal-e-transition-bucket/writer.tf
thomasleese: I'm not sure... we have the same line in the warehouse https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/infra-wal-e-warehouse-bucket/writer.tf#L26

I think there probably is a word missing, but I'm not sure what...

thomasleese: My guess is

> In theory referencing the bucket ARN should work but it doesn't so use * instead

What do you think?

barrucadu: That seems likely

terraform/projects/infra-wal-e-primary-bucket/main.tf
```diff
@@ -0,0 +1,166 @@
...
+  lifecycle_rule {
+    prefix  = "postgres/"
+    enabled = true
+
+    transition {
+      storage_class = "STANDARD_IA"
+      days          = 30
+    }
+
+    transition {
```
deanwilson: Do we need to archive wal-e logs off for long term storage?

thomasleese: It's not something we do at the moment, but I thought it would be worth having for parity with the warehouse WAL-E logs (which do archive them off to long term storage).

thomasleese:
Closing until we need this.

PR #682, opened 2018-10-22
Fix govukcli bug

elliotcm:
The escaping here seems to be insufficient to
prevent the script from trying to run
govuk_node_list locally rather than running it
on the jumpbox.

### DIFF
--- a/tools/govukcli
+++ b/tools/govukcli
@@ -246,7 +246,7 @@ function run_ssh {
           ssh -At $SSH_OPTS $SSH_USER@$JUMPBOX "ssh -q ${NODE_CLASS}"
         else
           # shellcheck disable=SC2029
-          ssh -At $SSH_OPTS $SSH_USER@$JUMPBOX "ssh -q \`govuk_node_list --single-node -c ${NODE_CLASS}\`"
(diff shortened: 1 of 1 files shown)


th31nitiate:
I am not too sure how to test this one. Please can you confirm that you done the appropriate level of testing ? @elliotcm

elliotcm:
@th31nitiate It's a good question and one I'm not sure about.  Really this is a "it works on my machine" situation, so really we'd want to rope in as many different devs as possible to try the branch on their machines.

This PR is also a bit of a discussion point since I had the following concerns about it that I put in Slack (and really should've put in this PR):

> I do wonder a) why has this suddenly broken, b) is it broken for other people, and c) why aren’t we re-using the ec2 instance check further up in the script

th31nitiate:
@elliotcm I have managed to execute the script on my local machine from the master branch. I have also achieve the same results on this current branch. 

I assume the issue is that the machine OS or shell you are using is some special in someway. I am happy to approve this in order to ensure that everyone is able to use it. I am on mac os running zsh. I have also tested on bash and it is all good.

PR #683, opened 2018-10-22
Enable HTTPS and HTTP redirect for bouncer 

th31nitiate:
This PR adds the ability for bouncer to work on both HTTP and HTTPs endpoints.

We encountered an issue where when we defined a HTTP and HTTPS listener. The terraform module was not able to perform create on the listener since it attempted to add a certificate to both listener. This was not what we intended and also an error was reported by terraform. 

We have now added an ADR explain our changes and why we made those changes. To review the solution please review the ADR.


fredericfran-gds:
close pull request because changes were added in PR: https://github.com/alphagov/govuk-aws/pull/684 due to github operational issues

PR #685, opened 2018-10-22
Add port 80 bouncer load balancer

rtrinque:
Add port 80 as a new listener for the bouncer load balancer

2 bugs have been found in the "aws/lb" terraform module and fixed:

ssl_policy is set for non-ssl listener. an option is set by its mere presence in terraform resource. Hence, there is a need to have 2 listener resource definitions: one for ssl and one for non-ssl.
secondary certificates were add to all listeners including the non-ssl listeners


ronocg:
Closing this as the tests did not run... reopening in a min...

PR #690, opened 2018-10-23
GovUK-Migr-Trello-16: Fix CNAMES for monitoring

ronocg:
alert and alert.cluster CNAMES should point to
alert.blue.'env'.govuk-internal.digital

But due to the way the code is setup alert will point correctly but
alert.cluster will end up pointing to alert.cluster.'env'.govuk-internal.digital
which does not go anywhere.

This fix will get both records pointing into the correct destination.

Solo: @ronocg

### DIFF
--- a/terraform/projects/infra-public-services/README.md
+++ b/terraform/projects/infra-public-services/README.md
@@ -58,6 +58,7 @@ This project adds global resources for app components:
 | mapit_public_service_names |  | list | `<list>` | no |
 | mongo_internal_service_names |  | list | `<list>` | no |
 | monitoring_internal_service_names |  | list | `<list>` | no |
+| monitoring_internal_service_names_cname_dest | This variable specifies the CNAME record destination to be associated with the service names defined in monitoring_internal_service_names | string | `alert` | no |
 | monitoring_public_service_names |  | list | `<list>` | no |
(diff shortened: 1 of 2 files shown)


th31nitiate:
Ok, correct if I am wrong but my understanding of this is that we want alert.cluster.blue.blah to point to the same location as alert.blue.blah. They should both resolve to the same host. This is what the current PR does.

terraform/projects/infra-public-services/README.md
```diff
@@ -57,6 +57,7 @@ This project adds global resources for app components:
 | mapit_internal_service_names |  | list | `<list>` | no |
 | mongo_internal_service_names |  | list | `<list>` | no |
 | monitoring_internal_service_names |  | list | `<list>` | no |
+| monitoring_internal_service_names_cname_dest |  | string | `alert` | no |
```
fredericfran-gds: an improvement would be to put in the description cell of this row:
"this variable specifies the CNAME subdomain to be associated with the service names specified in `monitoring_internal_service_names `"

th31nitiate: I think this may be auto generated by https://github.com/segmentio/terraform-docs

It is essence takes a the full variable or resource and prints that in to the markdown file.

PR #705, opened 2018-11-27
Add explicit PSK for AWS production VPN to Carrenza

schmie:
- Carrenza expects 32 digit alphanumerical PSK
- Carrenza needed to configure a single subnet of /20 range in order for
the VPN to work AWS side

aws-data part: https://github.com/alphagov/govuk-aws-data/pull/256

pair: @schmie @stephenharker


schmie:
Fixed whitespace errors and automated docs update only

PR #706, opened 2018-11-27
Add SG to permit access link-checker-api

th31nitiate:
- Add access to AWS for HTTP, HTTPS and ICMP. This should permit on-prem subnets.


terraform/projects/infra-security-groups/vpn.tf
```diff
@@ -0,0 +1,42 @@
+resource "aws_security_group" "vpn" {
+  name        = "${var.stackname}_vpn"
+  vpc_id      = "${data.terraform_remote_state.infra_vpc.vpc_id}"
+  description = "This is the group use to access AWS from Carrenza"
```
deanwilson: `used`, not `use` in this context

PR #716, opened 2018-12-10
Remove aws_security_group_rules that are not used

jstandring-gds:
# Context
The security group, aws_security_group.monitoring_external_elb.id, is applied
to loadbalancers that are only listening on port 443 and have members that
accept traffic on port 80. The high port numbers in these rules allow traffic
through but there is nothing is configured to listen.

This was discovered through the trusted advisor report in production
(https://console.aws.amazon.com/trustedadvisor/home?#/category/security).

# Decision 
Remove the redundant rules.


ronocg:
I just checked the current monitoring/alert instance in production carrenza and it does listen on 1614/5/6 for syslog coming in (from external fastly)..
So Im sure these rules will need to be in place and we will have to open up those ports on the load balancer.

PR #718, opened 2018-12-11
Add deploy script

rubenarakelyan:
This commit adds a Ruby deploy script to simplify the process of deploying Terraform.

The script accepts the same input as the Jenkins job, and also requires GitHub credentials.

It then uses `govukcli` (in the same directory) to generate temporary AWS credentials. The GitHub and AWS credentials, along with the initial command line input and a Jenkins “crumb” are used to queue the the `Deploy_Terraform_GOVUK_AWS` job in Jenkins.

Using Jenkins allows deployments to be auditable.

### DIFF
--- a/tools/deploy.rb
+++ b/tools/deploy.rb
@@ -1,0 +1,83 @@
+#!/usr/bin/env ruby
+
+require 'net/http'
+require 'net/https'
+require 'json'
+require 'uri'
+
+# Fields from the command line
+command = ARGV[0]
+environment = ARGV[1]
+stack = ARGV[2]
+project = ARGV[3]
+
+# Valid values for each field
+valid_commands = %w(plan apply destroy).freeze
+valid_environments = %w(integration staging production test).freeze
+valid_stacks = %w(blue green govuk).freeze
+
+usage = 'Usage: GITHUB_USERNAME=... GITHUB_TOKEN=... ruby deploy.rb <command> <environment> <stack> <project>'
+
+abort("GITHUB_USERNAME environment variable must be set\n#{usage}") unless ENV.has_key?('GITHUB_USERNAME')
+abort("GITHUB_TOKEN environment variable must be set\n#{usage}") unless ENV.has_key?('GITHUB_TOKEN')
+abort("command must be one of #{valid_commands.join(', ')}\n#{usage}") unless valid_commands.include?(command)
+abort("environment must be one of #{valid_environments.join(', ')}\n#{usage}") unless valid_environments.include?(environment)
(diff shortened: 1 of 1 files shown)


tools/deploy.rb
```diff
@@ -0,0 +1,72 @@
...
+# Jenkins details
+jenkins_url = 'https://ci-deploy.integration.publishing.service.gov.uk'.freeze
+jenkins_crumb_issuer_path = '/crumbIssuer/api/json'.freeze
+jenkins_job_path = '/job/Deploy_Terraform_GOVUK_AWS/buildWithParameters'.freeze
+jenkins_crumb_issuer_uri = URI.parse("#{jenkins_url}#{jenkins_crumb_issuer_path}")
+jenkins_job_uri = URI.parse("#{jenkins_url}#{jenkins_job_path}")
+
+# Get temporary AWS credentials
+puts 'Requesting temporary AWS credentials...'
+env = `govukcli aws invoke printenv`
```
barrucadu: I think the script should terminate here if this returns a nonzero status code

tools/deploy.rb
```diff
@@ -0,0 +1,72 @@
...
+env.each_line do |env_var|
+  key, value = env_var.split('=')
+  ENV[key] = value.chomp if key.start_with?('AWS_')
+end
```
barrucadu: I think we should instead explicitly set the needed environment variables (and bail out if not present in the output), rather than just copy cross all `AWS_` outputs.

tools/deploy.rb
```diff
@@ -0,0 +1,72 @@
...
+abort('Could not get crumb from Jenkins') unless jenkins_crumb_response.code == '200'
+jenkins_crumb = JSON.parse(jenkins_crumb_response.body)
+
+# Make a request to the Jenkins API to queue the build
+puts 'Queuing Jenkins job...'
+jenkins_job_http = Net::HTTP.new(jenkins_job_uri.host, jenkins_job_uri.port)
+jenkins_job_http.use_ssl = true
+jenkins_job_request = Net::HTTP::Post.new(jenkins_job_uri.path)
+jenkins_job_request.basic_auth(ENV['GITHUB_USERNAME'], ENV['GITHUB_TOKEN'])
+jenkins_job_request.set_form_data({
```
barrucadu: Is it worth printing all the parameters and asking the user for confirmation before starting the job?

rubenarakelyan: I've added a confirmation around the start of the script

tools/deploy.rb
rubenarakelyan: Done

tools/deploy.rb
rubenarakelyan: This now explicitly sets the three AWS related env vars

PR #720, opened 2018-12-13
Add S3 bucket and IAM user for support-api

rubenarakelyan:
This commit adds a new project which creates an S3 bucket and IAM user for support-api. This bucket will be used to store CSV files generated by support-api in reponse to requests from the support app.


terraform/projects/infra-support-api/main.tf
```diff
@@ -0,0 +1,80 @@
...
+}
+
+resource "aws_iam_policy_attachment" "s3_writer" {
+  name       = "archive-writer-policy-attachment"
+  users      = ["${aws_iam_user.support_api_app.name}"]
+  policy_arn = "${aws_iam_policy.s3_writer.arn}"
+}
+
+data "template_file" "s3_writer_policy_template" {
+  template = "${file("${path.module}/../../policies/support_api_s3_writer_policy.tpl")}"
```
katstevens: Does this policy file already exist, or is it generated somewhere?

rubenarakelyan: My bad, I forgot to copy it across!

terraform/policies/support_api_s3_writer_policy.tpl
```diff
@@ -0,0 +1,29 @@
...
+             ],
+            "Resource": [
+                "arn:aws:s3:::*"
+            ]
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:GetObject",
+                "s3:ListBucket",
```
barrucadu: Is `s3:ListBucket` needed here?  Isn't this covered by the section above?

terraform/projects/infra-support-api/main.tf
```diff
@@ -0,0 +1,80 @@
...
+provider "aws" {
+  region  = "${var.aws_region}"
+  version = "1.40.0"
+}
+
+resource "aws_s3_bucket" "support_api_csvs" {
+  bucket = "govuk-${var.aws_environment}-support-api-csvs"
+
+  tags {
+    Name            = "govuk-${var.aws_environment}-support-api-csvs"
```
barrucadu: Should this be capitalised?

terraform/policies/support_api_s3_writer_policy.tpl
rubenarakelyan: I think so. I've removed it from here but may need to add it back in.

terraform/projects/infra-support-api/main.tf
rubenarakelyan: Nope, copy and paste from someone else's code!

PR #724, opened 2018-12-14
 Migration-plan for feedback

ronocg:
Traffic from Carrenza to the feedback service in AWS will need to
traverse the VPN, therefore we cannot use an external load balancer to
load balance the traffic into the frontend machines in AWS (feedback
lives on the frontend).

So we need to use the existing internal load balancer that the frontends
have, so in order to use that we need a dns record in the external
govuk.digital zone. This puts that in place.

Solo: @ronocg <conor.glynn@digital.cabinet-office.gov.uk>


ronocg:
We are scrapping this and we are going to roll out a new external LB for the feedback service.

PR #731, opened 2019-01-08
BAU: adding a count for this resource

ronocg:
This resource is not really required for integration but to keep the
code and environments as consistent as possible, this code will execute
for the integration environment. So we are placing a count on the
resource so if the variable does not exist then the rule will not be created.

@jstandring-gds <julian.standring@digital.cabinet-office.gov.uk>
@ronocg <conor.glynn@digital.cabinet-office.gov.uk>


ronocg:
good spot, I have pushed up that change

PR #735, opened 2019-01-10
Add infra-assets to replace legacy terraform

benthorner:
https://trello.com/c/NTr4JsF2/513-investigate-syncing-s3-production-to-staging-integration
Depends on: https://github.com/alphagov/govuk-aws-data/pull/285

This adds a new project to move control of the govuk-assets-xxx and
govuk-assets-backup-xxx buckets out of our legacy terraform repo:

   * An app user with r/w access to the govuk-assets-xxx bucket
   * A cross-account policy for env-sync (except on prod)
   * A backup role with access to the govuk-assets-backup-xxx bucket

Most of the resources have been renamed to conform to the govuk-xxx-name
convention we now have, but to avoid downtime, we need to import:

   * govuk-assets-xxx
   * govuk-assets-backup-xxx
   * govuk-assets-production-user

Once this has been applied, the other resources that use legacy naming
can be detached/removed, namely:

   * asset-manager-user-iam-policy
   * asset-manager-backup-iam-role
   * asset-manager-backup-iam-policy
   * asset-manager-env-sync-bucket (if it still exists)
   * asset-manager-env-sync-user (policy)
   * asset-manager-env-sync-user (actually a user this time ;-)

Note that the backup configuration will now be applied in each
environment (not just in prod) to make them more consistent.


benthorner:
Link to the original project: https://github.com/alphagov/govuk-terraform-provisioning/tree/master/projects/asset-manager/resources

terraform/projects/infra-assets/main.tf
```diff
@@ -0,0 +1,95 @@
...
+  replication_configuration {
+    role = "${aws_iam_role.backup.arn}"
+
+    rules {
+      id     = "govuk-${var.aws_environment}-assets-replication-rule"
+      prefix = ""
+      status = "Enabled"
+
+      destination {
+        bucket        = "${aws_s3_bucket.assets_backup.arn}"
```
schmie: This provokes an error in AWS prod: 
```
Error: Error running plan: 1 error(s) occurred:

* aws_s3_bucket.assets: "replication_configuration.0.rules.0.destination.0.bucket" doesn't look like a valid ARN ("^arn:[\\w-]+:([a-zA-Z0-9\\-])+:([a-z]{2}-(gov-)?[a-z]+-\\d{1})?:(\\d{12})?:(.*)$"): "govuk-assets-backup-production"
```
The syntax looks correct to me (a bug in the .arn method?!)... I assume using `arn:aws:s3:::${aws_s3_bucket.assets_backup.id}` is a viable work around.

benthorner: Interesting - we don't get this error. Just to check: do you have the most recent pull of govuk-aws-data?

rtrinque: Runs fine for me with last pull

PR #737, opened 2019-01-11
Allow passing the context to govukcli ssh

cbaines:
Builds on top of https://github.com/alphagov/govuk-aws/pull/736

These changes mean that the following command works for example:

  govukcli ssh --context staging-aws backend

I personally would like the govukcli ssh behaviour to be more
predictable, based on the command being run. So rather than relying on
the correct context being set, instead, passing it in every time and
being explicit. I think this is easier to understand and use, as well
as being less error prone.

### DIFF
--- a/tools/govukcli
+++ b/tools/govukcli
@@ -104,7 +104,7 @@ function error {
 function check_allowed_contexts {
   if [[ ! $CONTEXT =~ $ALLOWED_CONTEXTS ]]; then
     if [[ $OUT != 'silent' ]]; then
-      error "Invalid context"
+      error "Invalid context '$CONTEXT'"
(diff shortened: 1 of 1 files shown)


cbaines:
I'm a bit torn with `govukcli`. To enable the `govuk data-standalone` script within govuk-guix, I've already implemented a similar thing for just the aws component [1]. I'm tempted to do similarly for the SSH component, and write a standalone script in either Ruby or Bash to do similar stuff. I haven't been using `govukcli ssh` up until now, but at the moment, I'm going to try to use and improve it, and see how it goes...

1: https://github.com/alphagov/govuk-guix/blob/master/bin/govuk-aws

PR #738, opened 2019-01-15
Add a security group for Gatling

thomasleese:
We will start Gatling FrontLine instances using this security group.

It allows inbound traffic on port 22 and 80 and all outbound traffic.

[Trello Card](https://trello.com/c/kxP64gJo/700-redo-the-load-test-to-see-what-impact-weve-had-with-our-changes)

### DIFF
--- a/terraform/projects/infra-security-groups/gatling.tf
+++ b/terraform/projects/infra-security-groups/gatling.tf
@@ -1,0 +1,57 @@
+#
+# == Manifest: Project: Security Groups: gatling
+#
+# Gatling needs to be accessible on ports:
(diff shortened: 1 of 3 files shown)


terraform/projects/infra-security-groups/gatling.tf
```diff
@@ -0,0 +1,57 @@
+#
+# == Manifest: Project: Security Groups: gatling
+#
+# The db-admin needs to be accessible on ports:
```
rubenarakelyan: `db-admin` => `gatling`

PR #741, opened 2019-01-15
Trello-1903: Removing prometheus elb

ronocg:
The prometheus elb is not required, there is an existing ALB that will
do the job it is spun up in infra-public-services.

Also allowing from the alb to the instance on port 80.

Putting config in place to roll prometheus to staging and production
also.

@ronocg <conor.glynn@digital.cabinet-office.gov.uk>
@jstandring-gds <julian.standring@digital.cabinet-office.gov.uk>


ronocg:
I have made those changes that you have suggested.
The module name has a "-1" as I think there could be multiple prometheis in the future, that was the idea... not sure as I inherited this code, I think it is ok tho...

PR #754, opened 2019-01-24
add draft_whitehall_frontend cnaming to whitehall_frontend

fredericfran-gds:
# Context

A interruptibe task was raised by gov.uk 2nd line (@kevindew) to cname the draft-whitehall-frontend internal dns name to the whitehall-frontend internal dns name.

# Decisions:
1. Modify the infra-public-services terraform project's logic to reflect this and then also modify the corresponding terraform aws data: https://github.com/alphagov/govuk-aws-data/pull/301

### DIFF
--- a/terraform/projects/infra-public-services/main.tf
+++ b/terraform/projects/infra-public-services/main.tf
@@ -262,6 +262,11 @@ variable "draft_frontend_internal_service_cnames" {
   default = []
 }
 
+variable "draft_whitehall_frontend_internal_service_names" {
+  type    = "list"
+  default = []
+}
+
 variable "email_alert_api_internal_service_names" {
   type    = "list"
   default = []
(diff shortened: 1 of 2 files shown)


terraform/projects/infra-public-services/main.tf
```diff
@@ -1943,6 +1948,16 @@ resource "aws_route53_record" "whitehall_backend_internal_service_cnames" {
 # whitehall_frontend
 #
 
+# draft_whitehall internal names are actually alias for whitehall internal names
+resource "aws_route53_record" "draft_whitehall_frontend_internal_service_names" {
+  count   = "${length(var.draft_whitehall_frontend_internal_service_names)}"
+  zone_id = "${data.terraform_remote_state.infra_root_dns_zones.internal_root_zone_id}"
+  name    = "${element(var.draft_whitehall_frontend_internal_service_names, count.index)}.${data.terraform_remote_state.infra_root_dns_zones.internal_root_domain_name}"
+  type    = "CNAME"
+  records = ["${element(var.whitehall_frontend_internal_service_names, count.index)}.blue.${data.terraform_remote_state.infra_root_dns_zones.internal_root_domain_name}"]
```
schmie: This rather needs `${element(var.draft_whitehall_frontend_internal_service_names, count.index)}`

cbaines:
So I was thinking about this a while ago, I made a few changes:

 - https://github.com/alphagov/govuk-puppet/pull/7795
 - https://github.com/alphagov/govuk-aws-data/pull/182

I had trouble deploying the necessary changes in AWS, so the work was left unfinished, but it would be interesting to know what was insufficient about these changes https://github.com/alphagov/govuk-aws-data/pull/182

terraform/projects/infra-public-services/main.tf
fredericfran-gds: hi this is fine. 
we want draft_whitehall_frontend to point to whitehall_frontend so satisfy Kevin's requirements.

fredericfran-gds:
@cbaines the logic inside the infra-public-services is wrong unfortunately. if you read it carefully you will find that the original code was cnaming draft_whitehall_frontend to draft_whitehall_frontend_blue which does not correct.

PR #756, opened 2019-01-29
set healthcheck to /healthcheck to match app endpoint

fredericfran-gds:
# Context

We need to test the healthcheck of the app running via nginx rather than nginx itself. Currently nginx provides a health check endpoint on URL `/_healthcheck` but the app will provide a health check endpoint on `/healthcheck`

# Decisions
1. change the health endpoint url from `/_healthcheck`to `/healthcheck`


fredericfran-gds:
might not be needed

PR #790, opened 2019-02-13
Add www to non-www redirect for data.gov.uk

rubenarakelyan:
data.gov.uk is the canonical host name, so we should redirect the www version.


elliotcm:
Linting, bruv

rubenarakelyan:
Fixed m8

PR #791, opened 2019-02-13
Switch data science EC2 instance type to p2.xlarge

bilbof:
https://trello.com/c/52Lcfble/359-migrate-to-p2xlarge-aws-instance-to-produce-related-links-using-graphsage

This is part of the related links project.


rubenarakelyan:
For the moment, we will need to spin these up manually until I can sort out SSH access, but good to have in here anyway for when we're ready.

PR #807, opened 2019-02-20
Publish the Elasticsearch logs to Cloudwatch

brucebolt:
AWS offer three types of logs for the managed Elasticsearch service:

- Application error logs
- Slow query logs
- Slow index logs

This PR exports all three types to Cloudwatch, with a retention period of 90 days.  Further work will deal with the exporting of these to S3.

Trello card: https://trello.com/c/jc1LM6nM/58-set-up-export-of-error-logs-from-elasticsearch-to-cloudwatch

### DIFF
--- a/terraform/projects/app-elasticsearch5/main.tf
+++ b/terraform/projects/app-elasticsearch5/main.tf
@@ -73,6 +73,21 @@ resource "aws_iam_service_linked_role" "role" {
   aws_service_name = "es.amazonaws.com"
 }
 
+resource "aws_cloudwatch_log_group" "elasticsearch5_application_log_group" {
+  name              = "/aws/aes/domains/${var.stackname}-elasticsearch5-domain/application-logs"
+  retention_in_days = "90"
+}
+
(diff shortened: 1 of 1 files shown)


terraform/projects/app-elasticsearch5/main.tf
```diff
@@ -73,6 +73,21 @@ resource "aws_iam_service_linked_role" "role" {
   aws_service_name = "es.amazonaws.com"
 }
 
+resource "aws_cloudwatch_log_group" "elasticsearch5_application_log_group" {
+  name              = "/aws/aes/domains/${var.stackname}-elasticsearch5-domain/application-logs"
```
barrucadu: Is it possible to refer to the `domain_name` directly here, rather than `${var.stackname}-elasticsearch5-domain`?

barrucadu: Hmm, maybe not, as then there would be a cyclic dependency between the resource for the domain and the resource for the log group...

PR #818, opened 2019-02-21
[WIP] Add S3 bucket for Fastly Do Not Track header logs

erino:
This bucket is to collect a few days worth of DNT headers for requests on GOV.UK. If a lot of browsers are using it then we can make the case for building support on GOV.UK

1. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/DNT


erino:
We're not going to collect this data

PR #826, opened 2019-02-25
Add database infrastructure for the Content Data API

cbaines:
The Content Performance Manager application is being renamed to the
Content Data API to better reflect it's current function.

Currently the Content Performance Manager uses the "warehouse"
infrastructure. The plan is to make the resource naming more
consistent by following the patterns for other applications like the
Publishing API.

The Terraform configuration added in this commit relates to the new
resources for the Content Data API. It's basically a copy of the
"warehouse" configuration.

Once the migration from the Content Performance Manager to the Content
Data API has been completed, the "warehouse" resources can be removed.


jenkins.sh
```diff
@@ -41,7 +41,7 @@ fi
 rm -rf govuk-aws-data
 
 echo "Cloning govuk-aws-data"
-git clone git@github.com:alphagov/govuk-aws-data.git
+git clone git@github.com:alphagov/govuk-aws-data.git -b add-database-infrastructure-configuration-for-the-content-data-api
```
fredericfran-gds: we are going to commit into master so remove this?

terraform/projects/app-content-data-api-db-admin/main.tf
```diff
@@ -0,0 +1,163 @@
...
+  internal        = "true"
+
+  access_logs {
+    bucket        = "${data.terraform_remote_state.infra_monitoring.aws_logging_bucket_id}"
+    bucket_prefix = "elb/${var.stackname}-content-data-api-db-admin-internal-elb"
+    interval      = 60
+  }
+
+  listener {
+    instance_port     = 22
```
fredericfran-gds: is this ok? looks like 22 is usually used for ssh

terraform/projects/app-content-data-api-postgresql/README.md
```diff
@@ -0,0 +1,37 @@
+## Project: projects/app-content-data-api-database
```
fredericfran-gds: why is this name different from the directory name?

terraform/projects/app-content-data-api-postgresql/main.tf
```diff
@@ -0,0 +1,180 @@
...
+  version = "1.40.0"
+}
+
+module "content-data-api-postgresql-primary_rds_instance" {
+  source = "../../modules/aws/rds_instance"
+
+  name                = "${var.stackname}-content-data-api-postgresql-primary"
+  engine_name         = "postgres"
+  engine_version      = "9.6"
+  default_tags        = "${map("Project", var.stackname, "aws_stackname", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "postgresql_primary")}"
```
fredericfran-gds: I'm not use about the tag postgresql_primary, maybe prepend the `content-data-api` to it.

terraform/projects/app-content-data-api-postgresql/main.tf
```diff
@@ -0,0 +1,180 @@
...
+  zone_id = "${data.terraform_remote_state.infra_stack_dns_zones.internal_zone_id}"
+  name    = "content-data-api-postgresql-primary.${data.terraform_remote_state.infra_stack_dns_zones.internal_domain_name}"
+  type    = "CNAME"
+  ttl     = 300
+  records = ["${module.content-data-api-postgresql-primary_rds_instance.rds_instance_address}"]
+}
+
+module "content-data-api-postgresql-standby_rds_instance" {
+  source = "../../modules/aws/rds_instance"
+
```
fredericfran-gds: awkward the linting did not detect and fix these: `=` not aligned and empty between parameters

terraform/projects/app-content-data-api-postgresql/main.tf
```diff
@@ -0,0 +1,180 @@
...
+  name                       = "${var.stackname}-content-data-api-postgresql-standby"
+  default_tags               = "${map("Project", var.stackname, "aws_stackname", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "postgresql_standby")}"
```
fredericfran-gds: same comment as before about tags: `"postgresql_standby"`

jenkins.sh
cbaines: Yeah, this was only so I could test on Integration. I've removed it now.

terraform/projects/app-content-data-api-db-admin/main.tf
cbaines: I realise it isn't exactly set out in the Terraform configuration, but as the DB Admin machine is used for administration, SSH is really the only service it provides.

I'm not certain, but I think you might be able to SSH through the ELB, to the actual DB Admin machine.

In reality, this is here because I copied it from the app-warehouse-db-admin project.

terraform/projects/app-content-data-api-postgresql/README.md
cbaines: Good spot :) This is just me not putting the correct name in when copying/pasting.

terraform/projects/app-content-data-api-postgresql/main.tf
cbaines: Indeed, I'm not sure where that came from. I've changed it to `content_data_api_postgresql_primary` now.

terraform/projects/app-content-data-api-postgresql/main.tf
cbaines: I'm guessing that the linter perhaps only considers blocks of parameters without newlines between them. I'll remove the newline, and correct the indentation.

terraform/projects/app-content-data-api-postgresql/main.tf
cbaines: Same as above, I'll remove the newline and correct the indentation.

terraform/projects/app-content-data-api-db-admin/main.tf
fredericfran-gds: cool. So the users of the db_admin machines of content-api will use the load balancer as the entry point to run commands via ssh on those. I guess you have more than 1 db_admin machines for content-api?

terraform/projects/app-content-data-api-postgresql/main.tf
```diff
@@ -0,0 +1,178 @@
...
+  name    = "content-data-api-postgresql-primary.${data.terraform_remote_state.infra_stack_dns_zones.internal_domain_name}"
+  type    = "CNAME"
+  ttl     = 300
+  records = ["${module.content-data-api-postgresql-primary_rds_instance.rds_instance_address}"]
+}
+
+module "content-data-api-postgresql-standby_rds_instance" {
+  source                     = "../../modules/aws/rds_instance"
+  name                       = "${var.stackname}-content-data-api-postgresql-standby"
+  default_tags               = "${map("Project", var.stackname, "aws_stackname", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "postgresql_standby")}"
```
fredericfran-gds: Did you miss this tag update here for : `postgresql_standby` to `content_data_api_postgresql_standby`?

cbaines: Ah, yeah, I didn't read your comment here [1] properly, and thought it was about the `source = ` indentation issue. I'll fix this now...

1: https://github.com/alphagov/govuk-aws/pull/826#discussion_r263439166

terraform/projects/app-content-data-api-db-admin/main.tf
cbaines: There's just one, and I think that's ok. I think for the other RDS instances, they also only have one DB Admin instance.

I've made sure with the Content Data API, that this DB Admin instance is only used for administration tasks (e.g. data syncs), so I think it's fine just having the one instance.

I'm not entirely sure why there's an ELB, that's just copied from the warehouse configuration.

PR #829, opened 2019-02-28
Add ALB for search-api external access, use a CNAME for internal access

barrucadu:
This PR removes the search-api internal service name, that should more properly be a CNAME pointing to the "search" hostname, as search-api lives on the app-search machines (just like rummager).

It also adds an ALB so that Carrenza IPs can talk to search-api over the public internet, which is needed for:

1. Traffic replay from rummager to search-api while we have both running in parallel
2. Direct communication from search-admin and whitehall when rummager is switched off

govuk-aws-data PR: https://github.com/alphagov/govuk-aws-data/pull/339

---

[Trello card](https://trello.com/c/YU9bS9jR/76-get-search-api-running-in-aws-staging)


brucebolt:
Everything looks ok to me, but might be worth asking someone in RE to take a look, since they are far more knowledgable than me about the Carrenza/AWS link.

PR #831, opened 2019-03-05
Change the deploy.rb script to use `govuk aws`

cbaines:
The `govuk aws` script provides some advantages over the `govukcli`
script currently being used. Using it doesn't require changing the
context set for govukcli, and it provides a JSON export which takes
less code to process the output.

### DIFF
--- a/tools/deploy.rb
+++ b/tools/deploy.rb
@@ -39,16 +39,11 @@ jenkins_job_uri = URI.parse("#{jenkins_url}#{jenkins_job_path}")
 
 # Get temporary AWS credentials
 puts 'Requesting temporary AWS credentials...'
-`govukcli set-context #{environment}`
-env = `govukcli aws invoke printenv`
+
+aws_credentials_string = `govuk aws --profile govuk-#{environment} --export-json`
 abort('Could not get temporary AWS credentials') unless $?.exitstatus.zero?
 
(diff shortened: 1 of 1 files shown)


cbaines:
> Is the `govuk aws` tool this one: https://github.com/alphagov/govuk-guix/blob/6d8797df5ff4e4f3cff23d150c050c4ced07c240/bin/govuk-aws?

Yep.

> If so, that's not an official script and therefore we shouldn't be calling it from here.

Ah, ok. While I realise this change adds a dependency, I only proposed it for the reasons I gave in the commit message. I didn't realise there was a officiation process for scripts.

> You're more than welcome to propose changes to the `govukcli` script to make it better, but we shouldn't be relying on scripts maintained by a single person originating in a repo marked as "personal project" for things like deploying all our AWS Terraform.

I wouldn't read too much in to the GitHub description, I don't know where that came from, and I've mostly left it there because of that.

As for changing the `govukcli` script, I don't particularly like the way it handles contexts in a "global variable" style. I've tried to workaround this for the SSH component, but it wasn't particularly elegant [1], so I quite like having a smaller script that just does things relating to AWS and that handles profiles in a less stateful way.

1: https://github.com/alphagov/govuk-aws/pull/737

I don't use this script, I just thought this change might be helpful for those who do. It looks from the Git history like your the sole author though, so I'll trust your judgement.

PR #833, opened 2019-03-05
Set up a location for elasticsearch 5 backups in s3

barrucadu:
This will be used by the env sync script in puppet.

---

[Trello card](https://trello.com/c/QQHOpt56/75-update-data-sync-for-elasticsearch-5)


barrucadu:
Going to do this with snapshots instead, which means we need a more annoying S3 set-up.

PR #835, opened 2019-03-05
Add CORS config for s3 mirrors

jstandring-gds:
Context
We are seeing CORS errors when the s3 mirrors are in use, i.e. when fastly
health checks fail and traffic is redirected to the s3 bucket we see
Cross-Origin Resource Sharing errors in the browser.

Decision
Add the allowed origins to the s3 bucket config.


ronocg:
Closing this, as we carried out further testing in Staging and the reason why the s3 asset mirrors did not work was that the backedn s3 mirror was set incorrectly - closing

PR #836, opened 2019-03-05
Fix the jenkins.sh script for a persistent workspace

cbaines:
Now that the Jenkins job definition has been changed in govuk-puppet
to not wipe the workspace, this directory needs deleting so that it
can be cloned.

It's probably possible to do this more efficiently, but this is a
simple way of making it work again.


cbaines:
This relates to the change here https://github.com/alphagov/govuk-puppet/pull/8773

PR #871, opened 2019-03-27
Remove CSW target account ID

tlwr:
The latest version of CSW will not need target account ID.

Related PRs:

- https://github.com/alphagov/csw-client-role/pull/14
- https://github.com/alphagov/govuk-aws-data/pull/358


tlwr:
Just ran `tf fmt` and force pushed

PR #873, opened 2019-03-27
Fix Pingdom probe sh script

jstandring-gds:
Context
The probe to determine pingdom ips doesn't filter out "NULL" entries.

Decision
Remove the invert-match ("-v") and add regex specifically for IPv4 addresses.
(regex ref: https://www.shellhacks.com/regex-find-ip-addresses-file-grep/)

### DIFF
--- a/terraform/projects/infra-s3-mirrors/pingdom_probe_ips.sh
+++ b/terraform/projects/infra-s3-mirrors/pingdom_probe_ips.sh
@@ -11,7 +11,7 @@
 # https://www.terraform.io/docs/providers/external/data_source.html
 # https://github.com/hashicorp/terraform/issues/12256
 
-curl -s https://my.pingdom.com/probes/feed | grep "pingdom:ip" | sed -e 's|</.*||' -e 's|.*>||' | sort -n -t . -k 1,1 -k 2,2 -k 3,3 -k 4,4 | grep -v ":" | awk '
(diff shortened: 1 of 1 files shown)


terraform/projects/infra-s3-mirrors/pingdom_probe_ips.sh
```diff
@@ -11,7 +11,7 @@
 # https://www.terraform.io/docs/providers/external/data_source.html
 # https://github.com/hashicorp/terraform/issues/12256
 
-curl -s https://my.pingdom.com/probes/feed | grep "pingdom:ip" | sed -e 's|</.*||' -e 's|.*>||' | sort -n -t . -k 1,1 -k 2,2 -k 3,3 -k 4,4 | grep -v ":" | awk '
+curl -s https://my.pingdom.com/probes/feed | grep "pingdom:ip" | sed -e 's|</.*||' -e 's|.*>||' | sort -n -t . -k 1,1 -k 2,2 -k 3,3 -k 4,4 | grep -vE '[A-Z]|[a-z]|:' | awk '
```
fredericfran-gds: you could use `grep -oE "\b([0-9]{1,3}\.){3}[0-9]{1,3}\b"` which will match IPv4 addresses.
Ref: https://unix.stackexchange.com/questions/296596/how-to-check-if-any-ip-address-is-present-in-a-file-using-shell-scripting

jstandring-gds: Opted for a full match of IPv4 addresses. It's much better to do a positive instead of inverted match.

PR #874, opened 2019-04-01
Make data.gov.uk Fastly backend domain configurable

thomasleese:
This allows us to set up a different staging configuration that points to the staging data.gov.uk PaaS instance.

[Trello Card](https://trello.com/c/MomTjKnV/910-ensure-only-whitelisted-ips-can-purge-dgu-cache)


elliotcm:
How and where do we need to make this configuration?

thomasleese:
The actual values for these variables are stored in govuk-aws-data, for example: https://github.com/alphagov/govuk-aws-data/pull/361

PR #877, opened 2019-04-04
Give search machines a 600s warmup period

barrucadu:
The healthcheck only checks nginx is running, not that the app is running.  We've seen this cause occasional problems elsewhere, so just set it to a large value for the search machines, given that we're planning to grow the ASG to 3 nodes.

### DIFF
--- a/terraform/modules/aws/node_group/main.tf
+++ b/terraform/modules/aws/node_group/main.tf
@@ -115,6 +115,12 @@ variable "instance_target_group_arns_length" {
   default     = 0
 }
 
+variable "asg_health_check_grace_period" {
+  type        = "string"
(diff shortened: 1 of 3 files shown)


terraform/modules/aws/node_group/main.tf
```diff
@@ -115,6 +115,12 @@ variable "instance_target_group_arns_length" {
   default     = 0
 }
 
+variable "asg_health_check_grace_period" {
+  type        = "string"
+  description = "The time to wait after creation before checking the status of the instance"
+  default     = "60"
```
fredericfran-gds: I don't know whether this should just be 60 without the quotes, well we will see in staging.

PR #895, opened 2019-04-24
Manage SOPS KMS key

afda16:
The SOPS KMS key has been created manually in the past. We want to add this resource
to `infra-security` to remove manual steps from an environment bootstrap.

We'll need to import the KMS resources into the existing environments


afda16:
I just updated the alias of the key, ready to merge

PR #896, opened 2019-04-24
Added training environment option

camdesgov:
Added training item to CONTEXTS array, gdscli aws
command can now be used with the new training environment,
ssh training option will need to be updated after we have set up
the jumpbox.

### DIFF
--- a/tools/govukcli
+++ b/tools/govukcli
@@ -9,7 +9,7 @@ set -e
 ### Global settings
 CONFIG_DIR=~/.govukcli
 CONTEXT_FILE=$CONFIG_DIR/current-context
-CONTEXTS='ci test tools integration staging production staging-aws production-aws'
+CONTEXTS='ci test tools integration staging production training staging-aws production-aws'
 
 ### Help output
 function usage {


tools/govukcli
```diff
@@ -9,7 +9,7 @@ set -e
 ### Global settings
 CONFIG_DIR=~/.govukcli
 CONTEXT_FILE=$CONFIG_DIR/current-context
-CONTEXTS='ci test tools integration staging production staging-aws production-aws'
+CONTEXTS='ci test tools integration staging production training staging-aws production-aws training-aws'
```
cbaines: From the week notes, I understood there being one training environment, similar to `integration`, rather than it being a new hybrid Carrenza/AWS environment. So you might just need the `training` context, rather than a `training` and `training-aws` context?

camdesgov: Ah you are quite right, thanks

afda16:
@cbaines , can you review again?

cbaines:
> @cbaines , can you review again?

Sure, the commit looks great now :+1:

PR #916, opened 2019-05-01
Update app-deploy for Training and non-blued support

afda16:
Update the app-deploy project to support non-blued infrastructure in
the new Training environment:
- we need to retrieve Route53 zones from variables instead of remote state files
- we need to bring the public ALB logic from infra-public-services to this module. This
needs to be optional. The default behaviour is to keep the original external ELB. If
we configure the variable `enable_public_alb` to true, the module does not create the
external ELB and creates a public ALB with same settings as originally implemented in
`infra-public-services`

We choose not to continue using the `lb` module because the code is complicated and
we are using LB resources in a simple way. We can investigate later if we can/need to
manage something in modules. For now we extracted the CloudWatch metrics to its own
module, same as other AWS resources.


afda16:
We are going to implement https://github.com/alphagov/govuk-aws/pull/921 instead of this

PR #918, opened 2019-05-02
add monitoring.* public dns name to monitoring classic load balancer

fredericfran-gds:
# Context

We need a monitoring.* public dns name for fastly to contact the monitoring classic load balancer. We can't use the existing alert.blue because the blue stack will be removed/merged later.

# Decisions
1. add the `monitoring.*` public dns name in the app-monitoring project

### DIFF
--- a/terraform/projects/app-monitoring/main.tf
+++ b/terraform/projects/app-monitoring/main.tf
@@ -67,6 +67,9 @@ data "aws_acm_certificate" "elb_internal_cert" {
   statuses = ["ISSUED"]
 }
 
+# This ELB is the entry point for the fastly rsyslog backup and hence,
+# the monitoring.* domain name should be directed to a suitable load-balancer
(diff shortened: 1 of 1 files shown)


terraform/projects/app-monitoring/main.tf
```diff
@@ -132,6 +132,9 @@ resource "aws_elb" "monitoring_external_elb" {
   tags = "${map("Name", "${var.stackname}-monitoring", "Project", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "monitoring")}"
 }
 
+# This ELB is the entry point for the fastly rsyslog backup and hence,
```
afda16: I think this comment is for `monitoring_external_elb`?

fredericfran-gds: yes, sorry. trying to do too many things at same time.

PR #924, opened 2019-05-03
Update app-email-alert-api-db-admin for Training environment

camdesgov:
Add backend to build app-email-alert-api-db-admin in the Training environment.

does not use the stack domain). app-email-alert-api-db-admin doesn't have
external ELB/DNS record.

### DIFF
--- a/terraform/projects/app-email-alert-api-db-admin/main.tf
+++ b/terraform/projects/app-email-alert-api-db-admin/main.tf
@@ -33,11 +33,26 @@ variable "remote_state_infra_database_backups_bucket_key_stack" {
   default     = ""
 }
 
+variable "internal_zone_name" {
+  type        = "string"
+  description = "The name of the Route53 zone that contains internal records"
+}
+
+variable "internal_domain_name" {
+  type        = "string"
+  description = "The domain name of the internal DNS records, it could be different from the zone name"
+}
+
 # Resources
 # --------------------------------------------------------------
 terraform {
   backend          "s3"             {}
   required_version = "= 0.11.7"
+}
+
+data "aws_route53_zone" "internal" {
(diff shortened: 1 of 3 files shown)


terraform/projects/app-email-alert-api-db-admin/main.tf
```diff
@@ -110,8 +125,8 @@ module "email-alert-api-db-admin" {
 }
 
 resource "aws_route53_record" "email_alert_api_db_admin_service_record" {
-  zone_id = "${data.terraform_remote_state.infra_stack_dns_zones.internal_zone_id}"
-  name    = "email-alert-api-db-admin.${data.terraform_remote_state.infra_stack_dns_zones.internal_domain_name}"
+  zone_id = "${data.aws_route53_zone.internal.zone_id}"
+  name    = "db-admin.${var.internal_domain_name}"
```
afda16: The name of the record still needs to be "email-alert-api-db-admin"

terraform/projects/app-email-alert-api-db-admin/main.tf
```diff
@@ -143,7 +158,7 @@ data "terraform_remote_state" "infra_database_backups_bucket" {
   config {
     bucket = "${var.remote_state_bucket}"
     key    = "${coalesce(var.remote_state_infra_database_backups_bucket_key_stack, var.stackname)}/infra-database-backups-bucket.tfstate"
-    region = "eu-west-1"
+    region = "eu-west-2"
```
afda16: We want to use var.aws_region so it works for all the environments

camdesgov:
We've discovered this app is no longer being used. It appears it was never migrated to AWS

PR #925, opened 2019-05-03
Update app-publishing-api-db-admin for Training environment

camdesgov:
Add backend to build app-publishing-api-db-admin in the Training environment.

does not use the stack domain), app-publishing-api-db-admin doesn't have
external ELB/DNS record.

### DIFF
--- a/terraform/projects/app-publishing-api-db-admin/main.tf
+++ b/terraform/projects/app-publishing-api-db-admin/main.tf
@@ -25,6 +25,16 @@ variable "remote_state_infra_database_backups_bucket_key_stack" {
   default     = ""
 }
 
+variable "internal_zone_name" {
+  type        = "string"
+  description = "The name of the Route53 zone that contains internal records"
+}
+
+variable "internal_domain_name" {
+  type        = "string"
+  description = "The domain name of the internal DNS records, it could be different from the zone name"
+}
+
 # Resources
 # --------------------------------------------------------------
 terraform {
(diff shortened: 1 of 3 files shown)


terraform/projects/app-publishing-api-db-admin/main.tf
```diff
@@ -92,8 +107,8 @@ module "publishing-api-db-admin" {
 }
 
 resource "aws_route53_record" "publishing-api_db_admin_service_record" {
-  zone_id = "${data.terraform_remote_state.infra_stack_dns_zones.internal_zone_id}"
-  name    = "publishing-api-db-admin.${data.terraform_remote_state.infra_stack_dns_zones.internal_domain_name}"
+  zone_id = "${data.aws_route53_zone.internal.zone_id}"
+  name    = "db-admin.${var.internal_domain_name}"
```
afda16: Keep the original record name "publishing-api-db-admin"

terraform/projects/app-publishing-api-db-admin/main.tf
```diff
@@ -125,7 +140,7 @@ data "terraform_remote_state" "infra_database_backups_bucket" {
   config {
     bucket = "${var.remote_state_bucket}"
     key    = "${coalesce(var.remote_state_infra_database_backups_bucket_key_stack, var.stackname)}/infra-database-backups-bucket.tfstate"
-    region = "eu-west-1"
+    region = "eu-west-2"
```
afda16: var.aws_region, so it works for all the environments

PR #929, opened 2019-05-03
Update app-warehouse-db-admin for Training environment

camdesgov:
Add backend to build app-warehouse-db-admin in the Training environment.

does not use the stack domain) app-warehouse-db-admin doesn't have external
ELB/DNS record.


cbaines:
Just noticed this Pull Request. The `app-warehouse-db-admin` project has somewhat been replaced by the https://github.com/alphagov/govuk-aws/tree/master/terraform/projects/app-content-data-api-db-admin project.

Obviously I've still got some work to do, as I need to tear down the infrastructure in Integration, and then remove the Terraform code...

camdesgov:
Closing pull request as this app has been replaced and is no longer needed in the Training environment. Thanks for the info @cbaines!

PR #930, opened 2019-05-07
Make Jumpbox external ELB optional for Training

afda16:
We want to make the creation of the external ELB in the `app-jumpbox`
project optional. The `infra-public-services` project didn't have
any alarms for this same Jumpbox ELB , so that needs to be added too.


camdesgov:
Confirm

PR #931, opened 2019-05-07
Update app-monitoring for Training environment

camdesgov:
Add backend to build app-monitoring in the Training environment.

does not use the stack domain). app-monitoring doesn't have external ELB/DNS record


afda16:
It needs the Route53 config for internal records from variable values

PR #935, opened 2019-05-10
Update app-content-data-api-db-admin for training environment

camdesgov:
Add terraform backend file for training environment
app-content-data-api-db-admin.

Update the database backups buckets to use the current aws region
automatically.

### DIFF
--- a/terraform/projects/app-content-data-api-db-admin/main.tf
+++ b/terraform/projects/app-content-data-api-db-admin/main.tf
@@ -77,7 +77,7 @@ data "terraform_remote_state" "infra_database_backups_bucket" {
   config {
     bucket = "${var.remote_state_bucket}"
     key    = "${coalesce(var.remote_state_infra_database_backups_bucket_key_stack, var.stackname)}/infra-database-backups-bucket.tfstate"
-    region = "eu-west-1"
+    region = "${var.aws_region}"
   }
 }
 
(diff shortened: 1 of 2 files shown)


terraform/projects/app-content-data-api-db-admin/main.tf
```diff
@@ -77,7 +77,7 @@ data "terraform_remote_state" "infra_database_backups_bucket" {
   config {
     bucket = "${var.remote_state_bucket}"
     key    = "${coalesce(var.remote_state_infra_database_backups_bucket_key_stack, var.stackname)}/infra-database-backups-bucket.tfstate"
```
schmie: Where is var.stackname defined in this case? Is it central for all of training, now?

camdesgov: Hi Sebastian, 

The stackname we've been using for training is  "govuk". We've been passing this as an argument when running  the "build-terraform-project.sh" (https://github.com/alphagov/govuk-aws/blob/master/tools/build-terraform-project.sh). For example:

   ./tools/build-terraform-project.sh -d ../govuk-aws-data/data -s govuk -e training -p app-mongo -c plan

jstandring-gds: Sounds good. Also confirmed with @schmie 👍

PR #937, opened 2019-05-10
Add support for ALB routing with specific application health checks - Option1

afda16:
We want to be able to define specific health checks per application


afda16:
Not needed, we'll use option2

PR #939, opened 2019-05-13
Add support for ALB routing with specific application health checks - Option2

afda16:
Update the original `lb_listener_rules` module, that hasn't been used so far,
to support a more specific case, where we want to add a target group and listener
rule for that target based on host headers. Each target group has its own health
check, so we can ensure traffic is only routing when the application is up and
healthy.

We are testing this module with the backend LB. We can choose when to use the
application healthchecks and target groups with the `enable_lb_app_healthchecks`
variable. By default the LBs continue using the default target group. When
we set `enable_lb_app_healthchecks` to true, we are creating the LB forward rules
from the service cnames variable.

This expression:
```
rules_host = ["${compact(split(",", var.enable_lb_app_healthchecks ? join(",", var.backend_public_service_cnames) : ""))}"]
```

is required because we can't use a conditional with list variables:
hashicorp/terraform#12453

This is fixed in Terraform 0.12, until then we need to implement a workaround.

### DIFF
--- a/terraform/modules/aws/lb_listener_rules/README.md
+++ b/terraform/modules/aws/lb_listener_rules/README.md
@@ -3,54 +3,32 @@
 This module creates Load Balancer listener rules and target groups for
 an existing listener resource.
 
-You can specify rules based on host header with the `rules_host` variable,
-path pattern with the `rules_path` variable, or both with the `rules_host_and_path`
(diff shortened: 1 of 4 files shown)


afda16:
Requires changes to default listener in the host

PR #944, opened 2019-05-14
Update the AWS Terraform provider version to 1.60.0

cbaines:
This is the latest minor version before major version 2.

From 1.40.0 in most cases, apart from app-prometheus and
infra-datagovuk-organogram-bucket which were on 1.14.0.

### DIFF
--- a/terraform/projects/infra-mirror-bucket/main.tf
+++ b/terraform/projects/infra-mirror-bucket/main.tf
@@ -57,13 +57,13 @@ terraform {
 
 provider "aws" {
   region  = "${var.aws_region}"
-  version = "1.40.0"
+  version = "1.60.0"
 }
 
 provider "aws" {
   region  = "${var.aws_replica_region}"
   alias   = "aws_replica"
-  version = "1.40.0"
+  version = "1.60.0"
 }
 
 data "aws_caller_identity" "current" {}
--- a/terraform/projects/app-apt/main.tf
+++ b/terraform/projects/app-apt/main.tf
@@ -74,7 +74,7 @@ terraform {
 
 provider "aws" {
   region  = "${var.aws_region}"
-  version = "1.40.0"
+  version = "1.60.0"
 }
 
 data "aws_route53_zone" "external" {
--- a/terraform/projects/app-asset-master/main.tf
+++ b/terraform/projects/app-asset-master/main.tf
@@ -34,7 +34,7 @@ terraform {
 
 provider "aws" {
   region  = "${var.aws_region}"
-  version = "1.40.0"
+  version = "1.60.0"
 }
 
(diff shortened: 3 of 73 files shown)


afda16:
https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1420-october-31-2018 -- I think vpc_id in private route53 zone resources, such as https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/infra-root-dns-zones/main.tf#L83, may need to be updated

afda16:
Apart from the aws_route53_zone changes, I think the rest of changes in the provider should be compatible

cbaines:
> https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1420-october-31-2018 -- I think vpc_id in private route53 zone resources, such as https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/infra-root-dns-zones/main.tf#L83, may need to be updated

Maybe I missed some, but this was actually the only example of `vpc_id` being used in a `aws_route53_zone` resource. I've now pushed a commit to use the `vpc { }` block though.

terraform/projects/infra-mirror-bucket/main.tf
```diff
@@ -57,13 +57,13 @@ terraform {
...
-  version = "1.40.0"
+  version = "1.60.0"
 }
 
 provider "aws" {
-  region  = "${var.aws_replica_region}"
-  alias   = "aws_replica"
-  version = "1.40.0"
+  region  = "${var.aws_backup_region}"
+  alias   = "aws_secondary"
```
afda16: If you update the alias, I think you need to update the line where it's referenced

terraform/projects/infra-mirror-bucket/main.tf
```diff
@@ -57,13 +57,13 @@ terraform {
...
   region  = "${var.aws_region}"
-  version = "1.40.0"
+  version = "1.60.0"
 }
 
 provider "aws" {
-  region  = "${var.aws_replica_region}"
-  alias   = "aws_replica"
-  version = "1.40.0"
+  region  = "${var.aws_backup_region}"
```
afda16: Does this variable exist in the file?

terraform/projects/infra-mirror-bucket/main.tf
cbaines: I believe this change was me just getting a rebase wrong, I've changed the commit to just change the version now. Good spot though!

terraform/projects/infra-mirror-bucket/main.tf
cbaines: Likewise, I believe this change was me just getting a rebase wrong, I've changed the commit to just change the version now.

PR #949, opened 2019-05-15
Bump CPU and disk of the mirrorer

schmie:
- We are testing to include all assets in the mirrorer in order to
provide a complete GOV.UK on a disk
- For testing our paradigm that this is working by simply extending the
scope of the crawler, we are temporarily upping the size of the mirrorer
significantly.
- Subsequently we should be able to significantly optimise the resource
requirements of the project

solo: @schmie


schmie:
Doing this via aws-data is more straightforward

PR #982, opened 2019-05-30
Elasticsearch6

bilbof:
Trello: https://trello.com/c/4bti7M9N/744

This duplicates the deployment code we have for provisioning the elasticsearch5 cluster for elasticsearch6.

It should work in the same way, except using elasticsearch version 6.7, the latest version available from AWS.

Once this is deployed, we will bump the number of master nodes to 3, and change this in the main.tf for this application.


barrucadu:
There will also need to be a PR to govuk-aws-data filling in some of the data: https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/1303/console

bilbof:
@barrucadu does this look OK? https://github.com/alphagov/govuk-aws-data/pull/451

bilbof:
@barrucadu I made a change to `terraform/projects/app-elasticsearch6/main.tf` to prefix the log names with `es6`, which I hope will fix the issue. Please could you re-review this?

terraform/projects/app-elasticsearch6/main.tf
```diff
@@ -0,0 +1,463 @@
...
+  }
+}
+
+data "aws_route53_zone" "internal" {
+  name         = "${var.internal_zone_name}"
+  private_zone = true
+}
+
+resource "aws_iam_service_linked_role" "role" {
+  aws_service_name = "es.amazonaws.com"
```
barrucadu: The name for this is clashing with the es5 one, so I think it'll need a `custom_suffix` (of `-es6`?) to fix it.

terraform/projects/app-elasticsearch6/main.tf
```diff
@@ -0,0 +1,463 @@
...
+}
+
+resource "aws_cloudwatch_log_group" "elasticsearch6_application_log_group" {
+  name              = "/aws/aes/domains/${var.stackname}-elasticsearch6-domain/es6-application-logs"
```
barrucadu: Specifying the ES version twice in the name is a bit unfortunate, but it's probably better to do it here than to do it in the `elasticsearch6_application_log_exporter.log_group_name` and have the names not match...

terraform/projects/app-elasticsearch6/main.tf
bilbof: Thanks, fixed

bilbof:
Thanks, will do.

terraform/projects/app-elasticsearch6/main.tf
```diff
@@ -0,0 +1,464 @@
...
+    }
+  }
+}
+
+data "aws_route53_zone" "internal" {
+  name         = "${var.internal_zone_name}"
+  private_zone = true
+}
+
+# resource "aws_iam_service_linked_role" "role" {
```
barrucadu: Remove the commented out bits and this is good to go

PR #983, opened 2019-05-30
Enlarge volume for app-draft-frontend instances

camdesgov:
Add new variable to main.tf to allow for a custom volume size.

One of the instances in the autoscaling group is running out of inodes
There doesn't seem to be anything significant we can delete, so we
are increasing the volume size.


afda16:
The default block size value in the node_group module is 20GB. Is the intention here to override that value in all the environments with 40GB?

PR #986, opened 2019-06-04
Update app-publishing-api for Training environment

camdesgov:
Add backend to build app-publishing-api in the Training environment.

does not use the stack domain).


afda16:
There isn't a publishing-api LB in the infra-public-services project

I think the external ELB here needs a create variable so we can test if we can stop using it, or if something requires publishing-api "blue" external  DNS records

PR #988, opened 2019-06-05
Update app-content-store for Training environment

camdesgov:
Add backend to build app-content-store in the Training environment.

does not use the stack domain).


afda16:
I think we want the external ELB to be created optionally, so we can test if we can remove it on all the environments, similar to commit 5f0bf677a394699e1363c71b0775a93a4f8d42e1

camdesgov:
Ah thanks for checking @afda16, I see what you mean.  I'll change this here and in the others, then run it by you again

PR #992, opened 2019-06-06
Add email-alert-api public ALB healthchecks

afda16:
Route traffic to email_alert_api_public based on Host header rules. Each rule uses
its own target port and health check path to ensure we only send traffic
to instances where the app is healthy.

email-alert-api instances also have an internal ELB, this change only covers traffic
from the public ALB.


sengi:
Obsoleted by recent work in preparation for migrating email-alert-api. We now have the app healthchecks configured on both LBs.

PR #1026, opened 2019-06-20
Fix rules target group names

afda16:
## lb_listener_rules improvements

Life cycle rules:
We are planning to change the target group name of existing resources and
this is going to recreate the target group and attachment managed by this
module, and refresh the listener rule. We want to ensure the old target
group is only destroyed when the new one has been created and the listener
rule has been updated.

Target Group names:
Previous target groups created by the module ended up with names not very
user friendly, for instance `draft-fron-draft-finder-frontend`. We are using
these names in the configuration of Icinga checks. This format, originated
from the trimming of the `name` variable and the rule name, is making people
confused as it looks like a typo. We need to trim the target group name in
first place because it's limited to 32 characters.

The change here renames the variable to `target_group_prefix` to make it clear
how/where it's used. The description of the variable is improved. We are also
going to enforce the limit of this variable to 4 characters. With the use of
a null_resource resource, the plan/apply of the code will fail if the user
configures the variable with something longer than 4 chars, for instance:

```
Error: module.internal_lb_rules.null_resource.check_target_group_prefix_length:
: invalid or unknown key: ERROR: The length of target_group_prefix is longer than 4 characters
```

Hopefully this will force the user to use a short prefix that uniquely identifies
the target groups. We are expecting to use it with the "zone" of the Loadbalancer,
for instance: `int`, `ext`, `pub`, etc. The Target group tags can be used to search
for TGs instead of expecting to do all the identification from the name.

## Update lb_listener_rules usage

Update how we use `lb_listener_rules`. We need to use `target_group_prefix` instead
of `name` and use a short prefix that helps creating unique TG names.

### DIFF
--- a/terraform/modules/aws/lb_listener_rules/README.md
+++ b/terraform/modules/aws/lb_listener_rules/README.md
@@ -3,7 +3,7 @@
 This module creates Load Balancer listener rules based on Host header and target groups for
 an existing listener resource.
 
-If the parameter `autoscaling_group_name` is non empty, the module also creates an attachment
+If the parameter `autoscaling_group_name` is not empty, the module also creates an attachment
 from each target group to the ASG with the specified name.
 
 Limitations:
@@ -21,7 +21,6 @@ so at the moment only one condition can be specified per rule
 | autoscaling_group_name | Name of ASG to associate with the target group. An empty value does not create any attachment to the LB target group. | string | `` | no |
 | default_tags | Additional resource tags | map | `<map>` | no |
 | listener_arn | ARN of the listener. | string | - | yes |
-| name | Prefix of the target group names. The final name is name-rulename. | string | - | yes |
 | priority_offset | first priority number assigned to the rules managed by the module. | string | `1` | no |
 | rules_host | A list with the values to create Host-header based listener rules and target groups. | list | `<list>` | no |
 | rules_host_domain | Host header domain to append to the hosts in rules_host. | string | `*` | no |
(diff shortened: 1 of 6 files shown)


fredericfran-gds:
can you give some examples of the name changes?

afda16:
> can you give some examples of the name changes?

name: "frontend-i-government-frontend" => "int-government-frontend" 
name: "frontend-i-info-frontend" => "int-info-frontend"
name: "frontend-i-manuals-frontend" => "int-manuals-frontend"
name: "frontend-i-service-manual-fronte" => "int-service-manual-frontend"

name: "calculator-calculators" => "int-calculators" 
name: "calculator-calendars" => "int-calendars"
name: "calculator-finder-frontend" => "int-finder-frontend" 
name: "calculator-licencefinder" => "int-licencefinder"
name: "calculator-smartanswers" => "int-smartanswers"

name: "draft-fron-draft-email-alert-fro" => "int-draft-email-alert-frontend" 
name: "draft-fron-draft-finder-frontend" => "int-draft-finder-frontend"
name: "draft-fron-draft-government-fron" => "int-draft-government-frontend"

name: "cache-www-origin" => "pub-www-origin"

name: "backend-content-performance-m" => "pub-content-performance-manager" 
name: "backend-content-publisher" => "pub-content-publisher" 
name: "backend-content-tagger" => "pub-content-tagger"
name: "backend-service-manual-publis" => "pub-service-manual-publisher"
name: "backend-short-url-manager" => "pub-short-url-manager"

issyl0:
This has conflicts and Ana isn't going to fix them 😅, plus it's >6 months old, so closing.

PR #1037, opened 2019-06-28
Mirror Cloudfront add cache origin and failover

afda16:
Add a new origin to the Mirror CloudFront distributions to use the cache
origin in first place and failover to the mirror bucket when 5xx errors
are detected.

Requires https://github.com/alphagov/govuk-aws/pull/1036, `origin_group` is only supported in provider > 2.3.0


afda16:
We can't implement this solution right now

PR #1042, opened 2019-07-03
Migrate mysql-xtrabackups from govuk-terraform-provisioning

afda16:
Migrated from:
https://github.com/alphagov/govuk-terraform-provisioning/tree/master/old-projects/mysql_xtrabackups


afda16:
I changed GLACIER policy to 90 days

PR #1045, opened 2019-07-09
create_before_destroy target_groups/listeners

chrisfarms:
changes to a launch configuration in a "node group" can cause a deadlock
if the node group (ASG) is used in target group for a load balancer ...
use create_before_destroy to ensure things are destroyed in the correct
order


chrisfarms:
closing as this doesn't have an attached story and we are not being impacted by the issue it prevents right now.

PR #1046, opened 2019-07-10
Add Documentation for new Training Environment

camdesgov:
Add Documentation for new Training Environment


schmie:
Unfortunately it was decided to deactivate training at this point in time out of cost considerations

PR #1047, opened 2019-07-11
Grant SSH for ITHC access ips

erino:


terraform/projects/infra-security-groups/backend.tf
```diff
@@ -120,3 +120,22 @@ resource "aws_security_group_rule" "backend-elb-external_egress_any_any" {
   cidr_blocks       = ["0.0.0.0/0"]
   security_group_id = "${aws_security_group.backend_elb_external.id}"
 }
+
+resource "aws_security_group" "backend_ithc_access" {
+  name        = "${var.stackname}_ithc_access"
```
rubenarakelyan: I think these need unique names for each security group.

erino: and also the `tags { Name = "` a few lines below?

PR #1059, opened 2019-07-26
Add Replica S3 Bucket for mysql-xtrabackups

camdesgov:
This bucket appears to contain essential database backups and
therefore should be replicated.


camdesgov:
This bucket is no longer going to be replicated. It is already being backed up to Glacier.

PR #1061, opened 2019-07-31
Add S3 Replication to custom-formats-mapit-storage bucket

camdesgov:
This change ensures that this bucket now has cross region replication.


terraform/policies/s3_govuk_custom_formats_mapit_storage_replication_role.tpl
```diff
@@ -0,0 +1,13 @@
+{
```
rtrinque: We already have identical policies, s3_backup_replica_role.tpl for example.
Our policies should be more generic and reusable rather than one project/one policy

PR #1063, opened 2019-08-01
Blacklisting load balancer healthchecks

chrisfarms:
we had healthchecks enabled for all target groups on the
cache/backend load balancers. however not all applications are
deployed in the aws side yet so this created a large amount of noise
in the form of 500 errors from failing healthchecks (60K a day).

A second issue with the use of terraform count and lists to generate
route53 resources makes it tricky to alter the list of
services/cnames without causing potential downtime.

So this commit introduces a blacklist that can be used to disable
the target group healthchecks. Ideally this logic would be removed
in the future once it is no longer needed.

:warning: depends on https://github.com/alphagov/govuk-aws-data/pull/538


chrisfarms:
Closing as we should complete the migration work to reduce this noise not add extra complexity as a workaround for noise

PR #1065, opened 2019-08-06
Add a variable to set the version of the lambda to use

rtrinque:
Terraform doesn't behave when trying to use $LATEST, there has to
be a better way to deal with this but in the meantime we have to
increment the version each time we publish a new lambda.


chrisfarms:
I _think_ what you want to do is keep the $LATEST but to add a `source_code_hash` to the `aws_lambda_resource` that is a hash of the zip ... this causes terraform to update the resource when it changes.

> source_code_hash - (Optional) Used to trigger updates. Must be set to a base64-encoded SHA256 hash of the package file specified with either filename or s3_key. The usual way to set this is filebase64sha256("file.zip") (Terraform 0.11.12 and later) or base64sha256(file("file.zip")) (Terraform 0.11.11 and earlier), where "file.zip" is the local filename of the lambda function source archive.

MuriloDalRi:
Hi, @rtrinque is this still needed or should we close this PR?

ChrisBAshton:
I'm going to assume there's no value in keeping this PR, as it's three years old, has conflicts, and has minimal changes.

Please reopen if you disagree.

PR #1069, opened 2019-08-09
Update Licensify-Frontend

camdesgov:
Add new security group and rules for internal load balancer.
Change internal elb loadbalancer to new alb loadbalancer.
Fix typo in autoscaling group tags.

### DIFF
--- a/terraform/projects/infra-security-groups/outputs.tf
+++ b/terraform/projects/infra-security-groups/outputs.tf
@@ -250,6 +250,10 @@ output "sg_licensify-frontend_external_elb_id" {
   value = "${aws_security_group.licensify-frontend_external_elb.id}"
 }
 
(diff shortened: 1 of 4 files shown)


terraform/projects/infra-security-groups/outputs.tf
```diff
@@ -250,6 +250,10 @@ output "sg_licensify-frontend_external_elb_id" {
   value = "${aws_security_group.licensify-frontend_external_elb.id}"
 }
 
+output "sg_licensify-frontend_internal_elb_id" {
+  value = "${aws_security_group.licensify-frontend_internal_elb.id}"
```
fredericfran-gds: should it still be elb when it has not changed to a alb?

camdesgov: Changes made, also replied on slack.

PR #1074, opened 2019-08-19
Set load balancer healthcheck paths for Licensify.

sengi:
These are the same paths as currently used by the Icinga healthchecks.

`licensify-frontend` serves both the frontend UI and the `uploadlicense`
service. There is therefore one target group and one healthcheck for
`licensify-frontend`. In other words, we are treating the app instance
as the failure domain for the purpose of LB healthchecking.

`licensify-feed` is not load-balanced (because it's a single-master,
manual-failover architecture).


sengi:
TF plan output in Integration:

`infra-public-services`: https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/1965/console
`licensify-frontend`: https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/1968/console
`licensify-backend`: https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/1969/console

PR #1086, opened 2019-08-29
Remove config for unused publishing_api_db_admin machine and RDS instance.

sengi:
This config was for an RDS instance which was never created.


sengi:
Related PRs:
https://github.com/alphagov/govuk-aws-data/pull/558
https://github.com/alphagov/govuk-puppet/pull/9543

sengi:
Bit more context: code for the RDS instance itself was deleted in https://github.com/alphagov/govuk-aws/pull/1083.

PR #1087, opened 2019-08-30
Make the terraforming of the necessary resources for a rabbitmq federation conditional depending on the environment

rtrinque:
We only want a rabbitmq federation in staging and production

### DIFF
--- a/terraform/projects/app-rabbitmq/main.tf
+++ b/terraform/projects/app-rabbitmq/main.tf
@@ -19,6 +19,12 @@ variable "aws_environment" {
   description = "AWS Environment"
 }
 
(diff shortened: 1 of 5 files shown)


terraform/projects/app-rabbitmq/main.tf
```diff
@@ -19,6 +19,12 @@ variable "aws_environment" {
   description = "AWS Environment"
 }
 
+variable "federation" {
+  type        = "string"
+  description = "Wether we do rabbitmq federation or not"
```
issyl0:
```suggestion
  description = "Whether we do RabbitMQ federation or not"
```

PR #1088, opened 2019-08-30
Use aws_iam_role_policy_attachment for related link roles

karlbaker02:
This commit places `aws_iam_policy_attachment`s with `aws_iam_role_policy_attachment`s, as the former only allows a policy to be associated with one role exclusively across projects.

From the Terraform docs:

_The aws_iam_policy_attachment resource creates exclusive attachments of IAM policies. Across the entire AWS account, all of the users/roles/groups to which a single policy is attached must be declared by a single aws_iam_policy_attachment resource._

See https://www.terraform.io/docs/providers/aws/r/iam_policy_attachment.html for more info.

Secondly, this commit adds the ARNs of policies `read_content_store_backups_bucket_policy` and `read_write_related_links_bucket_policy` as outputs, so that they can be reused by the upcoming knowledge-graph project.

### DIFF
--- a/terraform/projects/app-related-links/README.md
+++ b/terraform/projects/app-related-links/README.md
@@ -22,3 +22,10 @@ Run resource intensive scripts for data science purposes.
 | remote_state_infra_vpc_key_stack | Override infra_vpc remote state path | string | `` | no |
 | stackname | Stackname | string | - | yes |
 
+## Outputs
+
+| Name | Description |
+|------|-------------|
+| policy_read_content_store_backups_bucket_policy_arn | ARN of the policy used to read content store backups from the database backups bucket |
+| policy_read_write_related_links_bucket_policy_arn | ARN of the policy used to read/write data from/to the related links bucket |
+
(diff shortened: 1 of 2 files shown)


terraform/projects/app-related-links/README.md
```diff
@@ -22,3 +22,10 @@ Run resource intensive scripts for data science purposes.
 | remote_state_infra_vpc_key_stack | Override infra_vpc remote state path | string | `` | no |
 | stackname | Stackname | string | - | yes |
 
+## Outputs
+
+| Name | Description |
+|------|-------------|
+| policy_read_content_store_backups_bucket_policy_arn |  |
+| policy_read_write_related_links_bucket_policy_arn |  |
```
ChrisBAshton: Is there supposed to be a description here?

terraform/projects/app-related-links/main.tf
```diff
@@ -306,3 +306,10 @@ resource "aws_autoscaling_group" "related-links-ingestion" {
 # Outputs
 # --------------------------------------------------------------
 
+output "policy_read_content_store_backups_bucket_policy_arn" {
+  value = "${aws_iam_policy.read_content_store_backups_bucket_policy.arn}"
+}
```
sihugh: Some other outputs have descriptions here too.  Worth adding?

karlbaker02: Will do 👍

terraform/projects/app-related-links/README.md
karlbaker02: Thanks Chris, I'll add this 👍

PR #1090, opened 2019-08-30
Add HTTP->HTTPS redirect at the LB for licensify-admin.

sengi:
This already exists in UKCloud (on the licensify_lb nginx) and we need
to preserve the behaviour when migrating to AWS. The HTTPS redirect is
essential in rare cases, because HSTS preload cannot be relied upon
universally.

We add an ALB listener on port 80 with a default rule which serves
redirects to the same URL but with the https protocol. This way we don't
have to allow any HTTP traffic to be forwarded to the backend and we
keep the backend nginx config simple.

While we're there, also remove some superfluous entries from the ingress
ACL for the public LB. (It already allows port 443 from any source, so
it's not helpful to also list the office IP addresses in that rule. This was
copied from whitehall-admin I believe.)

### DIFF
--- a/terraform/projects/infra-security-groups/licensify-backend.tf
+++ b/terraform/projects/infra-security-groups/licensify-backend.tf
@@ -89,7 +89,18 @@ resource "aws_security_group_rule" "licensify-backend-external-elb_ingress_publi
   from_port         = 443
   protocol          = "tcp"
   security_group_id = "${aws_security_group.licensify-backend_external_elb.id}"
-  cidr_blocks       = ["0.0.0.0/0", "${var.office_ips}"]
+  cidr_blocks       = ["0.0.0.0/0"]
+}
+
+# Allow the public LB to serve HTTP->HTTPS redirects. See
+# licensify_backend_http_80 in ../infra-public-services/main.tf.
+resource "aws_security_group_rule" "licensify-backend-external-elb_ingress_public_http" {
+  type              = "ingress"
+  to_port           = 80
+  from_port         = 80
+  protocol          = "tcp"
+  security_group_id = "${aws_security_group.licensify-backend_external_elb.id}"
+  cidr_blocks       = ["0.0.0.0/0"]
 }
 
 resource "aws_security_group_rule" "licensify-backend-external-elb_egress_any_any" {
--- a/terraform/projects/infra-public-services/main.tf
+++ b/terraform/projects/infra-public-services/main.tf
@@ -1809,6 +1809,7 @@ resource "aws_lb_listener" "licensify_backend_http_80" {
       status_code = "HTTP_301"
     }
   }
+}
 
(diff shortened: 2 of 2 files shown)


terraform/projects/infra-security-groups/licensify-backend.tf
```diff
@@ -89,7 +89,18 @@ resource "aws_security_group_rule" "licensify-backend-external-elb_ingress_publi
   from_port         = 443
   protocol          = "tcp"
   security_group_id = "${aws_security_group.licensify-backend_external_elb.id}"
-  cidr_blocks       = ["0.0.0.0/0", "${var.office_ips}"]
+  cidr_blocks       = ["0.0.0.0/0"]
```
fredericfran-gds: My understanding of the govuk infrastructure is to lockdown integration and staging to us only (this does not preclude allowing other services to access as long as they restrict access upstream to us only. E.g. allow fastly to access staging and integration where appropriate because staging and integration fastly services have their own access control).

terraform/projects/infra-security-groups/licensify-backend.tf
```diff
@@ -89,7 +89,18 @@ resource "aws_security_group_rule" "licensify-backend-external-elb_ingress_publi
...
+
+# Allow the public LB to serve HTTP->HTTPS redirects. See
+# licensify_backend_http_80 in ../infra-public-services/main.tf.
+resource "aws_security_group_rule" "licensify-backend-external-elb_ingress_public_http" {
+  type              = "ingress"
+  to_port           = 80
+  from_port         = 80
+  protocol          = "tcp"
+  security_group_id = "${aws_security_group.licensify-backend_external_elb.id}"
+  cidr_blocks       = ["0.0.0.0/0"]
```
fredericfran-gds: see my comment above

terraform/projects/infra-security-groups/licensify-backend.tf
sengi: This doesn't appear to be the case currently. If I'm reading the config right, other backends (for example backend and whitehall-backend) can receive HTTP requests from anywhere.

This PR doesn't change the effect of the ACL; it only removes redundant entries.

fredericfran-gds: Yes I know but I think we should change that. It is true that these backend are usually protected by signon etc.

sengi: Ok, but I think it's a separate matter. I strongly think we should do it in a separate PR so as not to confuse matters.

fredericfran-gds: ok, cool

sengi:
> we should check the policy for backend access for non production environment and apply those in a separate PR.

Agreed! :)

PR #1091, opened 2019-08-30
Enable WAF for Licensing backend (licensify-admin).

sengi:
Enable the AWS web application firewall for licensify-admin.


sengi:
> looks good. Might be better to include a description in the PR next time for people not familiar with waf or for future reference.

Done :)

PR #1092, opened 2019-08-30
Increase number of calculator frontend boxes to 10

bevanloon:
To handle potential increased load against finder frontend, we are
increasing the number of finder frontend boxes.


terraform/projects/app-calculators-frontend/main.tf
```diff
@@ -33,7 +33,7 @@ variable "elb_internal_certname" {
 variable "asg_size" {
   type        = "string"
   description = "The autoscaling groups desired/max/min capacity"
-  default     = "5"
+  default     = "10"
```
cbaines: It's better to change the data in govuk-aws-data, rather than the default here.

bevanloon:
Closing in favour of https://github.com/alphagov/govuk-aws-data/pull/562

PR #1095, opened 2019-09-03
Add Knowledge Graph to integration

karlbaker02:
This PR adds a new project `app-knowledge-graph`, which contains the required resources to spin up an instance of the Neo4j Community AMI to iterate the GOV.UK knowledge graph. We will only be deploying this project to integration, which is why this is the only environment specified. Users will be connecting over HTTPS and BOLT to query the knowledge graph, with SSH being allowed only for Karl's key-pair if it's necessary to connect to the machine (we're not provisioning via Puppet and so don't have access to all SSH keys here).

This PR also adds two new security groups to `infra-security-groups` for the `app-knowledge-graph` project. These security groups are for the load balancer and for the instance itself, allowing access over SSH, HTTPS (via port 7473) and BOLT (a protocol based on TCP (https://boltprotocol.org), via port 7687).

Trello: https://trello.com/c/a9GlsqW7


terraform/projects/app-knowledge-graph/userdata.tpl
```diff
@@ -0,0 +1,134 @@
...
+aws s3 cp s3://${related_links_bucket_name}/roles_nodelist.csv /var/lib/neo4j/import/roles_nodelist.csv
+aws s3 cp s3://${related_links_bucket_name}/roles_to_orgs_edgelist.csv /var/lib/neo4j/import/roles_to_orgs_edgelist.csv
+
+# Get current Neo4j credentials and base64 encode
+credentials="neo4j:$instance_id"
+encoded_credentials=$(echo -n $credentials | base64)
+
+# Update Neo4j credentials with the password from SSM
+new_password=$(/usr/local/bin/aws ssm get-parameter --name govuk_knowledge_graph_shared_password --query "Parameter.Value" --region eu-west-1 --with-decryption | jq -r '.')
+curl -X POST -H "Accept: application/json" \
```
erino: What about setting the password much earlier? so there's a smaller window when the password isn't set?

karlbaker02: There's a password by default which is the instance id - always a password set 👍

erino: 👍 🔐

kevindew:
@karlbaker02 @erino  Could we have an [RFC](https://github.com/alphagov/govuk-rfcs) on the adding to this knowledge graph infrastructure to GOV.UK? We seem to be investing lots of time and resources into this work but, as I understand, most of the developer community are in the dark about what the needs behind it are. 

From my perspective I'm concerned that this is slowly heading towards us having a 3rd denormalisation of the Publishing API (after Content Store and Search API) in production which carries with it heavy spin-up and sync costs and that it introduces new and unfamiliar (to GOV.UK infrastructure) tooling to maintain. This is, of course, fine if this resolves a need, but that's another thing that's a bit unclear: what needs is this solving and can they be addressed by existing infrastructure? I feel an RFC could be beneficial to understanding why we're doing this, what we hope to get out of it, explore why alternatives are rejected and explain where this sits on a scale between data science experiment and production part of GOV.UK's infrastructure.

Also, if this is not a real part of our infrastructure (by it only being integration) should this instead be deployed to our test AWS environment?

karlbaker02:
@kevindew @erino 

Thanks very much for your comments Kevin, I can definitely see where you're coming from.

At the moment our team is looking at how the knowledge graph can help GOV.UK to better understand the data it has, including looking at relationships between data that we perhaps don't recognise and using these connections between data to start asking more detailed questions of our content.

Currently our data scientists each have their own version of the knowledge graph running on their local machines, which is problematic when it comes to iterating because we're iterating on data and the connections between it. The purpose of this particular PR is to provide the minimum necessary for our data scientists to be able to query an up-to-date version of the knowledge graph, removing the burden on them to maintain their own copy and making it easier to iterate on the graph.

The way that the graph is set-up at the moment, including the restoration of the content store from a backup of the content store as opposed to either querying it directly or using the Publishing API is necessary until we are ready to commit to productionising the knowledge graph. At the moment this work is in the discovery phase and should we move to productionise it, we will absolutely be looking at hooking it up to the Publishing API to ensure that it's data is kept up-to-date and we don't rely on the Content Store, which as you mentioned previously is a particular view of the Publishing API. It is also worth saying that the particular graph database technology we're using at the moment from the AMI that's included in this PR may not be what we roll with if we come to productionise, and that the model for hosting the technology we do use has the potential to be very different from what's in this PR.

In terms of the knowledge graph creating a third denormalisation of the Publishing API, I agree that is another place where we'll be storing content data and that we need to think more about how this might work. The knowledge graph will also contain data that the Publishing API does not and will likely never contain, such as additional information about content derived from various data science techniques and connections between data derived from the knowledge graph itself.

Regarding the creation of an RFC, I'm more than happy to write one for the knowledge graph, however now might be somewhat too early as the work is still very much in flux. Once we have more of a solid understanding of how the knowledge graph will be used and a better idea of the graph database technology we'll be using, we will create an RFC, as it would be absolutely great to get thoughts and ideas from everyone on the programme.

kevindew:
> @kevindew @erino
> 
> Thanks very much for your comments Kevin, I can definitely see where you're coming from.

Thanks for getting back to me so quickly Karl with your reasoned comments 👍 

> At the moment our team is looking at how the knowledge graph can help GOV.UK to better understand the data it has, including looking at relationships between data that we perhaps don't recognise and using these connections between data to start asking more detailed questions of our content.

This in particular seems to prompt an early question of why we need another data store to interrogate this information. The Publishing API stores the full graph of data and there hasn't been a justification why we a new mechanism to query it. I'm sure there are challenges in doing this giving it is based on postgres and not a dedicated graph database, but it is intriguing why we are creating new things to work out if there is value in data we already have available and is queryable.

As I recall there were numerous discussions over the years about whether the Publishing API should be powered by a graph database but this was rejected for the reliability and stability of Postgres. So it would be a strange twist if we ended up with both in separate places.

> Currently our data scientists each have their own version of the knowledge graph running on their local machines, which is problematic when it comes to iterating because we're iterating on data and the connections between it. The purpose of this particular PR is to provide the minimum necessary for our data scientists to be able to query an up-to-date version of the knowledge graph, removing the burden on them to maintain their own copy and making it easier to iterate on the graph.
> 
> The way that the graph is set-up at the moment, including the restoration of the content store from a backup of the content store as opposed to either querying it directly or using the Publishing API is necessary until we are ready to commit to productionising the knowledge graph. At the moment this work is in the discovery phase and should we move to productionise it, we will absolutely be looking at hooking it up to the Publishing API to ensure that it's data is kept up-to-date and we don't rely on the Content Store, which as you mentioned previously is a particular view of the Publishing API. It is also worth saying that the particular graph database technology we're using at the moment from the AMI that's included in this PR may not be what we roll with if we come to productionise, and that the model for hosting the technology we do use has the potential to be very different from what's in this PR.

I think what you've touched on here is one of the reasons I've been most keen to have wider discussion about this. It looks like we're considering "will a graph database add value? and if so, how do we productionise it?" this concerns me as it doesn't really consider do we need this technology to solve our problems? I don't anticipate it'll be hard to identify a use case a knowledge graph satisfies which may well mean we have to productionise it, but that may not be based on whether this was technology we need to solve the problem.

If we plan to just throw this all in the bin before any work on productionising it then that is fine, but that should be stated somewhere and be common knowledge. Otherwise we simply risk slipping this into the GOV.UK technology stack without any wider input.

> In terms of the knowledge graph creating a third denormalisation of the Publishing API, I agree that is another place where we'll be storing content data and that we need to think more about how this might work. The knowledge graph will also contain data that the Publishing API does not and will likely never contain, such as additional information about content derived from various data science techniques and connections between data derived from the knowledge graph itself.

The mix of denormalised and not data should be something to approach with caution. It reminds me of what was our highest priority piece of tech debt for a couple of years: [Unsure of our ability to recover the search-api index](https://trello.com/c/a1fll8qy/65-we-do-not-have-confidence-in-our-ability-to-recover-our-rummager-search-index-after-an-incident).

> Regarding the creation of an RFC, I'm more than happy to write one for the knowledge graph, however now might be somewhat too early as the work is still very much in flux. Once we have more of a solid understanding of how the knowledge graph will be used and a better idea of the graph database technology we'll be using, we will create an RFC, as it would be absolutely great to get thoughts and ideas from everyone on the programme.

Sure, It's cool for you to use your discretion. I think it'd be useful to have the stated intentions of what parts of infrastructure are meant to stick or be removed. There's also a whole lack of clarity as to if this is something we're setting up just for data scientists to do experiments or if we're hoping we can find something to plug it into on GOV.UK as soon as possible to prove value. If not an RFC it'd be useful to have that documented so it can be linked around to provide context.

Apologies to be a bit of a pain on this, it's just that adding a new database technology and/or adding a new denormalisation of Publishing API are majorly impactful things for GOV.UK and I worry we could easily sleepwalk our way into doing this or delay the key conversations until this is too far along to reasonably change any course.

PR #1096, opened 2019-09-05
Use search-api healthcheck, rather than nginx healthcheck

barrucadu:
This makes two changes, I'm not sure if both are necessary.

1. Change the ELB healthcheck to check for `/_healthcheck_search-api` over HTTP.  This means that requests will not be sent to an instance until the search-api healthcheck is passing.  This is fine, as search-api is the only app on the app-search machines.

2. Uses per-app healthchecks for external DNS - which is what Carrenza things (whitehall, search-admin) use to communicate with search-api.  I don't know if this is necessary with (1) in place.

I think we could use per-app healthchecks for internal DNS as well (rather than doing (1)), but I think that would require more restructuring, to set up a search-api ALB.

Plans:
- [app-search](https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/2217/console) (staging) - there's some instance changes which have been manually made in the console, I'll check with RE about the best way to deploy this (add those changes to govuk-aws-data, or just apply this change manually)
- [infra-public-services](https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/2216/console) (staging)

We should test both the AWS and Carrenza connection to search-api before deploying to production.  When these changes are both in staging, try publishing something from whitehall and then searching for it from finder-frontend.

---

[Trello card](https://trello.com/c/AT7HhBqP/987-prevent-search-api-accepting-requests-before-it-is-provisioned)

### DIFF
--- a/terraform/projects/infra-public-services/main.tf
+++ b/terraform/projects/infra-public-services/main.tf
@@ -2177,6 +2177,18 @@ resource "aws_route53_record" "search_api_public_service_names" {
   }
 }
 
+module "search_api_public_lb_rules" {
+  source                 = "../../modules/aws/lb_listener_rules"
+  name                   = "search-api"
+  autoscaling_group_name = "${data.aws_autoscaling_groups.search.names[0]}"
+  rules_host_domain      = "${var.aws_environment}.*"
+  vpc_id                 = "${data.terraform_remote_state.infra_vpc.vpc_id}"
+  listener_arn           = "${module.search_api_public_lb.load_balancer_ssl_listeners[0]}"
+  rules_host             = ["${compact(split(",", var.enable_lb_app_healthchecks ? join(",", var.search_api_public_service_names) : ""))}"]
+  priority_offset        = "1"
+  default_tags           = "${map("Project", var.stackname, "aws_migration", "search-api", "aws_environment", var.aws_environment)}"
+}
+
 resource "aws_autoscaling_attachment" "search_api_backend_asg_attachment_alb" {
   count                  = "${length(data.aws_autoscaling_groups.search.names) > 0 ? 1 : 0}"
   autoscaling_group_name = "${element(data.aws_autoscaling_groups.search.names, 0)}"
--- a/terraform/projects/app-search/main.tf
+++ b/terraform/projects/app-search/main.tf
@@ -118,7 +118,7 @@ resource "aws_elb" "search_elb" {
     unhealthy_threshold = 2
     timeout             = 3
 
-    target   = "TCP:80"
+    target   = "HTTP:80/_healthcheck_search-api"
     interval = 30
   }
 


terraform/projects/infra-public-services/main.tf
```diff
@@ -2177,6 +2177,18 @@ resource "aws_route53_record" "search_api_public_service_names" {
...
 }
 
+module "search_api_public_lb_rules" {
+  source                 = "../../modules/aws/lb_listener_rules"
+  name                   = "search-api"
+  autoscaling_group_name = "${data.aws_autoscaling_groups.search.names[0]}"
+  rules_host_domain      = "${var.aws_environment}.*"
+  vpc_id                 = "${data.terraform_remote_state.infra_vpc.vpc_id}"
+  listener_arn           = "${module.search_api_public_lb.load_balancer_ssl_listeners[0]}"
+  rules_host             = ["${compact(split(",", var.enable_lb_app_healthchecks ? join(",", var.search_api_public_service_names) : ""))}"]
```
sihugh: Should `var.search_api_public_service_names` have a `c` before `names` to make it `cnames`?

terraform/projects/infra-public-services/main.tf
```diff
@@ -2177,6 +2177,18 @@ resource "aws_route53_record" "search_api_public_service_names" {
   }
 }
 
+module "search_api_public_lb_rules" {
+  source                 = "../../modules/aws/lb_listener_rules"
+  name                   = "search-api"
+  autoscaling_group_name = "${data.aws_autoscaling_groups.search.names[0]}"
```
sihugh: This looks right to me, but I'm wondering how other implementations can use a direct property rather than the first item in an array (eg `data.aws_autoscaling_group.cache.name`)

terraform/projects/infra-public-services/main.tf
barrucadu: It probably should for consistency, but the code using `names` is already in place - changing it would require a corresponding change in govuk-aws-data

terraform/projects/infra-public-services/main.tf
barrucadu: infra-public-services isn't defining the ASG(s), so it has to find them somehow.  This is done in two ways:

Find an array, which we know is only going to have one element, eg:

```
data "aws_autoscaling_groups" "frontend" {
  filter {
    name   = "key"
    values = ["Name"]
  }

  filter {
    name   = "value"
    values = ["blue-frontend"]
  }
}
```

Find the one thing, eg:

```
data "aws_autoscaling_group" "backend" {
  name = "${var.app_stackname}-backend"
}
```

I'm not sure why there are two different ways of doing the same thing

PR #1097, opened 2019-09-05
Don't overwrite main default route on VMs with 2 interfaces.

sengi:
This should resolve the problem where instances with multiple interfaces
(for example mongo and rabbitmq instances) sometimes don't have Internet
connectivity.

For more context, see
https://aws.amazon.com/premiumsupport/knowledge-center/ec2-ubuntu-secondary-network-interface/

This fix comes from step 3 of the above guide.

### DIFF
--- a/terraform/userdata/10-associate-eni
+++ b/terraform/userdata/10-associate-eni
@@ -4,6 +4,9 @@
 # Snippet: associate-eni
 #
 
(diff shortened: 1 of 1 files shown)


sengi:
I plan to test this by applying TF for one of each type of affected host (mongo, rabbit and so on) in Integration, recycling an instance (avoiding primary ones) and checking that it comes up properly with eth0 as its default route on the main route table.

PR #1100, opened 2019-09-10
Change the deploy.rb script to use `govuk aws`

cbaines:
The `govuk aws` script provides some advantages over the `govukcli`
script currently being used. Using it doesn't require changing the
context set for govukcli, and it provides a JSON export which takes
less code to process the output.

There is a previous Pull Request [1], that stalled due to the `govuk aws` script not being official. I still don't know about how to officiate scripts, but I'm still using this approach, and I think it would be worth switching to using `govuk aws`, rather than the `govukcli`.

1: https://github.com/alphagov/govuk-aws/pull/831

### DIFF
--- a/tools/deploy.rb
+++ b/tools/deploy.rb
@@ -38,16 +38,11 @@ jenkins_job_uri = URI.parse("#{jenkins_url}#{jenkins_job_path}")
 
 # Get temporary AWS credentials
 puts 'Requesting temporary AWS credentials...'
-`govukcli set-context #{environment}`
-env = `govukcli aws invoke printenv`
+
+aws_credentials_string = `govuk aws --profile govuk-#{environment} --export-json`
 abort('Could not get temporary AWS credentials') unless $?.exitstatus.zero?
 
-# Set up the environment variables for the temporary AWS credentials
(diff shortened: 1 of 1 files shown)


AlanGabbianelli:
I think this looks good but I wonder if we should wait on the outcome of [`RFC-111: Standardise on one CLI tool`](https://github.com/alphagov/govuk-rfcs/pull/111) maybe? 🤔

cbaines:
> I think this looks good but I wonder if we should wait on the outcome of [`RFC-111: Standardise on one CLI tool`](https://github.com/alphagov/govuk-rfcs/pull/111) maybe? thinking

My view on that RFC is that it isn't relevant yet, as the `gds-cli` tooling isn't being used by developers on GOV.UK, and even if there was a decision made to use it, I don't believe it would be trivial to use in this script (due to the way different users need to assume different roles).

issyl0:
Using the GDS CLI would make this script _simpler_ as you'd be able to remove all of (`master`) lines 39-50 and instead do:

```sh
GITHUB_USERNAME=issyl0 GITHUB_TOKEN=asdf gds aws govuk-integration-admin -- ruby tools/deploy.rb plan integration govuk infra-security
```

issyl0:
Superseded by #1118.

PR #1101, opened 2019-09-10
Upsize draft-cache from 4 to 16 GB RAM in Staging and Prod.

sengi:
This is intended as a temporary workaround for `router` using
significantly more RAM when built with Go 1.12. It gives draft-cache the
same amount of RAM as (non-draft) cache.

We also switch from t2 to t3 because it's slightly cheaper and should be
otherwise equivalent or better.


sengi:
Upon closer inspection, the memory usage metrics for these boxes suggest that more RAM isn't needed and the problem is probably caused by `govuk_app_high_memory.sh` restarting `router` unnecessarily.

PR #1102, opened 2019-09-10
Remove config for unused DocDB cluster.

sengi:
- This cluster was originally created for Licensify with the intention
   of sharing it with other apps later on, hence its not having
   'licensify' in the name.
 - This turned out to be unworkable because DocDB does not currently
   support per-user access controls.
 - We therefore created a separate, dedicated cluster for Licensify
   (`app-licensify-documentdb`) and planned to repurpose
   `app-documentdb` as a shared cluster for other GOV.UK apps which
   don't store user data. Router was going to be the first app to use
   this.
 - The work to migrate Router to DocumentDB was reprioritised and will
   no longer happen in Q3. We therefore no longer need this cluster.
 - This cluster doesn't currently exist (in any environment).
 - We don't want to have un-applied config hanging around, so let's
   delete it for now. It's easy enough to get it back.

This essentially reverts https://github.com/alphagov/govuk-aws/pull/1064.


sengi:
> Maybe I'm mistaken but please don't delete yet. I thought this project is the basis for the shared documentdb for future applications like router. We can discuss tomorrow

I don't believe we have any work planned which is going to use this cluster. The Router->DocDB work was rolled back and I don't think we have any plans to resume it in Q4.

I agree that it might be useful at some point in the future, but that's what the repo is for :) The idea here is that we don't want TF to be out of sync with reality. @cbaines has been auditing the TF for unapplied changes, which is how this came up.

Unless I'm mistaken and we really are about to create this cluster?

PR #1111, opened 2019-09-12
Add tools/bulk-plan.rb

cbaines:
The `tools/deploy.rb` script allows more easily running Terraform for
a single project, but what if you want to potentially apply changes to
more than one project. Also, if you want to find out what unapplied
Terraform changes there are, how would you do that?

This script aims to meet those needs, it automates taking one or more
projects and running the `tools/deploy.rb` script for those
projects (taking in to account the stacks for the environment).

It's limited to running Terraform to plan, as I don't think it's
sensible to apply in bulk, that should only be done on a project by
project basis. Planning is safe to do in bulk, as it's just inspecting
the state of the infrastructure.

### DIFF
--- a/tools/bulk-plan.rb
+++ b/tools/bulk-plan.rb
@@ -1,0 +1,131 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+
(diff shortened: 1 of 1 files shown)


cbaines:
Thanks for taking a look Dean!

PR #1117, opened 2019-09-23
Support gds-cli and deploying from branch in deploy.rb.

sengi:
This lets RE GOV.UK team use deploy.rb to deploy Terraform instead of repeatedly copy/pasting tokens into Jenkins (as happens currently).

My intention is to keep deploy.rb working for existing users who are set up with govukcli, hence the heuristic to detect a configured gds-cli and otherwise fall back to govukcli as before.

Using gds-cli is necessary in order for RE GOV.UK to use this tool, because to use govukcli would be a retrograde step in terms of security since (all?) RE GOV.UK team members already use awsvault for managing their AWS admin creds.

Also support deploying from a branch. This is for temporary testing, as implemented in the Jenkins job. The idea is to reduce the temptation to run Terraform locally in such situations.

### DIFF
--- a/tools/deploy.rb
+++ b/tools/deploy.rb
@@ -6,7 +6,7 @@ require 'json'
 require 'uri'
 
 # Fields from the command line
-command, environment, stack, project, *rest= ARGV
+command, environment, stack, project, branch, data_branch, *rest= ARGV
 
 abort("too many arguments: #{rest}") unless rest.empty?
 
@@ -15,13 +15,17 @@ valid_commands = %w(plan apply plan-destroy destroy).freeze
 valid_environments = %w(integration staging production test tools training).freeze
 valid_stacks = %w(blue green govuk).freeze
 
-usage = 'Usage: GITHUB_USERNAME=... GITHUB_TOKEN=... ruby deploy.rb <command> <environment> <stack> <project>'
+usage = 'Usage: GITHUB_USERNAME=... GITHUB_TOKEN=... ruby deploy.rb <command> <environment> <stack> <project> [<branch>] [<data_branch>]'
 
 abort("GITHUB_USERNAME environment variable must be set\n#{usage}") unless ENV.has_key?('GITHUB_USERNAME')
 abort("GITHUB_TOKEN environment variable must be set\n#{usage}") unless ENV.has_key?('GITHUB_TOKEN')
 abort("command must be one of #{valid_commands.join(', ')}\n#{usage}") unless valid_commands.include?(command)
 abort("environment must be one of #{valid_environments.join(', ')}\n#{usage}") unless valid_environments.include?(environment)
 abort("stack must be one of #{valid_stacks.join(', ')}\n#{usage}") unless valid_stacks.include?(stack)
+
+# Use master branch if branch isn't specified.
+branch = 'master' if branch.nil?
+data_branch = 'master' if data_branch.nil?
 
 # Make sure the user is happy to go ahead
 puts "You're about to #{command} the #{stack}/#{project} project in #{environment}"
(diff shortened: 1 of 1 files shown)


issyl0:
A similar change has been considered in https://github.com/alphagov/govuk-aws/pull/1100. This probably supersedes that, as it a) improves security, but also b) keeps the existing `govukcli` functionality for those that use that.

cbaines:
Thanks for looking in to this @sengi . I can see how this would have worked in the past, but the identifiers for the environments within the `gds-cli` changed in https://github.com/alphagov/gds-cli/pull/165 to now be environments and roles (as previously, you could only assume the admin role in different environments). So this won't work if you update `gds-cli`.

This isn't a problem for the `govukcli`, as it relies on the profile in `~/.aws/config` which includes the ARN for the role.

I was thinking about this myself, and I think there are two approaches. Maybe try and also use information in `~/.aws/config`, or have the role for each environment be something that you can configure within the `gds-cli`, or have some other way of finding the right role in `deploy.rb` (maybe it could be an argument). Or, you could move out the AWS credentials concern from the `deploy.rb` script (something Issy mentioned here https://github.com/alphagov/govuk-aws/pull/1100#issuecomment-531136822 ). I think every approach has disadvantages, so I haven't really come up with any ideas about this yet.

issyl0:
You’re totally right, @cbaines. Delegating all of the Amazon work to the GDS CLI would solve this, but I’m loathe to start that debate again. I’d be in favour of that, though. How often do not-SREs run this? Could we have a flag for `—sre` that skips the govukcli code and assumes that credentials are already present?

The reason we have this problem with so many roles is because we started small and outgrew policy length limits on individual roles (eg `govuk-powerusers` had to split into `platformhealth-powerusers`, etc.).

I feel another tech-community survey brewing about other programmes’ approches to role names and policies. :grimacing: I know that Verify and Observe and some other things have roles that are `firstname.lastname-admin`, and it would be so much easier for things like this if GOV.UK could standardise on that too. Then we never hit role policy limits, because everyone is unique. And there’s just one role per person and also everyone’s an admin or read-only.

cbaines:
> You’re totally right, @cbaines. Delegating all of the Amazon work to the GDS CLI would solve this, but I’m loathe to start that debate again. I’d be in favour of that, though. How often do not-SREs run this? 

Occasionally, running Terraform isn't a very regular thing.

> Could we have a flag for `—sre` that skips the govukcli code and assumes that credentials are already present?

We could do, but just checking the environment would work as well. This works for example: https://github.com/alphagov/govuk-aws/pull/1118

> I feel another tech-community survey brewing about other programmes’ approches to role names and policies. I know that Verify and Observe and some other things have roles that are `firstname.lastname-admin`, and it would be so much easier for things like this if GOV.UK could standardise on that too. Then we never hit role policy limits, because everyone is unique. And there’s just one role per person and also everyone’s an admin or read-only.

If there's a better role approach that would be more consistent with the way other programs work on GDS, then I don't see any problem with looking in to that further.

issyl0:
Superseded by #1118.

cbaines:
> Superseded by #1118.

This Pull Request included support for deploying branches, which definitely isn't "superseded" by the Pull Request you refer to.

I'd also still be interested in hearing @sengi 's thoughts on this, rather than jumping straight away to considering it closed.

issyl0:
:hankey: Yes it did. Sorry, @sengi! I'll re-open this, but it needs a rebase.

sengi:
Thanks, @issyl0 and @cbaines, for landing a better version of this! :)

Shall I open a separate PR for the deploying-from-branch feature? Might be easier at this point.

cbaines:
> Thanks, @issyl0 and @cbaines, for landing a better version of this! :)

Well, better in terms of the changes I suggested worked with the newer versions of gds-cli, but your approach was better with the limitations of the older gds-cli versions.

> Shall I open a separate PR for the deploying-from-branch feature? Might be easier at this point.

Yeah, I think that would be good.

PR #1119, opened 2019-09-27
Fix and clean up load balancer configs for whitehall-admin.

sengi:
This replaces the Classic ELBs for `whitehall-admin` with ALBs and
uses the healthcheck which goes through to the application instead of
the generic nginx healthcheck which succeeds even when the app is not
running.

The healthcheck and ALB setup is described in
https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0037-alb-health-checks.md

We also remove some unused domain names which refer to
`whitehall_backend`. Nothing is configured to serve traffic for these
hostnames, so their existence only leads to confusion and surprise.


sengi:
Terraform plan output:

* [`infra-public-services`](https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/2570/consoleFull)
* [`app-whitehall-backend`](https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/2571/console)

PR #1120, opened 2019-09-27
Add ingress rule for Knowledge Graph SG to access PostgreSQL

karlbaker02:
This commit adds a new ingress rule for the `knowledge-graph` SG to access the PostgreSQL `primary` node (in integration) for the purposes of querying the Publishing API database directly. We run this query at present to generate a dataset of people and of roles, which we're experimenting with to improve the Knowledge Graph and the use cases we're testing off the back of it.

This change is temporary (for the lifetime of the Knowledge Graph), after which if we move forward with a knowledge graph then we will do so from the ground-up and not necessarily productionise what we already have in integration at present.

Trello: https://trello.com/c/jXF0o7bT

### DIFF
--- a/terraform/projects/infra-security-groups/postgresql.tf
+++ b/terraform/projects/infra-security-groups/postgresql.tf
@@ -82,3 +82,16 @@ resource "aws_security_group_rule" "postgresql-primary_ingress_db-admin_postgres
   # Which security group can use this rule
   source_security_group_id = "${aws_security_group.db-admin.id}"
 }
+
+resource "aws_security_group_rule" "postgresql-primary_ingress_knowledge-graph_postgres" {
+  type      = "ingress"
+  from_port = 5432
+  to_port   = 5432
+  protocol  = "tcp"
+
+  # Which security group is the rule assigned to
+  security_group_id = "${aws_security_group.postgresql-primary.id}"
+
+  # Which security group can use this rule
+  source_security_group_id = "${aws_security_group.knowledge-graph.id}"
+}


karlbaker02:
I think perhaps it'd be good for @erino and myself to chat to you about the database user and our considerations around that. 

In terms of data retrieval, some of the data we're querying currently can't be retrieved from the HTTP API. As was mentioned when we last caught up about this, at the moment I'd say it's probably too early to know if we'll actually want to add this to said API (until such time as we have some confidence that it's actually valuable), but if this is determined, then we would definitely want to add to the API rather than direct database querying.

kevindew:
> I think perhaps it'd be good for @erino and myself to chat to you about the database user and our considerations around that.
> 
> In terms of data retrieval, some of the data we're querying currently can't be retrieved from the HTTP API. As was mentioned when we last caught up about this, at the moment I'd say it's probably too early to know if we'll actually want to add this to said API (until such time as we have some confidence that it's actually valuable), but if this is determined, then we would definitely want to add to the API rather than direct database querying.

Sure, is there somewhere we're recording that? even if it turns out there isn't a use now it is arguably a hole in Pub API if it's difficult to access data it's the broker for. It doesn't surprise me too much as most of the existing API is written as RPCs for scenarios rather than a logical exposure mechanism for the data stored. If it's not already being recorded could be either a Github issue on Pub API or could use a relatively old Trello board we have of [Publishing API Wishlist](https://trello.com/b/VYLJBqJ9/publishing-api-wishlist).

PR #1121, opened 2019-10-01
Delete duplicate S3 buckets for CloudTrail.

sengi:
The permissions were already removed from these buckets in
https://github.com/alphagov/govuk-aws/pull/1070.


sengi:
The `force_destroy = true` change from #1070 has already been deployed in Integration, Staging and Prod.

PR #1130, opened 2019-10-07
Remove prefix from lifecycle_rule

MuriloDalRi:
This lifecycle rule will not apply with this prefix.

See https://github.com/alphagov/govuk-aws/pull/1099


issyl0:
Looks like #1099 does more than this PR (ie it applies to two places rather than this that only does one)?

MuriloDalRi:
> Looks like #1099 does more than this PR (ie it applies to two places rather than this that only does one)?

🤦‍♂ I should read the bucket names properly next time

PR #1138, opened 2019-10-11
Change publishing-api ELB healthchecks to hit the app, not the nginx.

sengi:
This switches the load balancer healthchecks for publishing-api from TCP
healthchecks (which only test the nginx proxy) to HTTP healthchecks
which are routed through to the app instance.

This is in line with ADR 37: https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0037-alb-health-checks.md

Co-authored-by: Camille Descartes <camille.descartes@digital.cabinet-office.gov.uk>


sengi:
Tested by applying manually in the AWS console in Staging and verifying that there's no diff in the Terraform plan.

PR #1151, opened 2019-10-30
update terraform aws plugin and set progresql autoscale

fredericfran-gds:
set rds postgresql to autoscale up to a maximum limit.
We need to update the aws provider version to use this
functionality.

data PR: https://github.com/alphagov/govuk-aws-data/pull/612


sengi:
Thanks for updating all the provider versions!

PR #1155, opened 2019-11-05
Support messages in Email Alert API archive

benthorner:
https://trello.com/c/z3REh76h/178-export-regular-data-on-email-subscriptions-for-insights-team


benthorner:
@kevindew I've just tried this out, and had to make a change to put the column on the end. I've amended the commit to explain.

benthorner:
@kevindew from the error I got, it seems so. The possibility of a delayed update did occur, but there's also this vague line about 'this does not apply to structs and arrays' in that doc I linked to, which lead me to doubt it would fix itself.

PR #1157, opened 2019-11-06
Set default ASG size to 3 for cache instances

issyl0:
- Staging and Production are reverting to pre-spike instance numbers.
  This was 3 in both environments. It makes the default 3, with
  integration the outlier with 2 instances which we set in
  govuk-aws-data.


issyl0:
We are unsure about decreasing the number of cache machines given recent incidents, so I'm closing this PR for now.

PR #1158, opened 2019-11-07
Add Variable For EBS Volume Size

camdesgov:
This allow the volume size to be customisable
between different environments.

### DIFF
--- a/terraform/projects/app-graphite/main.tf
+++ b/terraform/projects/app-graphite/main.tf
@@ -82,6 +82,12 @@ variable "instance_type" {
   default     = "m5.xlarge"
 }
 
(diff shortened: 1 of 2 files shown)


terraform/projects/app-graphite/main.tf
```diff
@@ -82,6 +82,12 @@ variable "instance_type" {
   default     = "m5.xlarge"
 }
 
+variable "ebs_volume_size" {
+  type        = "string"
+  description = "EBS Volume size"
```
fredericfran-gds: I think you should specify in the description that it is in GB

camdesgov: Good idea, changes made

PR #1166, opened 2019-11-12
Add Shared DocumentDB Cluster

fredericfran-gds:
This documentdb cluster is intended to be used by many
govuk apps that require mongodb.

For now, only asset-manager will use it.


terraform/projects/app-shared-documentdb/main.tf
```diff
@@ -0,0 +1,118 @@
...
+}
+
+variable "master_password" {
+  type        = "string"
+  description = "Password of master user on DocumentDB cluster"
+}
+
+variable "tls" {
+  type        = "string"
+  description = "Whether to enable or disable TLS for the Licensify DocumentDB cluster. Must be either 'enabled' or 'disabled'."
```
sengi:
```suggestion
  description = "Whether to enable or disable TLS for the DocumentDB cluster. Must be either 'enabled' or 'disabled'."
```

terraform/projects/infra-security/main.tf
```diff
@@ -269,6 +269,11 @@ resource "aws_kms_key" "licensify_documentdb" {
   policy      = "${data.aws_iam_policy_document.kms_sops_policy.json}"
 }
 
+resource "aws_kms_key" "shared_documentdb" {
+  description = "Senstive data in Shared DocumentDB"
```
sengi:
```suggestion
  description = "Encryption key for Shared DocumentDB"
```

fredericfran-gds: will fix and above too

terraform/projects/app-shared-documentdb/main.tf
fredericfran-gds: Will fix

PR #1169, opened 2019-11-13
Make Replication Settings Adjustable For Database Backups

camdesgov:
This change will allow us to have different settings for each
environment, which will help us to save money. Integration
Backups for example, do not need replication and therefore this setting should
be disabled there.


terraform/projects/infra-database-backups-bucket/main.tf
```diff
@@ -76,6 +76,12 @@ variable "expiration_time_whisper_mongo" {
   default     = "7"
 }
 
+variable "replication_setting" {
+  type        = "string"
+  description = "Whether or not replication is Enabled or Disabled"
```
sengi:
```suggestion
  description = "Whether replication is Enabled or Disabled"
```

PR #1179, opened 2019-11-20
Add support for journey visualisation tool

karlbaker02:
This PR adds support for the journey visualisation tool, hosted as part of the knowledge graph and using it to power searches of analytics between content pages. 

Specifically, this PR adds new rules to the knowledge graph SG and knowledge graph ELB SG, in addition to calling a new script within userdata to start the tool.

This PR also makes a small change to the ASG scheduled actions to ensure that we continue to generate base data at weekends.

### DIFF
--- a/terraform/projects/app-knowledge-graph/main.tf
+++ b/terraform/projects/app-knowledge-graph/main.tf
@@ -262,7 +262,7 @@ resource "aws_autoscaling_group" "knowledge-graph_asg" {
 resource "aws_autoscaling_schedule" "knowledge-graph_schedule-spin-up" {
   autoscaling_group_name = "${aws_autoscaling_group.knowledge-graph_asg.name}"
   scheduled_action_name  = "knowledge-graph_schedule-spin-up"
-  recurrence             = "0 9 * * MON-FRI"
+  recurrence             = "0 9 * * MON-SUN"
   min_size               = -1
   max_size               = -1
   desired_capacity       = 1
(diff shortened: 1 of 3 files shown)


terraform/projects/app-knowledge-graph/main.tf
```diff
@@ -276,6 +277,15 @@ resource "aws_autoscaling_schedule" "knowledge-graph_schedule-spin-down" {
   desired_capacity       = 0
 }
 
+resource "aws_autoscaling_schedule" "knowledge-graph_schedule-spin-down-weekend" {
+  autoscaling_group_name = "${aws_autoscaling_group.knowledge-graph_asg.name}"
+  scheduled_action_name  = "knowledge-graph_schedule-spin-down-weekend"
+  recurrence             = "55 9 * * SAT-SUN"
+  min_size               = -1
+  max_size               = -1
+  desired_capacity       = 0
```
barrucadu: I'm not too clear on what the desired capacity of an ASG is, and the amazon docs didn't really help.  Does this (+ the rule above) mean:

- At 9:00am every day, the ASG is scaled to 1 instance
- At 9:55am on Saturday and Sunday, the ASG is scaled to 0 instances

karlbaker02: Yes, that's correct. I'm not sure if this explains it any better: https://www.terraform.io/docs/providers/aws/r/autoscaling_schedule.html#desired_capacity/

The reason it's scaled up for 55 minutes at the weekend is so that we can continue to generate our base data, which can then be put together later on without needing to run queries over external sources, such as BigQuery, and incur cost again.

PR #1181, opened 2019-11-21
InfraFastlyLogs: Add a `cache_response` log column

issyl0:
- The pre-existing `cache_hit` column's description was incorrect, so
  fix it.
- We want to find out which paths are served from the cache and how.
  This includes more Fastly-specific variables as per
  https://docs.fastly.com/en/guides/useful-variables-to-log#cache-related-logging-variables.

https://trello.com/c/FZoxuRQ5/1536-3-log-to-athena-whether-a-request-was-served-from-cache

### DIFF
--- a/terraform/projects/infra-fastly-logs/main.tf
+++ b/terraform/projects/infra-fastly-logs/main.tf
@@ -224,7 +224,7 @@ resource "aws_glue_catalog_table" "govuk_www" {
       name = "ser_de_name"
 
       parameters {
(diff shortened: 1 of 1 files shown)


terraform/projects/infra-fastly-logs/main.tf
```diff
@@ -586,7 +598,7 @@ resource "aws_glue_catalog_table" "bouncer" {
       name = "ser_de_name"
 
       parameters {
-        paths                 = "client_ip,request_received,request_received_offset,method,url,status,request_time,time_to_generate_response,content_type,user_agent,data_centre,cache_hit"
+        paths                 = "client_ip,request_received,request_received_offset,method,url,status,request_time,time_to_generate_response,content_type,user_agent,data_centre,cache_hit,cache_response"
```
issyl0: Do we need this for the Bouncer logs? 🤔

PR #1183, opened 2019-11-27
Policy to Allow Users to Rotate IAM AccessKeys

erino:
This police is for attaching to a role to allow 2ndline users to rotate IAM Access Keys.


erino:
@issyl0 That's the plan.

PR #1193, opened 2019-12-06
Use combined TLS cert for licensify-admin, to work with non-SNI clients.

sengi:
Unfortunately licensify-admin has some clients which don't support SNI and can't upgrade yet. This allows them to continue working for now.

Using the combined cert is less messy that swapping the primary and secondary certs.


sengi:
Tested in Staging, ran the Qualys tool on it and it no longer shows the "This site works only in browsers with SNI support." warning.

PR #1200, opened 2019-12-16
Add a Concourse pipeline for the `terraform-docs` updatedness checks

issyl0:
- This installs `terraform-docs` as the latest version from Homebrew, via the `Homebrew/brew` docker image. This makes CI as consistent as possible with people's local docs generation.
- Previously, the docs check was run on Jenkins, and to update the version of the `terraform-docs` package involved the [very complicated Aptly dance](https://docs.publishing.service.gov.uk/manual/debian-packaging.html), so no-one ever did. This meant that there have been many commits like 5e7a7bf5937c4c5080b84fd03c50b1d4c5bab442 as the local version and CI version diverged on every release.

TODO:

- [x] Give govuk-ci user access to this repo and add a token to the Concourse.
- [x] Push the pipeline and test that it works.
- [x] Delete the old Jenkinsfile config (and the `terraform-docs` package from Puppet and all machines).


issyl0:
Rookie mistake on my part: Whatever Concourse uses under the hood isn't [allowed to run as root for Homebrew](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/brew.sh#L390). Either we add it, or we add a new user as part of this job that isn't `root`.

pipeline.yml
```diff
@@ -0,0 +1,63 @@
...
+          platform: linux
+          image_resource:
+            type: docker-image
+            source:
+              repository: homebrew/brew
+              tag: 2.2.1
+          inputs:
+            - name: govuk-aws-pr
+              path: repo
+          run:
```
tlwr:
```suggestion
          run:
            user: linuxbrew
```

issyl0: Thanks!

issyl0:
I made an upstream PR for the inclusion of Concourse in the allowed list: https://github.com/Homebrew/brew/pull/6849.

pipeline.yml
issyl0: Didn't work. I resorted to some `sed` to include `garden` in the allow list.

pipeline.yml
```diff
@@ -0,0 +1,71 @@
...
+            - |
+              # It's not a Concourse pipeline without some sed (until this gets into a Homebrew release)...
+              sed -i 's/azpl_job|docker|kubepods/azpl_job|docker|garden|kubepods/g' $(brew --repo)/Library/Homebrew/brew.sh
+              brew update
+              brew install terraform-docs
+
+              echo "Checking the updatedness of README files..."
+              ./tools/update-docs.sh
+
+              if [[ ! $(git diff --exit-code) ]]; then
```
issyl0: This doesn't currently work. I think I mis-adapted from the Jenkins config.

issyl0: Thanks, @barrucadu!

issyl0:
Next on the list are [the other small linting checks that Jenkins runs for us](https://github.com/alphagov/govuk-aws/blob/master/Jenkinsfile), but other people can do those, and in separate PRs.

pipeline.yml
```diff
@@ -0,0 +1,71 @@
+resource_types:
+- name: pull-request
+  type: docker-image
+  source:
+    repository: teliaoss/github-pr-resource
```
thomasleese: Should this be indented?

pipeline.yml
```diff
@@ -0,0 +1,71 @@
...
+            HOMEBREW_NO_AUTO_UPDATE: 1
+            HOMEBREW_DEVELOPER: 1
+          run:
+            path: bash
+            dir: repo
+            args:
+            - -c
+            - |
+              # It's not a Concourse pipeline without some sed (until this gets into a Homebrew release)...
+              sed -i 's/azpl_job|docker|kubepods/azpl_job|docker|garden|kubepods/g' $(brew --repo)/Library/Homebrew/brew.sh
```
thomasleese: 😬

thomasleese: We should make sure to change this when Homebrew gets released!

pipeline.yml
issyl0: Good spot. Done!

PR #1203, opened 2019-12-19
Bump the email-alert-api machines to m5.xlarge

cbaines:
From m5.large machines. Before the migration to AWS, the 3 machines in
Carrenza had 8 cores. The m5.large machines just have 2, so double the
number of cores to 4 as we're seeing high load on the machines when
lots of delivery receipts are being received from Notify.


sengi:
Looking at the history in Graphite, these machines don't look CPU constrained to me. They've never been more that 80% utilised and that was a short-lived peak. Highest recorded load average is 4 (not very high), and that was for about an hour. Are you sure they need more cores?

PR #1204, opened 2019-12-20
Fix the JSON syntax for AWS Glue

kevindew:
These parameters need to be wrapped in quotes in order to be valid JSON.

We'll need to apply this to the fastly configuration 

/cc @sengi @timblair


timblair:
Let's also ensure that the log format config in Fastly for all three of these actually matches what we're saying it should be. From @kevindew's investigation, `bouncer` and `govuk_assets` might not have had the change applied. `govuk_www` has been updated, and now has the correct quoting.

kevindew:
Just a heads up that to get things working again I've had to move all the broken JSON files to a different directory in the bucket.

We now have s3://govuk-production-fastly-logs/govuk_www-broken-nov-2019 and s3://govuk-staging-fastly-logs/govuk_www-broken-nov-2019 which contain all the invalid JSON (some valid JSON is in there too from before and after the changes of https://github.com/alphagov/govuk-aws/pull/1181). Re-running the crawler without these files fixes all the many excess tables and gets the regular govuk_www tables working again. In theory if the JSON files that were invalid are fixed they can just be dropped back into the govuk_www directory and will be re-classified when the crawler runs on it's 4 hour interval.

PR #1213, opened 2020-01-08
Remove redundant completions

chao-xian:
The auto-completions for govukcli aws is no longer needed as aws
features were removed.


chao-xian:
@issyl0 thanks - fixed

PR #1224, opened 2020-01-20
Fix internal load balancer healthcheck for whitehall-frontend.

sengi:
This changes the load balancer healthcheck for whitehall-frontend to
talk to the app itself rather than the nginx reverse proxy in front of
the app (which always returns healthy even if the app is down).

See https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0037-alb-health-checks.md


sengi:
Tested in Integration by deploying the branch, running Puppet agent on the `whitehall_frontend` nodes and then running Smokey.

sengi:
Tested in Integration; healthchecks passing.

PR #1226, opened 2020-01-21
Merge pull request #1226 from alphagov/clarify-what-terraform-docs-is

AlanGabbianelli:
When I first saw this message, I didn't know what `terraform-docs` was.
This should save someone else a few seconds of googling.

### DIFF
--- a/tools/update-docs.sh
+++ b/tools/update-docs.sh
@@ -3,7 +3,7 @@
 # Update docs across all projects and modules
 #
 if [[ ! $(which terraform-docs) ]]; then
-  echo "Must install terraform-docs"
+  echo "Must install terraform-docs: https://github.com/segmentio/terraform-docs"
   exit 1
 fi
 


tools/update-docs.sh
```diff
@@ -3,7 +3,7 @@
 # Update docs across all projects and modules
 #
 if [[ ! $(which terraform-docs) ]]; then
-  echo "Must install terraform-docs"
+  echo "Must install terraform-docs: https://github.com/segmentio/terraform-docs"
```
issyl0: Can we make this `brew install terraform-docs` and add a few lines of description above it? Or link to https://github.com/alphagov/govuk-aws/blob/master/doc/guides/styleguide.md? (Though that in itself is pretty old and needs some work.)

AlanGabbianelli: I thought of writing `brew install terraform-docs` in there but then I preferred to link to the repo instead because there's more info than simply the command to install it (which is still present).

AlanGabbianelli: @issyl0 I added a link to https://github.com/alphagov/govuk-aws/blob/master/doc/guides/styleguide.md too.

AlanGabbianelli: Ah, you just deleted that file! 😅

issyl0: Yeah sorry I had second thoughts and thought I'd bin it instead as it was wholly out of date.

AlanGabbianelli: Are you with just linking to the repo then?
I'd rather have that than only the command to install it.

PR #1228, opened 2020-01-24
Abort deploy script if an invalid project is given

AlanGabbianelli:
The `deploy` script accepts four fields: `stack`, `project`,
`environment` and `command`. For each of them, there are only specific
possible values.

For the `stack`, `environment` and `command` fields, these values were
specified in the script and there was logic to abort if different values
were given.

This introduces the same logic for the `project` field.

I also reordered the variables and the `abort` statements to reflect the
order in which the fields must be given in the command line.

### DIFF
--- a/tools/deploy.rb
+++ b/tools/deploy.rb
@@ -11,17 +11,19 @@ stack, project, environment, command, *rest= ARGV
 abort("too many arguments: #{rest}") unless rest.empty?
 
 # Valid values for each field
-valid_commands = %w(plan apply plan-destroy destroy).freeze
-valid_environments = %w(integration staging production test tools training).freeze
 valid_stacks = %w(blue green govuk).freeze
+valid_projects = Dir.chdir("#{Dir.home}/govuk/govuk-aws/terraform/projects") { Dir.glob('*').select { |f| File.directory? f } }.freeze
+valid_environments = %w(integration staging production test tools training).freeze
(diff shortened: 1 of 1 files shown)


tools/deploy.rb
```diff
@@ -11,17 +11,19 @@
 abort("too many arguments: #{rest}") unless rest.empty?
 
 # Valid values for each field
-valid_commands = %w(plan apply plan-destroy destroy).freeze
-valid_environments = %w(integration staging production test tools training).freeze
 valid_stacks = %w(blue green govuk).freeze
+valid_projects = Dir.chdir("#{Dir.home}/govuk/govuk-aws/terraform/projects") { Dir.glob('*').select { |f| File.directory? f } }.freeze
```
issyl0: You can use `Dir.entries` to simplify this.

AlanGabbianelli: I only want to return directories (without "." and "..") and for that, I use `File.directory?(f)` which must receive the absolute path to work. So it's either that or
```
Dir.entries("#{Dir.home}/govuk/govuk-aws/terraform/projects").select {|entry| File.directory? File.join("#{Dir.home}/govuk/govuk-aws/terraform/projects", entry) and !(entry =='.' || entry == '..') }.freeze
```
😱

Unless you were suggesting something else? 🤔

issyl0: Ohhh 'cuz you want directories! Sorry, I misunderstood.

PR #1229, opened 2020-01-24
Add life cycle rules to the mirror replica

rtrinque:
We have life cycle for the mirror bucket but not for its replica,
 it doesn't really make sense to keep all this data.

### DIFF
--- a/terraform/projects/infra-mirror-bucket/main.tf
+++ b/terraform/projects/infra-mirror-bucket/main.tf
@@ -86,6 +86,18 @@ variable "cloudfront_assets_certificate_domain" {
   default     = ""
 }
 
+variable "lifecycle_main" {
+  type        = "string"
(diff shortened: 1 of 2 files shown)


terraform/projects/infra-mirror-bucket/main.tf
```diff
@@ -214,6 +214,28 @@ resource "aws_s3_bucket" "govuk-mirror-replica" {
...
   }
+
+  lifecycle_rule {
+    id      = "main"
+    enabled = true
+
+    prefix = ""
+
+    noncurrent_version_expiration {
+      days = 5
```
sengi: Might be worth making these variables now that they're in two places, just so that they can't get out of sync if they're changed in future.

seb-szy: Additional variables implemeted

PR #1184, opened 2019-11-28
Enable the DocumentDB slow query profiler.

sengi:
* Enable slow query logs for the shared DocDB cluster in all environments.
* Enable slow query logs for the Licensify DocDB cluster in all environments.
* Make the profiler settings configurable via vars.

This can be rolled out now that https://github.com/terraform-providers/terraform-provider-aws/issues/10953 is fixed and we've upgraded to `terraform-provider-aws` 2.46.0.

PR #1232, opened 2020-01-28
Upgrade terraform-provider-aws from 2.33.0 to 2.46.0.

sengi:
This version allows configuring the slow query profiler on DocumentDB.

Generated with `find terraform/projects -name main.tf |xargs sed -i bak 's/version = "2.33.0"/version = "2.46.0"/'` (and `tools/update-docs.sh` to regenerate the docs).

Release notes: https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#2460-january-23-2020

Prerequisite for #1184.


sengi:
Tested by running `gds aws govuk-integration-admin -- env GITHUB_USERNAME=sengi GITHUB_TOKEN=... tools/bulk-plan.rb integration \*` and inspecting the plan output.

Found an occurrence that I'd missed; fixed in #1234.

PR #1233, opened 2020-01-28
Migrate shellcheck CI check from Jenkins to Concourse

issyl0:
- This is hopefully easier to follow than having to [go to another repo and understand 35 lines of code](https://github.com/alphagov/govuk-jenkinslib/blob/60a7e9c3b6be83eead36987d7a9c04eaa90d7b09/vars/govuk.groovy#L1083-L1118).
- Use some YAML anchoring to avoid specifying the `homebrew/brew` Docker `image_resource` twice.

- Also some performance improvements in the Concourse pipeline, and a version bump.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,15 +15,6 @@ node ("terraform") {
       govuk.bundleApp()
     }
 
-    stage("Shellcheck") {
-      govuk.shellcheck([
-        "tools/*.sh",
-        "tools/govukcli",
-        "terraform/userdata/*",
-        "jenkins.sh",
-      ])
-    }
-
     stage("ADR check") {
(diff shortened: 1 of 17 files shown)


issyl0:
This isn't pushed to Concourse yet and confirmed working, hence why it's draft. But I needed a PR to be able to do that. 🌱

issyl0:
Of course. The version of `shellcheck` in the Trusty apt repos and the version in Homebrew are very far apart, so this pipeline surfaces issues that Jenkins doesn't.

sengi:
LGTM. Separately, I think it's worth considering using a prebuilt container rather than installing shellcheck (and updating the homebrew database) on every test run.

PR #1245, opened 2020-02-05
Add ability to monitor fastly_logs Athena databases

kevindew:
Trello: https://trello.com/c/93jgKiUX/1329-set-up-a-smoke-test-that-athena-is-still-operating-on-new-data

We want to have the means to monitor whether our fastly_logs athena
databases continue to return results, this is in response to a bad
deploy a few months ago where we discovered that the database was no
longer queryable.

To monitor Athena we need a user that has access to Athena, to AWS Glue
to access table details, to have read access to the bucket the data is
in and to have write access to a bucket to store results. The intention
is that this users credentials will be stored within Jenkins and
periodically a job will run that makes a query to Athena. This job will
be deemed successful if results can be pulled from the database.

This PR adds a new IAM user for this monitoring and creates a bucket
where results can be stored. The bucket has a short retention of 7 days
as I don't anticipate we'll have much (if any) use for the data it returns once
the query is completed. It then applies a policy so that this user can
access all the necessary resources.

I've applied this terraform to the test environment and it worked successfully.

### DIFF
--- a/terraform/policies/fastly_logs_athena_monitoring_policy.tpl
+++ b/terraform/policies/fastly_logs_athena_monitoring_policy.tpl
@@ -1,0 +1,50 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+       "Effect":"Allow",
+       "Action":[
+          "athena:*"
+       ],
+       "Resource":[
+          "*"
+       ]
+    },
+    {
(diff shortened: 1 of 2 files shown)


sengi:
Oh, I think you just need to run `tools/update-docs.sh` to make the Concourse lint happy.

PR #1246, opened 2020-02-05
Customise EBS volume size between environments

sarahseewhy:
* We needed to increase the volume size in production as part of the Aptly migration.
* Aptly has more packages in Carrenza than in AWS and required a larger EBS volume to achieve feature parity.


kevindew:
@sarahseewhy I see you're also seeing a fail around docs. I'm just working on fixing this at the mo thought I'd let you know in case you are too. Feel free to ping me if you want to collaborate on it.

sarahseewhy:
@kevindew cheers! It was also my mistake –– I forgot to run the automatic generate terraform docs script. I'll fix it now.

PR #1258, opened 2020-02-17
Linting CI: Switch Concourse for GitHub Actions

issyl0:
- Now we can use GitHub Actions again (for non-sensitive things), this
  switches the linting CI checks use it. Benefits include:
  - Not having to be on the VPN to see the output.
  - Native macOS runners, so Homebrew works more easily rather than
    having to wait for Docker containers to pull, run, `brew update` and
    install a load of dependent packages every time.
  - GitHub Actions has secrets for the repo embedded in each run, so we
    don't need to add them to any secrets store.
  - Cloning the repo just happens.

I tested this over in https://github.com/issyl0/govuk-aws/runs/450608126?check_suite_focus=true and it takes under a minute.


whi-tw:
It's a weird choice to run these in an environment built with homebrew (rather than by using specific containers?!), when docker by design enables skipping the whole 'set up the container on every individual run' step, and run tests with specific versions of tooling. I wonder if as well as switching to actions, we should consider removing what smells a lot to me like an anti-pattern.

I wonder if doing something similar to PaaS's [docker-cloudfoundry-tools](https://github.com/alphagov/paas-docker-cloudfoundry-tools) approach and creating specific testing containers might be more appropriate.

For the case of this repo, images already exist:

* https://hub.docker.com/r/koalaman/shellcheck/
* https://hub.docker.com/r/cytopia/terraform-docs

issyl0:
The context behind the [original PR to add this](https://github.com/alphagov/govuk-aws/pull/1200) was:

> Previously, the docs check was run on Jenkins, and to update the version of the terraform-docs package involved the very complicated Aptly dance, so no-one ever did. This meant that there have been many commits like 5e7a7bf as the local version and CI version diverged on every release. This makes CI as consistent as possible with people's local docs generation.

I just wanted it to be better and stop the frustration of people having CI be terraform-docs version 0.3.0 and locally version 0.7.0 and having to hand-edit docs files, and I did that in the quickest way I knew how. And there were a lot of shellcheck changes when I upgraded it to a version more recent than the one in the Trusty apt repos. :scream:

Feel free to refactor this to be better in a different way, but I think 30 seconds is pretty good now that it's faster thanks to Actions and not having to `docker pull` the world inside Concourse.

issyl0:
Also: I don't have time (or inclination) to do the "switch to pre-built containers", and this is a nice stop-gap for speed and access improvements until someone does (if we ever get a firebreak again?)?

PR #1259, opened 2020-02-17
Install zip utilities on Knowledge Graph machine

karlbaker02:
This PR installs `zip` and `unzip` on the Knowledge Graph machine. These will be used to unzip a JAR file, amend its contents, then re-zip back to a JAR file.


terraform/projects/app-knowledge-graph/userdata.tpl
```diff
@@ -28,6 +28,10 @@ sudo add-apt-repository ppa:deadsnakes/ppa -y && sudo apt-get update
 sudo apt-get install -y python3.6
 sudo apt install -y python3-pip
 
+# Install zip and unzip utilities
+sudo apt install -y zip
+sudo apt install -y unzip
```
issyl0: Can we do this in a single line?

PR #1263, opened 2020-02-27
Enable storage autoscaling of RDS instances

smford:
This enables autoscaling of storage for compatible RDS instances.

This PR needs to be merged at the same time as: https://github.com/alphagov/govuk-aws-data/pull/695

Defining a default of 100GB for the upper limit for AWS to allow storage autoscaling growth, this needs to be set to instruct terraform to configure storage autoscaling.

- allocated_storage defines the initial provisioning size of the RDS instance
- max_allocated_storage defines the upper limit for AWS to allow autoscaling growth

Configure all existing RDS instances to support scalable storage

This requires max_allocated_storage to be defined to all instantiations of rds_instance

- define a max_allocated_storage for content-data-api-postgresql-primary_rds_instance
- apply terraform fmt to terraform/projects/app-content-data-api-postgresql/main.tf
- rename storage_size to allocated_storage in
  terraform/projects/app-mysql/main.tf to make it align with the rest of
  the terraform codebase
- apply sensible max_allocated_storage for transition-postgresql-primary_rds_instance
- apply terraform fmt to terraform/projects/app-transition-postgresql/main.tf
- apply sensible max_allocated_storage to transition-postgresql-standby_rds_instance

### DIFF
--- a/terraform/projects/app-transition-postgresql/main.tf
+++ b/terraform/projects/app-transition-postgresql/main.tf
@@ -102,22 +102,23 @@ resource "aws_db_parameter_group" "app_transition_pg" {
 module "transition-postgresql-primary_rds_instance" {
   source = "../../modules/aws/rds_instance"
 
-  name                 = "${var.stackname}-transition-postgresql-primary"
-  engine_name          = "postgres"
-  engine_version       = "9.6"
-  default_tags         = "${map("Project", var.stackname, "aws_stackname", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "transition_postgresql_primary")}"
-  subnet_ids           = "${data.terraform_remote_state.infra_networking.private_subnet_rds_ids}"
-  username             = "${var.username}"
-  password             = "${var.password}"
-  allocated_storage    = "120"
-  instance_class       = "${var.instance_type}"
-  instance_name        = "${var.stackname}-transition-postgresql-primary"
-  multi_az             = "${var.multi_az}"
-  security_group_ids   = ["${data.terraform_remote_state.infra_security_groups.sg_transition-postgresql-primary_id}"]
-  event_sns_topic_arn  = "${data.terraform_remote_state.infra_monitoring.sns_topic_rds_events_arn}"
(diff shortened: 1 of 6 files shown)


sengi:
Is this safe if an instance already has more than 100 GB? Perhaps 1 TB would be better default? 100GB is actually the minimum recommended size, if I remember correctly (for IOPS reasons).

smford:
> Is this safe if an instance already has more than 100 GB? Perhaps 1 TB would be better default? 100GB is actually the minimum recommended allocated size, if I remember correctly (for IOPS reasons).
> 

Hello, yup it is safe for instances that are larger than 100gb, there are overrides within govuk-aws-data for the apps (content-data-api-postgresql, app-mysql, app-transition-postgresql, app-postgresql) that are larger.

1TB is good, but I'd like to return to look at this in a subsequent PR.  This PR will focus on enabling autoscaling under the current setup, and the next PR will be focused on improving the performance and looking at the monitoring of the autoscaling.

> Arguably anything in Prod that's more than lightly used should have at least 1 TB allocated anyway, because otherwise [the I/O performance is inconsistent](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html#EBSVolumeTypes_gp2). Not suggesting we address that now, but it seems like a good reason to consider allowing autoscaling to at least 1 TB by default.

Good point, this will be captured in a subsequent PR.

terraform/projects/app-transition-postgresql/main.tf
```diff
@@ -137,6 +138,8 @@ module "transition-postgresql-standby_rds_instance" {
   instance_name              = "${var.stackname}-transition-postgresql-standby"
   security_group_ids         = ["${data.terraform_remote_state.infra_security_groups.sg_transition-postgresql-standby_id}"]
   create_replicate_source_db = "1"
+  allocated_storage          = "120"
+  max_allocated_storage      = "150"
```
sengi: Leaving aside the question of the default, I do think we should make the max several times bigger than the current, otherwise we only get the downsides of autoscaling without the benefit.

```suggestion
  max_allocated_storage      = "500"
```

PR #1269, opened 2020-03-11
Switch `backend` internal LB to ALB and remove classic ELBs.

sengi:
* Add an internal-facing Application Load Balancer (ALB/L7LB) for the `backend` apps, with per-app healthchecks configured.
* Switch `backend.*.govuk-internal` DNS names to point to the new internal ALB instead of the classic ELB.
* Remove the classic ELBs `blue-backend-external` and `blue-backend-internal` and their associated resources.

We need to use an ALB instead of a classic ELB in order to configure healthchecks correctly for each app, per https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0037-alb-health-checks.md. The same set of `backend` machines hosts many different apps whose instances need to be healthchecked independently in order to handle the failure of individual app instances correctly.

The motivation for addressing this healthcheck issue now is that we'll soon be migrating the publisher apps to AWS in production.

The `blue-backend-external` classic ELB is no longer used. Its request logs consist entirely of AWS Security Scanner and bogus traffic where the Host header doesn't match any of the services. `govuk_uptime_collector` used to use it, but that was fixed in alphagov/govuk-puppet#10234.

The aws/lb module for the replacement ALBs creates the associated resources like CloudWatch alarms for us, hence we no longer need to define those explicitly in `app-backend/main.tf`.

#### Testing

The `blue-backend-external` classic ELB has already been removed in Staging to verify that nothing depends on it. I also checked its request logs and I couldn't find any legitimate-looking traffic.

Terraform plans:

* [Integration](https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4064/console)
* [Staging](https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4065/console)
* [Prod](https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4066/console)

#### Tips for reviewing

This is probably easiest to review commit-by-commit. The overall diff comes out in a less than entirely helpful way because Git's diff heuristics munge some of the stanzas together.

### DIFF
--- a/terraform/projects/app-backend/README.md
+++ b/terraform/projects/app-backend/README.md
@@ -18,11 +18,7 @@ Backend node
 | asg\_size | The autoscaling groups desired/max/min capacity | `string` | `"2"` | no |
 | aws\_environment | AWS Environment | `string` | n/a | yes |
 | aws\_region | AWS region | `string` | `"eu-west-1"` | no |
-| create\_external\_elb | Create the external ELB | `bool` | `true` | no |
-| elb\_external\_certname | The ACM cert domain name to find the ARN of | `string` | n/a | yes |
 | elb\_internal\_certname | The ACM cert domain name to find the ARN of | `string` | n/a | yes |
-| external\_domain\_name | The domain name of the external DNS records, it could be different from the zone name | `string` | n/a | yes |
-| external\_zone\_name | The name of the Route53 zone that contains external records | `string` | n/a | yes |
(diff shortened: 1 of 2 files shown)


sengi:
Fixed a couple of copy-pastos in `module "internal_lb"`: use the private subnets and the internal security group.

PR #1272, opened 2020-03-13
Fix internal load balancer healthcheck for MapIt.

sengi:
This changes the load balancer healthcheck for MapIt to
talk to the app itself rather than the nginx reverse proxy in front of
the app (which always returns healthy even if the app is down).

See https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0037-alb-health-checks.md


sengi:
Tested in Staging, HCs still pass.

Rolled out to Integration and Prod by changing the config by hand and then running TF plan to verify, because there are unrelated hanging diffs in Staging and Prod and I couldn't tell whether those were safe to roll out.

PR #1275, opened 2020-03-16
Per-app healthchecks on `backend` internal LB.

sengi:
### What

* Define a new ALB `govuk-backend-internal` for internal traffic.
* Use a single target group per app. Where an app is supposed to be reachable both internally and externally, point both LBs at the app's target group rather than creating duplicate target groups, which we can't afford because of the 50 TG per ASG hard limit.
* Move the existing `govuk-backend-public` ALB definition into the `app-backend` project. The two ALBs need to be in the same place in order to use the same set of TGs without excessive complexity.
* Move the service DNS records along with the ALBs. These belong together because they're inherently tightly coupled to the ALBs.
* Prevent ingress from the Internet to internal-only services via the public/external load balancer.

### Why

Our `lb_listener_rules` module has become unsuitable for configuring the `backend` load balancers. This is because:

* We need to do [per-app healthchecks](https://github.com/alphagov/govuk-aws/blob/master/doc/architecture/decisions/0037-alb-health-checks.md) in order for our services to be sufficiently reliable.
* The `backend` autoscaling group hosts 31 services and needs both an external and an internal LB. Most, but not all, of the services are available both internally and externally.
* In order to do per-app healthchecks on the internal LB, we need to upgrade it from classic ELB to ALB (like the external one).
* AWS has a hard limit of 50 target groups on a given autoscaling group. This isn't one of the quotas that they can raise for us.
* We only need a single TG for each application service (in order to healthcheck each app independently), but `lb_listener_rules` prevents us from doing this because it encapsulates the creation of the listener rule, TG and ASG attachment. With `lb_listener_rules`, we end up trying to attach 62 or more TGs to a single ASG and failing.

Therefore, given that we need per-app healthchecks before migrating the publisher apps to AWS, we have to move away from `lb_listener_rules`.

The lb_listener_rules module also makes extensive use of Terraform lists and `count`, which causes severe problems when there are many services on the same load balancer. These include the inability to remove a service without causing downtime for other services. We have encountered situations in which removing a service from near the top of the list results in Terraform failing part-way through the rollout, breaking the live configuration and entering a state which cannot be recovered from by re-running `apply`.

### Design goals / nonfunctional requirements

* Ease of understanding (relative to what we had before).
* Ease of maintenance.
* Flexibility.
    * We need to be able to easily adapt config to meet unforeseen future needs, not just anticipated ones.
* Straightforward changes like adding/removing a service or changing settings for a particular service should be simple and obvious to make and safe to roll out.
* Changes to one service should not be able to affect other services.
* Keeping related (or closely coupled or interdependent) config together.
    * If two config items need to be read together in order to understand what's going on, they should be in the same place unless there's an overriding reason for them not to be (for example if one of the config items is shared between many other TF projects and therefore warrants being in a central place or its own project).
    * We want to avoid making the user jump between lots of different files and having to keep an unnecessarily large amount of state in their head when making changes.
* Avoiding multiple definitions of information which would cause problems if they were to go out of sync.

#### Non-goals

 * Minimising the number of lines of code.
 * Avoiding repetition just for the sake of avoiding repetition.
    * It's important to remember that this is configuration code, not application logic. Good config code contains almost no logic.
    * This isn't supposed to be a program which generates config. It's supposed to *be* the config.

### Tips for reviewing

This is probably not one to review line-by-line 😅 I'd recommend going commit-by-commit, focusing on the _Avoid using lb_listener_rules module for backend LBs_ commit and assessing the overall approach and structure against the design goals above. The `terraform plan` output is as important as the code changes, of course (see links in comments below).

I'm not too worried about detail mistakes getting missed in review, because we're likely to catch those in testing. My primary concern is that we converge on a structure and approach that's maintainable for the future as well as solving the immediate practical problem.

### Requisite PRs in other repos

* alphagov/govuk-aws-data/pull/703

### DIFF
--- a/terraform/projects/app-backend/README.md
+++ b/terraform/projects/app-backend/README.md
@@ -19,6 +19,7 @@ Backend node
 | aws\_environment | AWS Environment | `string` | n/a | yes |
 | aws\_region | AWS region | `string` | `"eu-west-1"` | no |
 | elb\_internal\_certname | The ACM cert domain name to find the ARN of | `string` | n/a | yes |
+| elb\_public\_certname | The ACM cert domain name to find the ARN of | `string` | n/a | yes |
 | instance\_ami\_filter\_name | Name to use to find AMI images | `string` | `""` | no |
 | instance\_type | Instance type used for EC2 resources | `string` | `"m5.2xlarge"` | no |
 | internal\_domain\_name | The domain name of the internal DNS records, it could be different from the zone name | `string` | n/a | yes |
@@ -26,6 +27,7 @@ Backend node
 | remote\_state\_bucket | S3 bucket we store our terraform state in | `string` | n/a | yes |
 | remote\_state\_infra\_monitoring\_key\_stack | Override stackname path to infra\_monitoring remote state | `string` | `""` | no |
 | remote\_state\_infra\_networking\_key\_stack | Override infra\_networking remote state path | `string` | `""` | no |
+| remote\_state\_infra\_public\_services\_key\_stack | Override stackname path to infra\_public\_services remote state | `string` | `""` | no |
 | remote\_state\_infra\_root\_dns\_zones\_key\_stack | Override stackname path to infra\_root\_dns\_zones remote state | `string` | `""` | no |
 | remote\_state\_infra\_security\_groups\_key\_stack | Override infra\_security\_groups stackname path to infra\_vpc remote state | `string` | `""` | no |
 | remote\_state\_infra\_stack\_dns\_zones\_key\_stack | Override stackname path to infra\_stack\_dns\_zones remote state | `string` | `""` | no |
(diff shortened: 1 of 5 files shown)


sengi:
Terraform plans:

* [`app-backend` in Staging](https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4133/console)
* [`infra-public-services` in Staging](https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4134/console)

sengi:
Here's my audit of which services are behind the existing internal vs external LBs and comparing Integration against Production: https://docs.google.com/spreadsheets/d/1QqTS3Y4xw8FMoLn4_2jBPqXuhbQWeY1Jq48TTVp-sdo/edit

rtrinque:
Love the design goals!

issyl0:
Sorry but you're going to need to rebase this on master after I changed CI. :innocent:

fredericfran-gds:
superseded by https://github.com/alphagov/govuk-aws/pull/1287

PR #1286, opened 2020-04-20
ci: Migrate more checks to GitHub Actions

issyl0:
- There were a bunch of lint-type checks running on Jenkins still, and
  this made me sad.
- This switches to running checks whenever branches are pushed, not just
  when PRs are opened.
- I'm not 100% sure of the utility of keeping the Jenkins "if master
  then generate a release tag" step around - does anyone know if the
  release tags are ever used for anything?


rtrinque:
Yeah the release tag is not used anywhere as far as I know

issyl0:
Thanks! I'll push the removal of the Jenkins config when I'm back from the shops!

PR #1310, opened 2020-06-23
Revert "Switch to c5.xlarge instances for the Email Alert API"

cbaines:
8GB of memory should be fine for the Email Alert API, but for the last
4 Fridays, some of the Sidekiq processes for the Email Alert API have
used lots of memory, beyond the limit set for the Icinga checks.

Until we've managed to address the excessive memory usage, and
reliability of the jobs (Sidekiq isn't exactly reliable), going back
to 16GB of memory might help to avoid issues with Sidekiq using all
the memory.

This reverts commit dae54679cf50585abddb6c4413b7b2a26b0492e5.

### DIFF
--- a/terraform/projects/app-email-alert-api/README.md
+++ b/terraform/projects/app-email-alert-api/README.md
@@ -26,7 +26,7 @@ email-alert-api node
 | aws\_region | AWS region | `string` | `"eu-west-1"` | no |
 | elb\_internal\_certname | The ACM cert domain name to find the ARN of | `string` | n/a | yes |
 | instance\_ami\_filter\_name | Name to use to find AMI images | `string` | `""` | no |
-| instance\_type | Instance type used for EC2 resources | `string` | `"c5.xlarge"` | no |
+| instance\_type | Instance type used for EC2 resources | `string` | `"m5.xlarge"` | no |
 | internal\_domain\_name | The domain name of the internal DNS records, it could be different from the zone name | `string` | n/a | yes |
 | internal\_zone\_name | The name of the Route53 zone that contains internal records | `string` | n/a | yes |
 | remote\_state\_bucket | S3 bucket we store our terraform state in | `string` | n/a | yes |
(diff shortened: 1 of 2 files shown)


benthorner:
@cbaines it looks like this change may be unnecessary, given the following historical data. The logs also indicate the increase in memory itself has little/no effect until [we changed the restart thresholds](https://github.com/alphagov/govuk-puppet/pull/10476).

Would it be advantageous or effortful to switch back?

Screenshot 2020-07-01 at 12 35 41

cbaines:
> @cbaines it looks like this change may be unnecessary, given the following historical data.

Well, the image you share sets out why this change was made. I assume the green lines are memory usage, and when it touches the red line on Friday the 12th and Friday the 19th, that's one of the machines running out of memory.

> The logs also indicate the increase in memory itself has little/no effect until [we changed the restart thresholds](https://github.com/alphagov/govuk-puppet/pull/10476).

The increase back to 16GB of memory was only meant to reduce a repeat of the issues on the 12th and 19th. I think we can say in hindsight that the change was unnecessary (because no machine used more than 8GB of memory after the 19th), but that's not quite the same as saying it was ineffective.

> Would it be advantageous or effortful to switch back?

The above graph shows that there hasn't been a similar issue with memory since the 19th, but I definitely don't know enough about the situation to say that this means that the services on these machines will be able to get by with a reduced amount of memory in the future. Someone more informed should make that decision if warranted.

benthorner:
> Well, the image you share sets out why this change was made.

So it does - apparently I visually filter out spikes 😬.

> Someone more informed should make that decision if warranted.

My take is that the difference in cost is pretty negligable, and we don't have any data to suggest we're suffering from slower CPUs, but we do have data to suggest having more memory is valuable. It's unclear _exactly_ how much more memory we need(ed), but the new worker thresholds at least have the potential to consume it.

Conclusion: let's stay as we are until we have motivation to change.

PR #1315, opened 2020-07-01
S3 bucket for email attachments from whitehall

benjamineskola:
As part of the migration to Notify, Whitehall needs to send emails as links rather than attachments. Following the example of specialist-publisher (alphagov/specialist-publisher#1576, alphagov/specialist-publisher#1625, #1254, #1256) this is being done by uploading them to s3 and then sending a link which will serve the file.

https://trello.com/c/5xKHcam8/2012-5-whitehall-replace-attachments-in-documentlistexport-so-that-notify-can-be-used

### DIFF
--- a/terraform/policies/whitehall_s3_writer_policy.tpl
+++ b/terraform/policies/whitehall_s3_writer_policy.tpl
@@ -1,0 +1,28 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:ListBucket",
+                "s3:ListAllMyBuckets",
+                "s3:GetBucketLocation"
+             ],
+            "Resource": [
+                "arn:aws:s3:::*"
+            ]
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:GetObject",
+                "s3:PutObject",
+                "s3:DeleteObject"
+            ],
+            "Resource": [
+                "arn:aws:s3:::${bucket}",
+                "arn:aws:s3:::${bucket}/*"
+            ]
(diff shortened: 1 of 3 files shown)


cbaines:
> As part of the migration to Notify, Whitehall needs to send emails as links rather than attachments. Following the example of specialist-publisher (specialist-publisher#1576, specialist-publisher#1625, #1254, #1256) this is being done by uploading them to s3 and then sending a link which will serve the file.
> 
> https://trello.com/c/5xKHcam8/2012-5-whitehall-replace-attachments-in-documentlistexport-so-that-notify-can-be-used

If this stuff is useful to say, then it would be useful to have in the commit message.

terraform/projects/infra-whitehall/main.tf
```diff
@@ -0,0 +1,80 @@
...
+  tags {
+    name            = "govuk-${var.aws_environment}-whitehall-csvs"
+    aws_environment = "${var.aws_environment}"
+  }
+
+  logging {
+    target_bucket = "${data.terraform_remote_state.infra_monitoring.aws_logging_bucket_id}"
+    target_prefix = "s3/govuk-${var.aws_environment}-whitehall-csvs/"
+  }
+}
```
cbaines, line 47: Did you consider putting this stuff in the app-whitehall-backend project [1]? That at least seems quite related in name.

1: https://github.com/alphagov/govuk-aws/tree/master/terraform/projects/app-whitehall-backend

benjamineskola: Ah, I missed that — I just copied and renamed files from a previous PR that had done the same thing for specialist-publisher.

PR #1323, opened 2020-07-15
Add the --full-index flag to bundle install in bootstrap script

rtrinque:
To provision puppetmaster machines in AWS we run a bootstrap script
that will go through a bundle install at some point.
Recently this part of the process has started to fail with this
error : Retrying dependency api due to error (2/3):
Bundler::HTTPError Net::HTTPUnprocessableEntity: Too many gems
(use --full-index instead)

We can't be sure when this process broke since we provision
puppetmaster very rarely, I'm not sure what the underlying cause
of the problem is either, using --full-index seems to fix the issue


issyl0:
We had this problem when we provisioned the test environment. [Here's our commit](https://github.com/alphagov/govuk-aws/commit/4441f07c1de193a2bcedd0443f8f2a886f39eac0). Maybe we should have merged our branch!

PR #1327, opened 2020-08-19
fix typo in volume tag

fredericfran-gds:
without this, ci_master instance will not be able to finds its volume and attach it


rtrinque (approved):
Ok

PR #1330, opened 2020-08-25
change deploy.rb to redirect deploy terraform to AWS integration

fredericfran-gds:


issyl0 (approved):
I'm OK with this change to fix the URL.

However, for developer experience and not having multiple tools, I wonder if it's about time to delete this, now we've documented `gds govuk terraform` for long enough? Can we do a poll of who uses it? Then we can clean up some docs too.

PR #1331, opened 2020-08-25
Add accessibility report runner to Tools

karlbaker02:
This PR sets up the `govuk-accessibility-reports` repo to run on an EC2 instance in integration. It does this by:

- Setting up a relevant security group for the EC2 instance in `infra-security-groups`
- Updating `infra-mirror-bucket` to add a bucket access policy for integration to list and read contents
- Creating `app-accessibility-reports`, which contains the requirements of the EC2 instance which generates the accessibility reports and the userdata to provision the machine.

It is the intention that this machine will be manually controlled via the auto-scaling group as and when it's required, which is the reason why nothing else exists in this PR to trigger this automatically / on a schedule.


karlbaker02:
Will fix-up commits when changes have been reviewed

terraform/projects/app-accessibility-reports/userdata.tpl
```diff
@@ -0,0 +1,70 @@
...
+echo "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDRnpXveFLpYZYLP7RtHpREJmY6lSYzUM38vcCAgDAA/x4PQce+40KI9JBNTtTsU7nLmPSZw/DpC5gekiNOujFnDgZDE/ugDdIxJeA4ohA388gEXAB8P9sk0SGw/CgtIbyjOXxvh4RyQfCgGMYhJdQcGn4eFp0U4if1PP+IGQViU2On6+U92qIgMlhz0BS/Nrs0Ci27hQmyYJAOBJ4nmL4FLsOEPzzznkd+zz5i7/zLd1haN/COyN4lzrZLQ+KaW9re5DaiaFIV3t7iFywbo/2xNsp1ZB3k8gVC+kF/1sQbLoep2hdm4Vs3L5wxgDxrAs0Aj1/DvyRNNwh6bGz9iLjIGb1uZh3EB6Pf780KEBwiOnOY/trrMtOW3O3vhvZD2RwQr1nITkr54dPGI8RhREAcwJ1qpnuK7a8LeZ+liGJrt2chYHFlag+kT1Bp1vsLNX5N0ejSl6VIO2y2piNBClwYlbMNZ4oJjvO3RpEn0g8wNLHJ61zbb0+5hR6u4yWqkzeZ2i+9vlvV0SJLk8J62I/nlLFhBgwyNMyc0rkwl7NGhoD2ZEFLNP8Y2EB0cY/CY7qviEnrUf4qLFtVdB12WxmetPWbBNwAcO1DglIqqduaswFP4GtT8522xEFNtUGbfdF+odjBlGmBeO4mb/JZNvRkYt25LdH01Qr19t9fuuekYw== oscar.wyatt@digital.cabinet-office.gov.uk" >> /home/ubuntu/.ssh/authorized_keys
```
issyl0, line 10: IIRC there's a size limit for user-data, so too many more of these and we might have a problem. But it's alright for now. Might be worth a comment to say who's key is the Yubikey one - just for when we come to do leavers tickets?

issyl0 (approved):
Overall, looks like Terraform to me and you've tested it and it works. A lot nicer to review than the first iteration - I'm glad I waited until the end! Tiny nit about commenting the SSH keys, but apart from that, looks good!

terraform/projects/app-accessibility-reports/userdata.tpl
karlbaker02: Thanks Issy. I remember you saying this previously, and my intention is to actually use the `90-data-science-base` file which I included as a var to the userdata script but haven't actually used it yet. Ideally we don't have this in different places but in only a single place to make it easier for joiners/leavers as you point out.

karlbaker02: Will address this in a separate PR, as the problem exists in a few other projects also.

richardTowers (approved):
Looks good to me!

PR #1333, opened 2020-09-01
Upgrade python to version 3.7

oscarwyatt:
The knowledge graph requires python 3.7 for some key packages


karlbaker02 (commented):
Looks good. Would suggest we test this out first before merging - will help out doing this.

karlbaker02:
Actually - one thing I'd like to check. We don't need to upgrade the Python on the data-science-data machine, only the knowledge-graph machine? (data-science-data is the machine which runs the data pipeline)

karlbaker02 (approved):
Looks good 👍  I have tested this via a branch and both data generation and knowledge graph work as expected.

PR #1341, opened 2020-09-10
increase ci agents ebs volume to 130 gb to match carrenza

fredericfran-gds:


issyl0 (approved):
13 was rather small...

PR #1351, opened 2020-10-12
Add 6 more Mapit instances

thomasleese:
Unfortunately the Terraform config is currently hard coded for 2 instances, so to up it to 8, I've had to copy and paste the resources manually.

It's definitely worth refactoring this in the future, but for now we need to get the instances up relatively quickly, so I think this will do.

### DIFF
--- a/terraform/projects/app-mapit/main.tf
+++ b/terraform/projects/app-mapit/main.tf
@@ -221,6 +221,222 @@ resource "aws_ebs_volume" "mapit-2" {
   }
 }
 
+module "mapit-3" {
+  lc_create_ebs_volume          = "${var.lc_create_ebs_volume}"
+  source                        = "../../modules/aws/node_group"
+  name                          = "${var.stackname}-mapit-3"
+  default_tags                  = "${map("Project", var.stackname, "aws_stackname", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "mapit", "aws_hostname", "mapit-3")}"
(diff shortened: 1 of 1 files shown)


terraform/projects/app-mapit/main.tf
```diff
@@ -221,6 +221,222 @@ resource "aws_ebs_volume" "mapit-2" {
...
+    aws_environment = "${var.aws_environment}"
+  }
+}
+
+module "mapit-8" {
+  lc_create_ebs_volume          = "${var.lc_create_ebs_volume}"
+  source                        = "../../modules/aws/node_group"
+  name                          = "${var.stackname}-mapit-8"
+  default_tags                  = "${map("Project", var.stackname, "aws_stackname", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "mapit", "aws_hostname", "mapit-8")}"
+  instance_subnet_ids           = "${matchkeys(values(data.terraform_remote_state.infra_networking.private_subnet_names_ids_map), keys(data.terraform_remote_state.infra_networking.private_subnet_names_ids_map), list(var.mapit_6_subnet))}"
```
bevanloon, line 409:
```suggestion
  instance_subnet_ids           = "${matchkeys(values(data.terraform_remote_state.infra_networking.private_subnet_names_ids_map), keys(data.terraform_remote_state.infra_networking.private_subnet_names_ids_map), list(var.mapit_8_subnet))}"
```

bevanloon (changes requested):

terraform/projects/app-mapit/main.tf
thomasleese: I knew there would be one!

bevanloon (approved):

PR #1356, opened 2020-10-12
Block XSS and SQLi on AWS WAF

thebeanogamer:
This is actioning a `FIXME` for July 2019 which was added in #1055.

At the time it was noted that the this should be left on `ALLOW` for a while so the potential impact can be evaluated, it's not clear if this has occurred, but it is well past the original `FIXME` date, so if no one has had a chance to look at this, it may be worth extending the date listed.

Unrelated, the existence of this repository is awesome!

### DIFF
--- a/terraform/projects/infra-public-services/waf.tf
+++ b/terraform/projects/infra-public-services/waf.tf
@@ -17,7 +17,7 @@ resource "aws_wafregional_web_acl" "default" {
 
   rule {
     action {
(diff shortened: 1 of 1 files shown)


bilbof:
Hi @thebeanogamer, thank you for this! We've decided these rules would not be suitable for production so have removed them in #1379.

thebeanogamer:
:wave: @bilbof No problem! Happy to help, even if it is just drawing attention to things that should be removed.

PR #1358, opened 2020-10-14
Change search instance type to c5.xlarge

thomasleese:
At the moment the search machines are `c5.large` which has 4 GB of memory. We seem to be pretty close to that limit at the moment as most machines are using around 3.5 GB of memory. We've also noticed that Ruby 2.7 seems to use slightly more memory than earlier versions so I think it makes sense to increase the memory on these machines in advance of that. The `c5.xlarge` instance type has 8 GB of memory available.


fredericfran-gds (approved):
looks good

PR #1370, opened 2020-11-17
Set Gatling ASG to 1

edwardkerry:
This PR sets the Gatling ASG to 1 to allow loadtesting.


fredericfran-gds (approved):
ok, next time, might be better to do it via govuk-aws-data as it will not require docs regeneration.

sihugh (approved):

PR #1373, opened 2020-11-26
Edit app-cache ags_size

AlanGabbianelli:
This updates the default AWS auto scaling group size for `cache` from 6
to 9. The only environment using 6 was staging, and this will align the
default to what production uses instead.

Follow up PR: https://github.com/alphagov/govuk-aws-data/pull/804

Co-authored-by: Alan Gabbianelli <Alan.Gabbianelli@digital.cabinet-office.gov.uk>
Co-authored-by: Rebecca Pearce <Rebecca.Pearce@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/projects/app-cache/README.md
+++ b/terraform/projects/app-cache/README.md
@@ -22,7 +22,7 @@ Cache application servers
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | app\_service\_records | List of application service names that get traffic via this loadbalancer | `list` | `[]` | no |
-| asg\_size | The autoscaling groups desired/max/min capacity | `string` | `"6"` | no |
+| asg\_size | The autoscaling groups desired/max/min capacity | `string` | `"9"` | no |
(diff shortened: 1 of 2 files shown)


barrucadu (commented):
This won't actually affect anything, as we specify the values of all the variables in govuk-aws-data, eg: https://github.com/alphagov/govuk-aws-data/blob/master/data/app-cache/production/common.tfvars#L7

And didn't we scale up to more than 9 cache machines?

bilbof (approved):

AlanGabbianelli:
> This won't actually affect anything, as we specify the values of all the variables in govuk-aws-data, eg: [alphagov/govuk-aws-data@`master`/data/app-cache/production/common.tfvars#L7](https://github.com/alphagov/govuk-aws-data/blob/master/data/app-cache/production/common.tfvars#L7)

Yes @barrucadu this one goes together with https://github.com/alphagov/govuk-aws-data/pull/804
I'll update the description.

PR #1375, opened 2020-11-30
Remove unused S3 bucket for email archiving

benthorner:

Depends on: https://github.com/alphagov/email-alert-api/pull/1486

This is unused since [1]. Note that we need to force destroy the
bucket since it still has data in it [2]. We will remove all the
remaining terraform in a later commit.

[1]: https://github.com/alphagov/email-alert-api/pull/1486
[2]: http://salewski.email/2017/04/30/terraform-howto-delete-a-non-empty-aws-s3-bucket.html


benthorner:
Need to test this on Integration before merging.

kevindew (approved):
Nice one

PR #1376, opened 2020-11-30
Empty bucket manually instead of force destroying

benthorner:

Previously we tried using the "force_destroy" flag, but this doesn't
seem to have the intended effect (nothing changes), and actually just
leaves Terraform thinking it's already deleted the bucket. Turns out
we can empty the bucket manually, so this just removes it.

Note that we will delete the project entirely in a future commit.


smford (approved):
lgtm

PR #1377, opened 2020-12-01
Remove email archive project

benthorner:

This was originally used to provision an S3 bucket for long term
email storage, which we did not need in practice.


kevindew (approved):
👋

PR #1378, opened 2020-12-02
Allow Jenkins to send https to Bouncer internal ELB

fredericfran-gds:
This is required so that Jenkins can run smokey tests against Bouncer
for continuous deployment reasons. Jenkins will access Bouncer via
the internal ELB of Bouncer on HTTPS (port 443).


ChrisBAshton (approved):
It worked! https://deploy.blue.staging.govuk.digital/job/Smokey/7924/console

Thanks

PR #1379, opened 2020-12-02
Remove unused SQLI and XSS rules

bilbof:
These rules were unsuitable for production.

We plan to migrate to WAF v2 from WAF Classic. We could also remove the X-Always-Block rule too, but we may want it when we migrate.


sengi (approved):
Thanks for cleaning up this abandoned config!

PR #1380, opened 2020-12-07
Remove S3 lifecycle rule for business support database backup

theseanything:
The backup has been transferred to another department and will no longer be stored. Cleaning up this extra lifecycle rules as its no longer needed.


gclssvglx (approved):
LGTM 👍

PR #1381, opened 2020-12-07
Add RateLimit WAF v2 rule

bilbof:
> Note: We need to upgrade our Terraform version to `>0.12` in order to merge this. This PR requires Terraform AWS provider version [3.3.0](https://github.com/hashicorp/terraform-provider-aws/blob/master/CHANGELOG.md#330-august-20-2020), which introduces the `forwarded_ip_config` block to `rate_based_statement`. However, this provider is not supported by our Terraform version (0.11). Version 3.3.0 of the AWS provider only supports Terraform versions 0.12 and up. Leaving in draft state until we can enable this.

Two changes:

- Upgrades the WAF.

WAF v2 is the latest version of the AWS WAF, replacing WAF Classic in November 2019.

Upgrading the WAF brings us some benefits. Importantly for our use case, WAF v2 permits matching requests with rules using the FORWARDED_IP rather than the IP. This is required for rate limiting since we would like to rate limit users based on the True-Client-IP (the real IP address of the user) rather than using the IP addresses of Fastly.

- Adds RateLimit WAF v2 rule (count only).

The rate limit rule uses a managed rate limiter provided by AWS to limit the requests per user IP to 2000 per 5 minute period. This rule has the action 'count'. This means we will log the number of requests which would have been blocked, but this rule will not block any requests. Should we see elevated requests from a small set of IPs we should be able to easily set this rule to `block {}`.

This rule makes use of the Forwarded IP feature new to AWS WAF, it enables us to use the True-Client-IP set in our VCL, which is a header with a value of the real user's client IP.

Links

Forwarded IP: https://docs.aws.amazon.com/waf/latest/developerguide/waf-rule-statement-forwarded-ip-address.html
RateBasedStatement: https://docs.aws.amazon.com/waf/latest/APIReference/API_RateBasedStatement.html
True-Client-IP: https://github.com/alphagov/govuk-cdn-config/blob/4a7e2bcf6fd64217631a4ef4e59877b3fe8535c9/vcl_templates/www.vcl.erb#L311


bilbof:
Closing this, we can restore the branch when we upgrade our Terraform version

PR #1385, opened 2020-12-14
Scale mapit up from 8 to 12 instances

AlanGabbianelli:
The architecture of mapit is different from that of other GOV.UK apps and more work is needed to scale it up and spin up more instances, including:
- making the changes in [this `govuk-aws` PR](https://github.com/alphagov/govuk-aws/pull/1385)
- making the changes in [this `govuk-aws-data` PR](https://github.com/alphagov/govuk-aws-data/pull/817)
- going through [the `Fix Mapit database not available` docs](https://docs.publishing.service.gov.uk/manual/mapit-database-not-available.html)

This would be the right time to refactor and DRY this up but, unfortunately, a time-sensitive constraint means there's no time for it today.

Inspired by:
- https://github.com/alphagov/govuk-aws/pull/1351
- https://github.com/alphagov/govuk-aws/pull/1353

### DIFF
--- a/terraform/projects/app-mapit/README.md
+++ b/terraform/projects/app-mapit/README.md
@@ -32,6 +32,9 @@ Mapit node
 | internal\_domain\_name | The domain name of the internal DNS records, it could be different from the zone name | `string` | n/a | yes |
 | internal\_zone\_name | The name of the Route53 zone that contains internal records | `string` | n/a | yes |
 | lc\_create\_ebs\_volume | Creates a launch configuration which will add an additional ebs volume to the instance if this value is set to 1 | `string` | n/a | yes |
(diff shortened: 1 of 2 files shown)


richardTowers:
I think you'll need a corresponding PR in govuk-aws-data like this one:

https://github.com/alphagov/govuk-aws-data/pull/785

AlanGabbianelli:
Thanks @richardTowers, I updated the PR description.

richardTowers (approved):
Looks good to me.

I think it would be easy enough to refactor this in the future to just take three subnets as variables (one for each availability zone in AWS), and then use them multiple times. So mapit_4 could use subnet_1, mapit_5 could use subnet_2, mapit_6 could use subnet_3 and mapit_7 could go back to subnet_1. Then we wouldn't need the aws-data PRs to scale this up and down.

PR #1388, opened 2020-12-15
Actions: Add --deployment to bundle install

richardTowers:
Also adding some caching, following some of the guidance in https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html#simple-ruby-application

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ jobs:
         with:
           path: vendor/bundle
           key: bundle-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: bundle
       - run: bundle install --deployment
 
       - name: Docs Check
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -5,9 +5,13 @@ jobs:
...
-
-      - name: bundle install
-        run: bundle install
+      - uses: ruby/setup-ruby@v1
+      - uses: actions/cache@v1
+        with:
+          path: vendor/bundle
+          key: bundle-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: bundle
+      - run: bundle install --deployment
```
kevindew, line 14:
```suggestion
      - run: bundle install --deployment
      - run: bundle install --jobs 4 --retry 3 --deployment
```

Will make it just a bit more consistent and a little faster.

kevindew (approved):
Looks good,  I can see master failed on a [rebuild](https://github.com/alphagov/govuk-aws/runs/1557599998) so this looks like it will fix it. 

I've just added a non-blocking suggestion.

And as an aside, sooner or later we'll need to update our docs and our practices to deal with Bundler's frustrating deprecation of `--deployment` for the more verbose: `bundle config set --local deployment 'true'`

PR #1391, opened 2020-12-16
Use one subnet per availability zone in AWS

AlanGabbianelli:
This is a refactor commit, it doesn't change any functionality.

Currently, we use one subnet variable for each mapit instance and EBS volume. This means for each new mapit instance we add here in `govuk-aws`, we add a new variable in `govuk-aws-data`. This makes scaling mapit up or down more time-consuming and more error prone (the more code we add or change, the higher the chance to introduce bugs).

https://github.com/alphagov/govuk-aws-data/pull/820 introduced `mapit_subnet_a`, `mapit_subnet_b` and `mapit_subnet_c`, and removed the numbered variables, making this change necessary.

This means in the future we'll only have to make changes to `govuk-aws` to scale mapit up or down.


sengi (approved):
Definitely an improvement, thanks!

I verified there are no diffs (together with alphagov/govuk-aws-data#820).

```
$ gds aws govuk-test-admin -e staging -- tools/build-terraform-project.sh -c plan -d ../govuk-aws-data/data -e staging -s blue -p app-mapit
...
No changes. Infrastructure is up-to-date.
```

richardTowers (approved):
Thanks for this Alan, big improvement!

PR #1393, opened 2020-12-22
Create a shared memcached cluster for Mapit

richardTowers:
infra-security-groups will need to be applied first, as app-mapit
depends on a new output.

The r6g instance type is the default if you click through in the AWS
console. AWS suggest that the ARM based processers work well with
elasticache:

https://aws.amazon.com/about-aws/whats-new/2020/10/amazon-elasticache-now-supports-m6g-and-r6g-graviton2-based-instances/

The choice between m6g and r6g is arbitrary - r6g comes first in the
dropdown in the console.

The choice of `large` (rather than `xlarge` or higher) is also
arbitrary. I'm assuming that the failure mode for this will be graceful,
as it's just a cache, so starting small and scaling based on metrics
feels reasonable. You could argue the other way though (start big and
make it smaller based on metrics).

### DIFF
--- a/terraform/projects/app-mapit/main.tf
+++ b/terraform/projects/app-mapit/main.tf
@@ -154,6 +154,23 @@ resource "aws_route53_record" "mapit_service_record" {
   }
 }
 
+resource "aws_elasticache_subnet_group" "cache" {
+  name       = "${var.stackname}-mapit-cache"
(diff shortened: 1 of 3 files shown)


richardTowers:
Closing - the cache should be used by collections, not mapit.

PR #1394, opened 2020-12-22
Create a shared memcached cluster for Frontend

richardTowers:
The theory here is that having a shared rails cache should improve
performance of the frontend applications now that we've scaled them
up to a large number of instances.

We're particularly hoping to see improvements between `collections` and
`mapit`, as the responses from mapit are highly cacheable (postcodes
don't change local authority very often.

`infra-security-groups` will need to be applied first, as `app-frontend`
depends on a new output.

The `r6g` instance type is the default if you click through in the AWS
console. AWS suggest that the ARM based processers work well with
elasticache:

https://aws.amazon.com/about-aws/whats-new/2020/10/amazon-elasticache-now-supports-m6g-and-r6g-graviton2-based-instances/

The choice between `m6g` and `r6g` is arbitrary - `r6g` comes first in
the dropdown in the console.

The choice of `large` (rather than `xlarge` or higher) is also
arbitrary. I'm assuming that the failure mode for this will be graceful,
as it's just a cache, so starting small and scaling based on metrics
feels reasonable. You could argue the other way though (start big and
make it smaller based on metrics).

## Release Plan

* Merge this PR
* Apply `infra-security-groups` from master in integration
* Apply `app-frontend` from master in integration
* Apply `infra-public-services` from master in integration
* Apply `infra-security-groups` from master in staging
* Apply `app-frontend` from master in staging
* Apply `infra-public-services` from master in staging
* Apply `infra-security-groups` from master in production
* Apply `app-frontend` from master in production
* Apply `infra-public-services` from master in production
* Update https://github.com/alphagov/govuk-puppet/pull/10896 to target integration and staging, but leave production using local caches
* Merge govuk-puppet
* Apply govuk-puppet from master in integration
* Apply govuk-puppet from master in staging
* Apply govuk-puppet from master in production (should be almost a no-op - setting MEMCACHE_SERVERS to 127.0.0.1:11211 which is the default)
* Test the performance of the shared cache in staging (because we've traffic replay there, so it's more realistic)
* If we're happy, raise a govuk-puppet PR to use the shared cache in production
* ????
* Profit!

* Raise a PR in govuk-aws-data to scale down the elasticache memcached instance in integration

### DIFF
--- a/terraform/projects/app-frontend/main.tf
+++ b/terraform/projects/app-frontend/main.tf
@@ -178,6 +178,25 @@ module "internal_lb_rules" {
   default_tags           = "${map("Project", var.stackname, "aws_migration", "frontend", "aws_environment", var.aws_environment)}"
 }
 
+resource "aws_elasticache_subnet_group" "cache" {
+  name       = "${var.stackname}-frontend-cache"
+  subnet_ids = ["${data.terraform_remote_state.infra_networking.private_subnet_ids}"]
+}
+
+resource "aws_elasticache_cluster" "memcached" {
+  cluster_id = "${var.stackname}-frontend-cache"
+
+  engine          = "memcached"
+  engine_version  = "1.6.6"
+  port            = 11211
+  node_type       = "cache.r6g.large" # Memory optimized Graviton2 ($0.206/hour as of 2020)
+  num_cache_nodes = 1
+
+  parameter_group_name = "default.memcached1.6"
+  subnet_group_name    = "${aws_elasticache_subnet_group.cache.name}"
+  security_group_ids   = ["${data.terraform_remote_state.infra_security_groups.sg_frontend_cache_id}"]
+}
+
 module "frontend" {
   source                            = "../../modules/aws/node_group"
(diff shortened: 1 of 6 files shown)


barrucadu (approved):
Looks good to me

terraform/projects/app-frontend/main.tf
```diff
@@ -178,6 +185,33 @@ module "internal_lb_rules" {
   default_tags           = "${map("Project", var.stackname, "aws_migration", "frontend", "aws_environment", var.aws_environment)}"
 }
 
+resource "aws_elasticache_subnet_group" "cache" {
+  name       = "${var.stackname}-frontend-cache"
+  subnet_ids = ["${data.terraform_remote_state.infra_networking.private_subnet_ids}"]
+}
+
+resource "aws_elasticache_cluster" "memcached" {
+  cluster_id = "${var.stackname}-frontend-cache"
```
richardTowers, line 194: I'm just going to rename this to memcahced, as the other names have `redis` in the name, so it feels more consistent.

PR #1396, opened 2021-01-04
Improve health detection for Content Store

benthorner:
https://trello.com/c/EADdQIOp/277-investigate-improving-healthchecks-for-content-store-and-other-lbs

This makes a couple of changes to the LB and Target Group
config, in response to an incident on Staging. Making these
changes brings the config for Content Store on a par with
most other apps. Please see the commits for more details.

### DIFF
--- a/terraform/projects/infra-public-services/main.tf
+++ b/terraform/projects/infra-public-services/main.tf
@@ -954,7 +954,7 @@ module "content-store_public_lb" {
     "HTTPS:443" = "HTTP:80"
   }
 
-  target_group_health_check_path = "/_healthcheck"
+  target_group_health_check_path = "/_healthcheck_content-store"
   subnets                        = ["${data.terraform_remote_state.infra_networking.public_subnet_ids}"]
   security_groups                = ["${data.terraform_remote_state.infra_security_groups.sg_content-store_external_elb_id}"]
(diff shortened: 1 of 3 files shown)


hannako (approved):
The code changes make sense based on the explanations in the commit messages, but I'm not very familiar so might be worth getting someone else to double check.

terraform/projects/infra-public-services/main.tf
```diff
@@ -954,7 +954,7 @@ module "content-store_public_lb" {
     "HTTPS:443" = "HTTP:80"
   }
 
-  target_group_health_check_path = "/_healthcheck"
+  target_group_health_check_path = "/_healthcheck-content-store"
```
bilbof, line 957: Do you need an underscore rather than a dash?

```diff
-  target_group_health_check_path = "/_healthcheck-content-store"
+  target_group_health_check_path = "/_healthcheck_content-store"
```

E.g. https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/infra-public-services/main.tf#L1225

bilbof (changes requested):

terraform/projects/infra-public-services/main.tf
benthorner: Oh good spot, thanks! I've amended the last commit 👍.

bilbof (approved):

PR #1399, opened 2021-01-06
Update Ask Export IAM policy

theseanything:
This adds permissions to manage object ACLs because permission needs to be given to the cross account bucket owner to access the objects.

### DIFF
--- a/terraform/policies/ask_export_s3_writer_policy.tpl
+++ b/terraform/policies/ask_export_s3_writer_policy.tpl
@@ -14,7 +14,11 @@
         {
             "Effect": "Allow",
             "Action": [
-                "s3:PutObject"
+                "s3:PutObject",
+                "s3:PutObjectAcl",
+                "s3:PutObjectVersionAcl",
+                "s3:GetObjectAcl",
+                "s3:GetObjectVersionAcl"
             ],
             "Resource": [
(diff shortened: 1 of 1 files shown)


terraform/policies/ask_export_s3_writer_policy.tpl
```diff
@@ -14,7 +14,11 @@
         {
             "Effect": "Allow",
             "Action": [
-                "s3:PutObject"
+                "s3:PutObject",
+                "s3:PutObjectAcl",
+                "s3:PutObjectVersionAcl",
+                "s3:GetObjectAcl",
+                "s3:PutObjectVersionAcl"
```
sengi, line 21: There's a duplicate of `s3:PutObjectVersionAcl` here. Is it meant to be GetObjectVersionAcl?
```suggestion
                "s3:GetObjectAcl",
                "s3:GetObjectVersionAcl"
```

fredericfran-gds (approved):
looks good

sengi (approved):

terraform/policies/ask_export_s3_writer_policy.tpl
theseanything: 🦅 👁️ - good catch!

PR #1401, opened 2021-01-07
Add logging for queries which are waiting on locks to be relased

thomasleese:
See individual commits for more thorough details of each change.

This change is designed to help us debug issues with slow queries. Specifically, it came out of some work I've been doing to investigate timeouts in the Email Alert API.

These changes are mostly useful for Email Alert API, but I've decided to apply them across all our databases for consistency.

[Trello Card](https://trello.com/c/AHKVvZb2/2290-investigate-500-errors-in-email-alert-frontend)

### DIFF
--- a/terraform/projects/app-content-data-api-postgresql/main.tf
+++ b/terraform/projects/app-content-data-api-postgresql/main.tf
@@ -127,6 +127,11 @@ resource "aws_db_parameter_group" "content_data_api" {
   }
 
   parameter {
+    name  = "deadlock_timeout"
+    value = 2500
+  }
+
+  parameter {
     name  = "log_lock_waits"
     value = true
   }
--- a/terraform/projects/app-content-data-api-postgresql/main.tf
+++ b/terraform/projects/app-content-data-api-postgresql/main.tf
@@ -126,6 +126,11 @@ resource "aws_db_parameter_group" "content_data_api" {
     value = "all"
   }
 
+  parameter {
+    name  = "log_lock_waits"
+    value = true
+  }
+
   tags {
     aws_stackname = "${var.stackname}"
   }
--- a/terraform/projects/app-postgresql/main.tf
+++ b/terraform/projects/app-postgresql/main.tf
@@ -116,6 +116,11 @@ resource "aws_db_parameter_group" "postgresql_pg" {
     name  = "log_statement"
     value = "all"
   }
+
+  parameter {
+    name  = "log_lock_waits"
+    value = true
+  }
 }
 
 module "postgresql-primary_rds_instance" {
--- a/terraform/projects/app-postgresql/main.tf
+++ b/terraform/projects/app-postgresql/main.tf
@@ -118,6 +118,11 @@ resource "aws_db_parameter_group" "postgresql_pg" {
   }
 
   parameter {
+    name  = "deadlock_timeout"
+    value = 2500
+  }
+
+  parameter {
     name  = "log_lock_waits"
     value = true
   }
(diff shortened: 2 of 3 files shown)


richardTowers (approved):
These look fine to me.

Logging lock waits seems totally uncontroversial, especially since you're investigating this area.

As you mention in your commit message, changing the deadlock_timeout feels a bit more controversial. I guess the effect of this is that users will have to wait 2.5 secs for an error, instead of 1 sec. If it was just email API that would be no problem, as it's all async. It's a shame to have to do it for all of our DBs, but since we share RDS instances aggressively I don't think that's entirely avoidable. Might be worth considering only updating it on app-postgresql, rather than on transition and content-data's DBs? But happy for this to be merged if you think it's better to be consistent.

benthorner:
Also agree with the first commit. I'm a bit confused by the motivation for the second, though, and in general for this PR. As I understand it, if the issues were due to deadlock, we would see a "deadlock detected" message in our logs. But I don't think we are. The first commit only seems to have value in the presence of deadlock, but again that doesn't seem to be the cause of the issue @thomasleese is investigating (unless I've missed something).

Could you help me understand the reasoning a bit better?

thomasleese:
> Could you help me understand the reasoning a bit better?

My understand of the documentation is that the `log_locks_wait` option will give us a log entry if a query is waiting on any kind of lock - not just if it's deadlocked. It just happens to use the `deadlock_timeout` option to decide how long to wait before logging that a query is waiting.

So my main justification _in this PR_ for increasing the `deadlock_timeout` option was to make it easier to spot these log entries by increasing the threshold and therefore reducing any potential noise. Now that has the side effect of also changing how long before deadlocks will be detected, but my reading of the documentation suggested that the default of 1 second is perhaps too short for our purposes anyway.

benthorner:
Oh, I see. Thanks for explaining again. I got a bit confused by read [this page](https://dba.stackexchange.com/questions/249901/what-is-the-performance-cost-of-enabling-log-lock-waits-on-postgres), but I think the interpretation is actually that these logs are written anyway, irrespective of actual deadlock.

The second commit still seems a bit arbitrary: do we have any evidence the initial output would be noisy? I don't think we expect any of our code to be waiting more than a second to acquire a lock, so any log entries could be symptomatic of bugs, and should hopefully be easy to filter out if they aren't. While deadlock should also hopefully be rare, I think @richardTowers has a point that this increases the response time in that scenario.

Would it be a problem to try without the second commit, initially?

(Correction: Email Alert API does serve user-facing requests synchronously, via various frontend apps.)

sengi (approved):
SGTM - reasoning for both changes seems sound.

I don't think the controversy over deadlock_timeout is warranted, given what you've pointed out. It seems a very low-risk change to me for all the reasons already mentioned. I say go for it.

thomasleese:
> The second commit still seems a bit arbitrary: do we have any evidence the initial output would be noisy? I don't think we expect any of our code to be waiting more than a second to acquire a lock, so any log entries could be symptomatic of bugs, and should hopefully be easy to filter out if they aren't. While deadlock should also hopefully be rare, I think @richardTowers has a point that this increases the response time in that scenario.

Yeah, my thinking that it could be noisy is just a hunch at this point, and I agree, the second commit is definitely not necessary for the logging which is the main motivation for this change. I also agree that 2.5 seconds is completely arbitrary.

However, it does appear that, based on reading the documentation, increasing the timeout a little bit could be beneficial for us. Given the docs say "The check for deadlock is relatively expensive" it made me wonder whether we might have a small performance boost if, for example, queries which normally take ~10 seconds to run (of which we definitely have some in the Publishing API when looking up content which are unlikely to be optimised further any time soon) only check for deadlocks 4 times rather than 10 times.

If we decide to take it out of this PR, I do think this is at least something we should investigate further.

benthorner (approved):
> only check for deadlocks 4 times rather than 10 times

That's a helpful argument. Since we don't expect deadlock for user-facing queries, the performance boost does seem to outweigh any negative user impact.

PR #1402, opened 2021-01-07
Revert "Improve health detection for Content Store"

benthorner:
Reverts alphagov/govuk-aws#1396

Referring to [this commit](https://github.com/alphagov/govuk-aws/pull/1396/commits/3f9393135177020393d79a993d053352b37d45f1), unfortunately this doesn't work because our Staging and Production environments don't have a "aws_migration" flag set, which means the "lb_healthchecks" file doesn't exist, which means we don't expose the app-specific healthcheck routes used in this PR.

We'll need to investigate exposing those routes in Puppet, first.


bilbof (approved):
Ah sorry about that, @benthorner

PR #1410, opened 2021-01-26
Upgrade infra security to terraform 0.13.6

karlbaker02:
This PR upgrades Terraform to version `0.13.6` for project `infra-security`.

Trello: https://trello.com/c/kkFNCkop


.github/workflows/ci.yml
```diff
@@ -32,11 +32,29 @@ jobs:
...
+      - name: terraform fmt
+        run: |
+          formatter_path="$(pwd)/tools/terraform-format.sh"
+
+          # Loop over each directory containing tf files and ensure correct format
+          for dir in $(find ~+ -type f -name '*.tf' -not -path '$(pwd)/.terraform/*' | sed -E 's|/[^/]+$||' | sort -u); do
+            cd $dir
+            echo "$dir"
+            find . -name '*.tf' | xargs $formatter_path
+          done
```
richardTowers, line 57: Gosh, this is complicated...

I'm not sure when it was added, but modern terraform fmt has a -recursive flag which probably does what you want without all these shenanigans (https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/ci.yml#L31)

richardTowers (approved):
Terraform upgrade looks good to me.

I have a bit of a nitpick around the complexity of the shell scripting in the terraform fmt github action, but I'm not too fussy about that if it can't be fixed in a way that's backwards compatible with old terraform.

.github/workflows/ci.yml
karlbaker02: The reason we're doing this is because `tfenv` is going to magically pick-up the correct Terraform version based on the presence of a `.terraform-version` file located in the directory into which we're `cd`ing. Previously we were just pulling out all of the Terraform files with the `find` command, but this isn't going to work now that we are using multiple versions of Terraform within the repo, as each version requires slightly different syntax.

PR #1411, opened 2021-01-27
Increase content store instance size

theseanything:
This changes the default instance size from m5.large to m5.2xlarge. The content store application currently runs with 15 workers and a large volume of requests quickly leads to CPU exhaustion ( as m5.large only has 2 vCPUs). This can cause increase latency in response times as requests wait for CPU cycles. An m5.2xlarge has 8 vCPUs which in load testing has improved throughput and latency.


rtrinque (approved):
Looks good to me, note that your change will apply to integration staging and prod, so you will need to terraform those environment as well. If your intent was to modify prod only you need to override the instance type variable in govuk-aws-data/data/app-content-store/production/common.tfvars

theseanything:
@rtrinque - good shout, I think I'll just need to add an override for integration as we don't need the capacity in that env.

PR #1412, opened 2021-01-27
Enable new iam roles

fredericfran-gds:
We want to modify IAM privileges so that:
1. admin, internal-admins, platformhealth-powerusers and powerusers can now assume admin role, aka govuk-administrators
2. admin, internal-admins, platformhealth-powerusers and powerusers can now assume the lower privilege powerusers role, aka govuk-powerusers since they are admins
3. admin, internal-admins, platformhealth-powerusers and powerusers can now assume the lower privilege users role, aka govuk-users (read-only) since they are admins.

This PR creates invidual roles with suffix firstname.lastname-role. Users of gds-cli will automatically have their usual cmd role map into these new roles.

Ref:
[trello card](https://trello.com/c/kkFNCkop/335-give-2ndline-iam-admin-access-to-aws)

Co-Authored-By: Karl Baker <karl.baker@digital.cabinet-office.gov.uk>


terraform/modules/aws/iam/gds_user_role/main.tf
```diff
@@ -0,0 +1,92 @@
...
+#--------------------------------------------------------------
+
+locals {
+  list_user_and_policy_arns = flatten([for user_arn in var.role_user_arns : [
+    for policy_arn in var.role_policy_arns : {
+      user_arn   = user_arn
+      policy_arn = policy_arn
+    }
+    ]
+  ])
```
richardTowers, line 46: I think this could be done slightly more neatly with [`setproduct`](https://www.terraform.io/docs/language/functions/setproduct.html).

Something like:

```terraform
  list_user_and_policy_arns = [
    for pair in setproduct(var.role_user_arns, var.role_policy_arns):
      {
        user_arn: pair[0],
        policy_arn: pair[0]
      }
  ]
```

terraform/modules/aws/iam/gds_user_role/main.tf
```diff
@@ -0,0 +1,92 @@
...
+        }
+      }
+    ]
+  }
+  EOF
+}
+
+resource "aws_iam_role_policy_attachment" "gds_user_role_policy_attachments" {
+  for_each = {
+    for pair in local.list_user_and_policy_arns : "${pair.user_arn}.${pair.policy_arn}" => pair
```
richardTowers, line 87: Not necessarily the end of the world, but I worry that the names terraform gives these resources in the statefile are going to be horrible if they've got two ARNs as the key...

I haven't checked the plan though - does it do something clever?

terraform/modules/aws/iam/gds_user_role/README.md
```diff
@@ -0,0 +1,30 @@
+## Module: aws/iam/gds\_user\_role
+
+This module creates a role with a trust relationship with IAM users from  
+a trusted account. We need to specify the list of trusted users and  
+policies to attach to the role.
```
richardTowers, line 5: I don't think this description makes it clear that this module will create _multiple_ roles, one for every combination of user and policy.

terraform/modules/aws/iam/gds_user_role/main.tf
richardTowers:
```
# module.gds_role_user.aws_iam_role_policy_attachment.gds_user_role_policy_attachments["arn:aws:iam::622626885786:user/roch.trinque@digital.cabinet-office.gov.uk.arn
:aws:iam::aws:policy/ReadOnlyAccess"] will be created
```
... so, a little bit gross. I don't think it will cause any issues, but it's not great.

terraform/modules/aws/iam/gds_user_role/main.tf
richardTowers: Maybe one step further, you could regex out the user name and policy name at this point, and then you could use them in your keys further down.

So:

```terraform
  list_user_and_policy_arns = [
    for pair in setproduct(var.role_user_arns, var.role_policy_arns):
      {
        user_arn: pair[0],
        policy_arn: pair[1],
        user_name: regex("^.*/(.+)@.*$", pair[0])[0],
        policy_name: regex("^.*/(.+)$", pair[1])[0],
      }
  ]
```

fredericfran-gds: Thanks, I was planning to do that after your first set of comments.

terraform/modules/aws/iam/gds_user_role/main.tf
fredericfran-gds: using the above new tuple, we can sort this out.

terraform/modules/aws/iam/gds_user_role/README.md
fredericfran-gds: I can update that to make it clearer

richardTowers (approved):
Absolutely cracking work, love it!

fredericfran-gds:
thanks.

richardTowers:
Hmmm, we should probably have had "Dismiss stale pull request approvals when new commits are pushed" on for this repo. I've enabled that now.

That `depends_on` doesn't look like it's going to work to me (shouldn't it refer to the specific role resource from the for_each?).

PR #1413, opened 2021-02-03
Update Mapit ELB health checks

theseanything:
This adds a query string so the health checks bypass the Ngnix cache and monitor the application.


gclssvglx (approved):
LGTM

PR #1416, opened 2021-02-12
increase max ttl for new gds roles

fredericfran-gds:
The max default ttl is 1 hour but before roles could be up to 8 hours. This PR set the new gds roles to have the same ttl.


karlbaker02 (approved):
👍

PR #1420, opened 2021-02-16
sync gcs later

fredericfran-gds:
This PR syncs GCS mirror buckets from S3 mirror buckets later. The mirrorer machine is configured/dimensioned to finish mirroring before 08 00 in the morning. The S3 sync from the the mirrorer machine takes on average 8 hours.

Hence we should start no earlier than 16 00. We add some extra hours to take into account for undercertainty in the duration of the crawling and syncing and also increase in number of assets/web pages on GOV.UK.


rtrinque (approved):
Looks good to me

PR #1422, opened 2021-02-18
Add a shared elasticache instance for Mapit

theseanything:
Mapit currently runs memcached per instance of mapit. This decreases our caching ability if we have instance number >= 2. Introducing a shared cache will allow us to scale Mapit if we need to handle load without impacting of cache. It also allows us to pre-warm our cache more easily, as we don't have to pre-warm every instance of Mapit.


rtrinque (approved):
Looks good

PR #1425, opened 2021-02-19
Scale up Mapit instance size and number

theseanything:
Mapit currently runs with 12 workers and Postgres on each instance. CPU exhaustion can been seen when handling a high number of uncached requests. Increasing the instance size from 2vCPUs to 8vCPUs should improve the number of uncached requests handled.

To this we need to scale up the number of instances from 2 to 4, so we have can rotate the old instances without downtime.


brucebolt (approved):
Looks good to me 👍

PR #1427, opened 2021-02-22
Scaling up Mapit instance size to c5.2xlarge

theseanything:
Upgrading the instance size from a burstable t3.xlarge to more reliable non-burstable c5.2xlarge instance type.


kevindew (approved):
Looks good to me 👍

PR #1433, opened 2021-03-29
Update knowledge graph startup time

chao-xian:
To accomodate the clock changes to BST and match working hours.


oscarwyatt (approved):
Super

PR #1434, opened 2021-03-31
Update data science data generation schedule

chao-xian:
To accommodate the clock changes to BST and match working hours.

See also https://github.com/alphagov/govuk-aws/pull/1433


oscarwyatt (approved):
Super

PR #1435, opened 2021-04-07
Switch RDS log exporter lambda to Python 3.8 + cleanup.

sengi:
* Make it work with Python 3.8.
* Increase the timeout so that in case it runs slowly for any reason, it'll be less likely to fail and never recover because after a failed run there would be an ever-increasing set of files to copy. (It only writes the timestamp file at the end of a successful run; it doesn't save intermediate progress.)
* Fix the instructions in the README so that it's possible to test locally.
* Format to PEP8, remove unused imports, fix lint errors and poorly-written conditionals.
* Simplify the logic and remove some redundant variables.
* Simplify error handling and include the original error messages from the API so the messages are more informative as to what failed.
* Use a buffer instead of concatenating immutable strings on every read.
* Start decomposing the logic into functions (only factored out one function so far but it's a start).
* Add a couple of tests.

This still isn't what I would consider production-grade because of the lack of tests, but at least it's in a much better state than before.

Tested:
* Fairly extensive manual testing in the integration account using `sam local invoke RDSLogsToS3 -e event.json` (see README) that the behaviour is still the same as before. Tested both the first-run (no timestamp file) case and the incremental (timestamp file exists so only copy newer files).
* The unit tests are passing (for what they're worth). They only cover reading the timestamp file though.

### DIFF
--- a/terraform/lambda/RDSLogsToS3/Makefile
+++ b/terraform/lambda/RDSLogsToS3/Makefile
@@ -1,0 +1,12 @@
+testenv : virtualenv localstack
+
+virtualenv:
+	python3 -m venv env
+	source env/bin/activate && python3 -m pip install -r requirements.txt
+
+localstack:
+	brew install localstack
+	localstack start
+
+test:
+	source env/bin/activate && python3 -m unittest main_test.py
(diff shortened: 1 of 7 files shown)


sengi:
Looks like nothing is using these logs (couldn't find anything in our repos and Cybersecurity says they're not using them either). So let's just get rid of this and switch on RDS logging to CloudWatch instead. The volume is single-digits GB/month so the ingestion cost is negligible.

PR #1439, opened 2021-04-22
Remove AWS lambdas running Python 2.7

brucebolt:
Python 2.7 is being deprecated from AWS Lambda on 15th July 2021.  These lambda functions export raw logs to S3.  However we already have these logs being stored in AWS CloudWatch, so these lambda functions (and the S3 log dumps) are now redundant, so the lambda functions can be removed.

Trello card: https://trello.com/c/7uaFnYGD


thomasleese (approved):
Nice to see this code going away!

PR #1440, opened 2021-04-22
Add Max Froumentin's ssh key

maxf:


oscarwyatt (changes requested):
You'll need to add yourself to `/terraform/projects/app-knowledge-graph/userdata-sandbox.tpl` to get access to the lab machine too, otherwise, looks good

oscarwyatt (approved):

PR #1444, opened 2021-04-27
Fix ITHC HTTPS access for mapit and support-api

barrucadu:
HTTPS rules need to belong to a security group which is attached to a
load balancer, which the previous groups weren't.  Since we already
have security groups attached to the public load balancers (defined in
infra-public-services), I've added the rules to them.


karlbaker02:
If you were to instead keep the security group rules attached to the `support-api_ithc_access` security group and then add that security group as an _addition_ security group to the appropriate load balancers, you'd then be able to keep the security group in place when testing is finished (as you'd simply remove the new security group from the load balancer, rather than needing to modify the active security group rules to remove access, at which point you'd need to repeat this work in the future).

Hope this makes sense, let us know if not.

barrucadu:
> rather than needing to modify the active security group rules to remove access, at which point you'd need to repeat this work in the future).

When the test is over I'll set the `ithc_access_ips` to `[]`, rather than delete the rule:  https://github.com/alphagov/govuk-aws-data/blob/master/data/common/staging/common.tfvars#L25

karlbaker02 (approved):
Looks good 👍

PR #1441, opened 2021-04-23
Upgrade ArtefactSync lambda to Python 3.8

brucebolt:
AWS are deprecating the support for lambdas running Python 2.7 on the 15th July 2021, so this lambda needs updating to Python 3.

All other Python 2.7 lambdas are no longer needed, so will be removed in https://github.com/alphagov/govuk-aws/pull/1439.

Trello card: https://trello.com/c/7uaFnYGD

PR #1445, opened 2021-05-07
Fix the ArtefactSync lambda and the infra-public-services module.

sengi:
* Remove the manual build step for this lambda that led to it becoming broken in #1441. Terraform now builds the zip file from the source code and deploys it to lambda, instead of deploying a manually-build zip that's checked into the repo 🤦 
    * This fixes the lambda and prevents the same issue from reoccurring.
* Pin the version of the `archive` Terraform provider. Our Terraform version is so old that it doesn't recognise the certificate that recent provider versions are signed with.
    * Apart from being necessary for eliminating the checked-in zip file, this also fixes the `infra-public-services` Terraform module, which was broken because it also had an unpinned version of the `archive` provider.

Tested:
* Applied `infra-artefact-bucket` in the staging account.
* Ran `terraform plan` for `infra-public-services`; now works instead of erroring out on the provider signing issue.

TO DO:
* Test the lambda


bilbof (approved):
Great work, thanks for the detailed explanation.

thomasleese:
Thanks for spotting this!

PR #1450, opened 2021-06-04
Set up public load balancer for account-api

barrucadu:
This is safe because all the API endpoints are protected by Signon
bearer tokens.

---

Plans:

- [infra-public-services](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/1309/console)
- [infra-security-groups](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/1308/console)

[Trello card](https://trello.com/c/6CsImbRT/817-make-account-manager-able-to-communicate-with-account-api)

### DIFF
--- a/terraform/projects/infra-security-groups/account.tf
+++ b/terraform/projects/infra-security-groups/account.tf
@@ -92,7 +92,7 @@ resource "aws_security_group" "account_elb_external" {
   }
 }
 
-resource "aws_security_group_rule" "account-elb-external_ingress_office_https" {
+resource "aws_security_group_rule" "account-elb-external_ingress_any_https" {
   type      = "ingress"
   from_port = 443
   to_port   = 443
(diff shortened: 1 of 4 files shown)


terraform/projects/infra-security-groups/account.tf
```diff
@@ -92,15 +92,15 @@ resource "aws_security_group" "account_elb_external" {
...
+resource "aws_security_group_rule" "account-elb-external_ingress_any_https" {
   type      = "ingress"
   from_port = 443
   to_port   = 443
   protocol  = "tcp"
 
   # Which security group is the rule assigned to
   security_group_id = "${aws_security_group.account_elb_external.id}"
-  cidr_blocks       = ["${var.office_ips}"]
+  cidr_blocks       = ["0.0.0.0/0"]
```
huwd, line 103: This is interesting, No office IPs anymore?

barrucadu: This security group was actually unused (because we didn't have a public load balancer), so they weren't used.

And `0.0.0.0/0` is the entire IPv4 space, so the office IPs are a subset of that.

huwd (approved):

PR #1454, opened 2021-07-05
Use the Prometheus readiness endpoint for the LB healthcheck.

sengi:
This was previously using /metrics, which wasn't mapped through nginx so was always failing with 500 'fell through to default vhost'.

The internal LB is already set to use `/-/ready`.

Requires alphagov/govuk-puppet#11187.


bilbof (approved):
Great one, thanks for doing this.

PR #1455, opened 2021-07-06
Make Prometheus storage volume configurable and upgrade to gp3.

sengi:
This enables us to add more timeseries storage space in prod, which was running out.

gp3 is cheaper and faster, so no reason not to use it.


fredericfran-gds (approved):
looks good

PR #1458, opened 2021-07-19
Download puppetlabs GPG key over https

richardTowers:
It might be a bit tinfoil hat to consider how downloading puppet's GPG
key over a plaintext connection could lead to a vulnerability, but it
seems like a good idea to download it over HTTPS since that seems to be
available.

Confirmed that it's the same key:

```
cmp <(curl --silent 'http://apt.puppetlabs.com/DEB-GPG-KEY-puppet-20250406') <(curl --silent 'https://apt.puppetlabs.com/DEB-GPG-KEY-puppet-20250406')
```


sihugh (approved):

bilbof (approved):
🔐

karlbaker02 (approved):
👍

PR #1460, opened 2021-07-21
Add overrides for instance type and number of nodes for redis

chao-xian:
So that we can override for production with a different value in [govuk-aws-data](https://github.com/alphagov/govuk-aws-data/pull/889).

Due to increased memory usage caused by data.gov.uk recently which triggered an incident. This may get reduced once the underlying excessive memory consumption is addressed.


richardTowers (approved):
Thank you for doing this properly 🙇🏻 

I had a look at it after the incident yesterday, saw that the instance type was hardcoded, and just went "nope".

PR #1462, opened 2021-07-30
Add redis_parameter_group_name variable to aws/elasticache_redis_cluster

bilbof:
This was incorrectly ommitted when setting redis_engine_version in
a previous commit - we need to update both the parameter group name
and the engine version name at the same time.


fredericfran-gds (approved):
looks good

PR #1465, opened 2021-08-11
Output role names and ARNs for roles created via gds_user_role module

karlbaker02:
This PR updates the `gds_user_role` module to output a map containing the role name and role ARN for each created role.

For example, creating a role of `*-admin` will output all of the generated role names and ARNs in a map of the format:

```
{
  ...,

  'joe.bloggs-admin': 'arn:aws:iam::123467890123:role/joe.bloggs-admin',

  ...,
}
```

This is being used in `infra-security` to output the role names and ARNs when creating `admin`, `poweruser` and `user` roles, allowing us to use this when creating resources elsewhere, such as in our new EKS cluster.

### DIFF
--- a/terraform/modules/aws/iam/gds_user_role/README.md
+++ b/terraform/modules/aws/iam/gds_user_role/README.md
@@ -38,4 +38,6 @@ No modules.
 
 ## Outputs
 
-No outputs.
+| Name | Description |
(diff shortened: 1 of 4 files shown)


sengi (approved):
LGTM

PR #1476, opened 2021-09-20
fix GitHub IP range returning IPv6

fredericfran-gds:
Fix GitHub IP Ranges Returning IPv6
The GitHub provider is returning IPv6 IPs in addition to IPv4
IPs for the hooks. This is not compatible with AWS security
rules which require IPv4 and IPv6 IPs to be seperate.

This is a [fixed bug](integrations/terraform-provider-github#883)
in 4.14 version of the provider.

### DIFF
--- a/terraform/projects/infra-security-groups/ci-master.tf
+++ b/terraform/projects/infra-security-groups/ci-master.tf
@@ -92,7 +92,7 @@ resource "aws_security_group_rule" "ci-master-elb_ingress_github_https" {
   protocol  = "tcp"
 
   security_group_id = "${aws_security_group.ci-master_elb.id}"
-  cidr_blocks       = ["${data.github_ip_ranges.github.hooks}"]
+  cidr_blocks       = ["${data.github_ip_ranges.github.hooks_ipv4}"]
 }
 
 resource "aws_security_group_rule" "ci-master-elb_egress_any_any" {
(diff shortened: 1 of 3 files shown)


terraform/projects/infra-security-groups/ci-master.tf
```diff
@@ -92,7 +92,7 @@ resource "aws_security_group_rule" "ci-master-elb_ingress_github_https" {
   protocol  = "tcp"
 
   security_group_id = "${aws_security_group.ci-master_elb.id}"
-  cidr_blocks       = ["${data.github_ip_ranges.github.hooks}"]
+  cidr_blocks       = ["${data.github_ip_ranges.github.hooks_ipv4}"]
```
benthorner, line 95:
```suggestion
  cidr_blocks      = ["${data.github_ip_ranges.github.hooks_ipv4}"]
  ipv6_cidr_blocks = ["${data.github_ip_ranges.github.hooks_ipv6}"]
```

From [this doc](https://github.com/integrations/terraform-provider-github/blob/8a197474b6f5ec265b6ad426c0ed11cc25ffa8d8/website/docs/d/ip_ranges.html.markdown#attributes-reference) the `hooks_ipv4` returns a subset, but I suspect we want _all_ IPs in case some of the servers use an IPv6 address.

Reference: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group_rule#example-usage

fredericfran-gds: I think this is fine, IPv6 is not enable on gov.uk as far as I know.

benthorner (approved):
Great fix! The documentation for the provider seems to be out-of-date, so nice one finding the functionality we need in spite on that ⭐.

PR #1481, opened 2021-10-05
Automatically replace unhealthy Apt instances

bilbof:
This changes the health check type to ELB for the Apt autoscaling group, and increases the `asg_max_size` to 3.

The effect of this will be that apt instances will be deregistered from the ASG and replaced when their ELB
health checks fail.

I have also increased the grace period for a new instance to 5 minutes as this instance may take a while to start.

### DIFF
--- a/terraform/projects/app-apt/main.tf
+++ b/terraform/projects/app-apt/main.tf
@@ -192,6 +192,11 @@ resource "aws_route53_record" "gemstash_external_service_record" {
 module "apt" {
   source                            = "../../modules/aws/node_group"
   name                              = "${var.stackname}-apt"
+  asg_health_check_grace_period     = "300"
+  asg_health_check_type             = "ELB"
+  asg_desired_capacity              = "2"
+  asg_max_size                      = "3"
+  asg_min_size                      = "2"
   default_tags                      = "${map("Project", var.stackname, "aws_stackname", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "apt", "aws_hostname", "apt-1")}"
   instance_subnet_ids               = "${matchkeys(values(data.terraform_remote_state.infra_networking.private_subnet_names_ids_map), keys(data.terraform_remote_state.infra_networking.private_subnet_names_ids_map), list(var.apt_1_subnet))}"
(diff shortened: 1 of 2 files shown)


karlbaker02 (approved):
Looks good!

terraform/projects/app-apt/main.tf
```diff
@@ -192,6 +192,11 @@ resource "aws_route53_record" "gemstash_external_service_record" {
 module "apt" {
   source                            = "../../modules/aws/node_group"
   name                              = "${var.stackname}-apt"
+  asg_health_check_grace_period     = "300"
```
sengi, line 195: How did you determine this? (It sounds quite short to me given how long some of our Puppet / deployment automation stuff often takes, but you've based this on measuring some actual startup times then no probs.)

terraform/projects/app-apt/main.tf
```diff
@@ -192,6 +192,11 @@ resource "aws_route53_record" "gemstash_external_service_record" {
...
+  asg_health_check_type             = "ELB"
```
sengi, line 196: I would probably put a comment here explaining that this is essentially a whack-a-mole type of workaround for an unsolved intermittent issue with these instances (and link to the Trello card that tracks the issue, if there is one).

sengi (approved):
I'm a bit concerned that this effectively just masks the underlying issue which has been causing these VMs to get wedged occasionally. If we're doing this consciously as a workaround because we don't want to invest more time in figuring out what's going wrong then that's fine, but just want to check that that's really what's intended.

Ordinarily of course it'd be fairly uncontroversial to do this, but given the state of our deployment automation it's a bit of a Faustian tradeoff 😅 (Not saying it's the wrong tradeoff,  - I'm sure you've made the right call based on what you've seen happening with these instances while oncall - just that it's worth explaining to folks what the tradeoff is.)

sengi:
Some other things to consider:

* How have you tested this?
* If this were to cause a restart loop, how easy/hard would be it be for the oncall to find out that it were happening?
* What happens when the apt services is entirely down? I'm assuming it only affects deployment changes rather than anything that's already running - is that right? If so then I think it's fine, but if it affects serving then I'd want to be quite thorough with testing it.

bilbof:
Discussed offline - @sengi made some great points here that motivate keeping the rebooting process manual, since there's not enough justification to introduce extra moving parts into our (soon to be legacy) infra.

PR #1488, opened 2021-11-03
Use cheaper and faster gp3 storage for CI agent EBS volumes.

sengi:
gp3 is 20% cheaper than gp2 and includes 3000 IOPS at any size instead of 3 IOPS/GB. This should speed up builds at least a bit.

While we're there, define the values once instead of copy-pasting.

The volumes can be upgraded in place without downtime.


sengi:
Force-push to fix up autogenerated docs.

karlbaker02 (approved):
Looks good, hopefully the performance gain will be 🚀

sengi:
And terraform fmt. Sigh.

PR #1494, opened 2021-12-08
Change m4 DB instances to m5

kevindew:
https://trello.com/c/ybS7Dnns/41-upgrade-the-legacy-m4-database-machine-instances

These are being updated because the m4 instance class is legacy and will
not support newer versions of PostgreSQL. So this is a pre-requisite
before the Databases on these instances can be upgraded to newer
versions of PostgreSQL.

As I understand, we don't use a "apply_immediately" flag in our RDS
configuration so this change will take effect during the next
maintenance window - which will be early Monday morning.


sengi (approved):
🚀

PR #1495, opened 2021-12-09
Ask export can create a bucket

kevindew:
Trello: https://trello.com/c/pmiW3jZk/249-tech-spike-ask-contingency-plan-port-the-script-we-run-on-concourse-to-jenkins

Pairs with: https://github.com/alphagov/govuk-aws-data/pull/941

This sets up a conditional terraform command to create an S3 bucket. It
has been introduced so that the GOV.UK integration environment can have
a test mechanism to run the Ask export.

The prompt for adding this was the retirement of big Concourse, which
has led to the export process being moved to Jenkins. The team need a
method to try the system out before configuring the production one.

---

I wasn't 100% sure whether the syntax for a boolean was conventionally `true` or `"true"` - I could see examples of both in this project. 

Is the convention to have a PR merged before running a plan or to wait until after merge?

### DIFF
--- a/terraform/projects/infra-ask-export/integration.govuk.digital
+++ b/terraform/projects/infra-ask-export/integration.govuk.digital
@@ -1,0 +1,4 @@
+bucket  = "govuk-terraform-steppingstone-integration"
+key     = "govuk/infra-ask-export.tfstate"
+encrypt = true
+region  = "eu-west-1"
--- a/terraform/projects/infra-ask-export/main.tf
+++ b/terraform/projects/infra-ask-export/main.tf
@@ -3,7 +3,7 @@
 *
 * GOV.UK Ask Export
 *
-* This data export process requires cross account permissions to S3 as an export target. Due to the cross account permissions, this module currently only works with the production backend currently.
+* This data export process requires cross account permissions to S3 as an export target. Outside of the production environment, this terraform can be configured to create a bucket.
 */
 
 variable "aws_region" {
@@ -27,6 +27,12 @@ variable "export_s3_bucket_name" {
   description = "Bucket name to allow write permissions"
 }
 
+variable "create_bucket" {
+  type        = "boolean"
+  description = "Whether to create the bucket, in production we're expecting to use a bucket in a different AWS account"
+  default     = "false"
+}
+
 # Resources
 # --------------------------------------------------------------
 terraform {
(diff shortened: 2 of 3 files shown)


terraform/projects/infra-ask-export/integration.govuk.digital
```diff
@@ -0,0 +1,4 @@
+bucket  = "govuk-terraform-steppingstone-integration"
+key     = "govuk/infra-ask-export.tfstate"
+encrypt = true
+region  = "eu-west-1"
```
sengi, line 4: I fairly sure Terraform is robust to this, but there's a missing line ending at the end of this file. (Editor config issue maybe? We could add a fixer for this if it happens often.)

terraform/projects/infra-ask-export/main.tf
```diff
@@ -27,6 +27,12 @@ variable "export_s3_bucket_name" {
   description = "Bucket name to allow write permissions"
 }
 
+variable "create_bucket" {
+  type        = "boolean"
+  description = "Whether to create the bucket, in production we're expecting to use a bucket in a different AWS account"
+  default     = false
```
sengi, line 33: Since this is ye olde crufty Terraform 0.11, the type system is quite tricksy. There's no boolean type for `variable`; it'll actually use string. The docs also [recommend using string literals "true" and "false" in this context](https://www.terraform.io/docs/configuration-0-11/variables.html#booleans), because TF 0.11 will coerce the bool literal to a string.
```suggestion
  description = "Whether to create the bucket, in production we're expecting to use a bucket in a different AWS account"
  default     = "false"
```

terraform/projects/infra-ask-export/main.tf
```diff
@@ -39,6 +45,17 @@ provider "aws" {
...
 
+resource "aws_s3_bucket" "ask_export_bucket" {
+  bucket = "${var.export_s3_bucket_name}"
+
+  tags {
+    Name            = "${var.export_s3_bucket_name}"
+    aws_environment = "${var.aws_environment}"
+  }
+
+  count = "${var.create_bucket ? 1 : 0}"
```
sengi, line 56: nit: I'd put the count hack right at the top for readability (I guess it's sort of a convention?)

sengi (approved):

terraform/projects/infra-ask-export/integration.govuk.digital
kevindew: Thank you! - Ironically I used an unfamiliar text editor for this as my vim does a terraform fmt that's far newer than our version of terraform causing massive diffs.

terraform/projects/infra-ask-export/main.tf
kevindew: Great thanks. Would you mind giving this a quick review: https://github.com/alphagov/govuk-aws-data/pull/942 I merged the govuk-aws-data PR before checking this one 🤦‍♂️

terraform/projects/infra-ask-export/main.tf
kevindew: Done

kevindew:
Thank you @sengi ⭐

sengi (approved):
Thanks for the fixes!

PR #1497, opened 2021-12-09
Refactor app-related-links to run via Jenkins

karlbaker02:
This PR refactors the `app-related-links` project to run via Jenkins, rather than Concourse, which is being deprecated in December 2021. It moves some of the provisioning that used to live in the `govuk-related-links-recommender` repo to two new userdata scripts in this commit, as well as moving the generation and ingestion machines from the public to the private subnet (as external access i.e. Concourse is no longer necessary).


terraform/projects/app-deploy/main.tf
```diff
@@ -287,6 +287,11 @@ resource "aws_iam_role_policy_attachment" "allow_reads_from_artefact_bucket" {
   policy_arn = "${data.terraform_remote_state.artefact_bucket.read_artefact_bucket_policy_arn}"
 }
 
+resource "aws_iam_role_policy_attachment" "related_links_jenkins_iam_role_policy_attachment" {
```
sengi, line 290: nit: since Terraform resources are namespaced by type (i.e. to refer to this resource you'd write `aws_iam_role_policy_attachment.related_links_jenkins_iam_role_policy_attachment`) it's slightly more readable to omit the suffix, like this:

```suggestion
resource "aws_iam_role_policy_attachment" "related_links_jenkins" {
```

(definitely don't worry about changing any pre-existing ones though!)

terraform/projects/app-related-links/provision-ingestion-instance.tpl
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -eo pipefail
+
+# Install updates and required packages
+sudo apt-get update -y
+sudo apt-get install -y awscli
+sudo apt-get install -y jq
+sudo apt-get install -y gnupg2
```
sengi, line 9: optional: it's probably slightly quicker (and equally readable since there's only 3 of them) to do this with one apt-get operation. The error messages on failure will be equally good as well.
```suggestion
sudo apt-get install -y awscli jq gnupg2
```
(and similarly in the generation template, if you want)

terraform/projects/app-related-links/provision-generation-instance.tpl
```diff
@@ -0,0 +1,14 @@
...
+# Install updates and required packages
+sudo apt-get update -y
+sudo apt-get install -y awscli
+sudo apt-get install -y jq
+
+echo "CONTENT_STORE_BUCKET=\"${database_backups_bucket_name}\"" >> /home/ubuntu/.profile
+echo "RELATED_LINKS_BUCKET=\"${related_links_bucket_name}\"" >> /home/ubuntu/.profile
+echo "$(aws secretsmanager get-secret-value --secret-id related_links-BIG_QUERY_SERVICE_ACCOUNT_KEY --query "SecretString" --region eu-west-1 | jq -r '.')" > /var/tmp/bigquery.json
+chmod 400 /var/tmp/bigquery.json
+chown -R ubuntu:ubuntu /var/tmp/bigquery.json
```
sengi, line 14: nit: the `-R` here is unnecessary and slightly confusing (I'm assuming that bigquery.json isn't a directory, right?)
```suggestion
chown ubuntu:ubuntu /var/tmp/bigquery.json
```

sengi (approved):
👍

terraform/projects/app-related-links/provision-generation-instance.tpl
karlbaker02: Good point, thanks 👍

terraform/projects/app-deploy/main.tf
karlbaker02: Thanks, another good point. I was following past precedent and also my own past precedent, but totally makes sense what you suggest here.

PR #1498, opened 2021-12-09
Fix terraform boolean

kevindew:
It turns out the terraform version we use, 0.11, has no concept of a
boolean and needs to be defined as a string in the type declaration.
This fixes that issue.

I did _finally_ work out to do a plan of a working branch so I'm finally
confident this works.


sengi (approved):
👍

PR #1499, opened 2021-12-10
Make elb_public_secondary_certname optional

fredericfran-gds:
Making `elb_public_secondary_certname` optional by adding an empty
string as default value. This will allow use to "migrate" from
using 2 certificate to only one certificate on the load balancers.

Ref:
1. [lb secondary cert code](https://github.com/alphagov/govuk-aws/blob/498581043d26e4e4e3474edf5d1718ad735f589f/terraform/modules/aws/lb/main.tf#L196)


sengi (approved):
LGTM

PR #1501, opened 2021-12-10
Configure postgresql-primary and mysql-primary for DMS replication

kevindew:
Trello: https://trello.com/c/OFuRPguC/35-investigate-whether-aws-dms-is-a-suitable-tool-for-large-postgresql-databases

This applies the configuration that AWS suggest (and what I've been using in the test environment) to our multi-tenant RDS instances so that they can be used as source database endpoints for DMS (Database Migration Service). The intention is to use this to start experiments migrating some of the bigger databases we host to gain confidence that this will be successful in production as a tool to minimise user downtime.

The PostgreSQL change is going to require a reboot to take effect and the MySQL one requires an additional manual query to be run.

More details in the commits.

### DIFF
--- a/terraform/projects/app-mysql/main.tf
+++ b/terraform/projects/app-mysql/main.tf
@@ -127,6 +127,22 @@ resource "aws_db_parameter_group" "mysql-primary" {
     value = 1
   }
 
+  # Configure database for AWS DMS as per: https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.MySQL.html#CHAP_Source.MySQL.AmazonManaged
+  parameter {
+    name  = "binlog_checksum"
+    value = "NONE"
+  }
+
+  parameter {
+    name  = "binlog_format"
+    value = "ROW"
+  }
+
+  parameter {
+    name  = "binlog_row_image"
+    value = "Full"
+  }
+
   tags {
     aws_stackname = "${var.stackname}"
   }
--- a/terraform/projects/app-postgresql/main.tf
+++ b/terraform/projects/app-postgresql/main.tf
@@ -126,6 +126,25 @@ resource "aws_db_parameter_group" "postgresql_pg" {
     name  = "log_lock_waits"
     value = true
   }
+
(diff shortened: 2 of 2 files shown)


terraform/projects/app-mysql/main.tf
```diff
@@ -127,6 +127,22 @@ resource "aws_db_parameter_group" "mysql-primary" {
...
+  }
+
+  parameter {
+    name  = "binlog_format"
+    value = "ROW"
+  }
+
+  parameter {
+    name  = "binlog_row_image"
+    value = "full"
```
ollietreend, line 143: From [the docs](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.MySQL.html#CHAP_Source.MySQL.AmazonManaged):
> - Set the `binlog_row_image` parameter to `"Full"`.

Should this value be capitalised? Not sure if it's case sensitive.

```suggestion
    value = "Full"
```

kevindew: Ooh - great catch. Thanks.

Interestingly in the AWS UI it's in lowercase with the available options provided as lowercase:
Screenshot 2021-12-14 at 10 37 29

It seems better to have it as per the docs though. I'll set that and try a terraform plan on it - I somewhat expect though that a Terraform plan doesn't actually validate it with AWS so I might not know for sure until we apply it.

richardTowers (approved):
Now that I've read the DMS docs and Kevin's explained a bit about how it works, I understand what we're doing here and I'm happy to approve.

On the pending-reboot thing: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_RebootInstance.html

> If a DB instance isn't using the latest changes to its associated DB parameter group, the AWS Management Console shows the DB parameter group with a status of pending-reboot. The pending-reboot parameter groups status doesn't result in an automatic reboot during the next maintenance window. To apply the latest parameter changes to that DB instance, manually reboot the DB instance.

... so the postgres one will require a manual reboot.

PR #1502, opened 2021-12-13
Add new govuk-rds project to create postgres & mysql instances

barrucadu:
See also https://github.com/alphagov/govuk-aws-data/pull/947

---

Add new govuk-rds project to create postgres & mysql instances
The approach I've taken is to look at app-db-admin, app-postgresql,
and app-mysql; inlining the relevant bits of those and the modules
they use; upgrading everything to a new version of Terraform; and
using a map to hold all the database configuration.

Best reviewed commit-by-commit, as there's a lot going on here.

---

[Plan](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/1850/console)
[Trello card](https://trello.com/c/rB7CoPoG/42-terraform-up-all-the-new-rds-instances-and-db-admin-ec2-instances)


kevindew:
If these are not going to have the `blue_` prefix should this instead have a project name prefixed with `infra-` rather than `app-` ? I admit I've never really understood the difference between infra and app but understand the blue/green (without the green) is one of the main distinguishing features.

barrucadu:
I wondered about the `app-` and `infra-` prefixes, as it doesn't seem likely we'll want to blue / green deploy RDS instances - but postgres primary and mysql are `app-` projects

richardTowers:
I think it's more consistent to keep these in an `app-` directory, and keeping them in the `blue` stack (rather than the `govuk` stack). I think that's where the other databases are. Losing the prefixes won't be a problem unless we try to create a `green` stack in the same account, which we've never done before and I doubt we'll ever do.

I think removing content-data-api and transition from this PR is a good idea. It might be nice to import them into this project later on (so all the DBs are configured in the same way), but for now keeping the separate reduces the moving parts.

I haven't had a chance to review the code properly yet. I'll have a close look at the security groups since you said you were less confident of those.

barrucadu:
This PR is now pretty close to something mergeable, and I've updated the plan:

I've made a draft PR in govuk-aws-data with all the database configuration and credentials.  There are unique credentials per environment, but I still need to set the alarm thresholds and environment-specific sizes and instance types.  I've also removed the `content-data-api` and `transition` configuration, as we have separate terraform projects for those.

After I sort out the govuk-aws-data PR I'll come back to this one and see what needs to be done to make this join up nicely with puppet.

barrucadu:
I've dropped the `is_postgres` parameter and changed the database configuration to specify the engine, version, and parameters needed.  This means we're now creating one `aws_db_parameter_group` per database.

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,727 @@
...
+  source_ids  = [aws_db_instance.instance[each.key].id]
+  event_categories = [
+    "availability",
+    "deletion",
+    "failure",
+    "low storage",
+  ]
+}
+
+resource "aws_cloudwatch_metric_alarm" "rds_cpuutilization" {
```
kevindew, line 237: For the cloudwatch alarms I think it'd be good to explain where these came from and what their thresholds are - basically to try help the next person who wonders what the logic was for the setting of alerts like you've experienced.

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,727 @@
...
+resource "aws_db_parameter_group" "engine_params" {
+  for_each = var.databases
+
+  name_prefix = "${each.value.name}-${each.value.engine}-"
+  family      = "${each.value.engine}${each.value.engine_version}"
+
+  dynamic "parameter" {
+    for_each = each.value.engine_params
+
+    content {
```
kevindew, line 164: If I've understood this right, we'd have a hash of engine_params like:

```
{ "log_min_duration_statement": "10000", "log_statement": "all" }
```

which gets mapped into parameter name and value?

If so that'll cause a problem with the additional params such as apply_method: https://github.com/alphagov/govuk-aws/blob/32131365cb04566dd7e11f7034ea0300e78a9f08/terraform/projects/app-postgresql/main.tf#L134

could it be that engine params is an array and the item value is a hash of keys with properties - that could allow it to deal with any new keys too I'd have thought?

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,727 @@
...
+  alarm_description   = "This metric monitors the amount of available storage space."
+
+  dimensions = {
+    DBInstanceIdentifier = aws_db_instance.instance[each.key].id
+  }
+}
+
+# -------------------------------------------------------------------------------
+
+data "aws_ami" "node_ami_ubuntu" {
```
kevindew, line 299: Does our looping mean we don't get to use the node_group definition: https://github.com/alphagov/govuk-aws/blob/32131365cb04566dd7e11f7034ea0300e78a9f08/terraform/modules/aws/node_group/main.tf - just wondering if we needed to inline it as well, since it's quite heavy on config

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,727 @@
...
+    name   = "name"
+    values = ["${var.instance_ami_filter_name}"]
+  }
+
+  filter {
+    name   = "virtualization-type"
+    values = ["hvm"]
+  }
+
+  owners = ["099720109477", "898082745236"]
```
kevindew, line 312: Might be good to say somewhere what these mean - I have no idea myself.

kevindew (commented):
Amazing stuff @barrucadu, thanks for getting this together.

I've perhaps only found the one issue myself. But I would suggest it'd be good to get another review from someone more AWS inclined than myself.

I would say this is really quite tough to review, there's so much going on in the terraform, a lot of things copied and pasted, and I wasn't super clear as to whether we had (or intended to) inline everything from existing modules or just the DB admin stuff?

I'd guess there isn't time (and perhaps not sufficient value) but I'd imagine this would be easier to follow with commits that add bits gradually. For example a commit to add the RDS instances in, a commit for their policies, a commit to add the monitoring in for them, then DB admin instances, etc. That might give more chance to explain in commit messages where values came from as I was finding myself trying to cross reference a lot of "is this new or copied"

terraform/projects/app-govuk-rds/.terraform-version
```diff
@@ -0,0 +1 @@
+1.1.0
```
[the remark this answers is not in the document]
barrucadu: Yeah, it uses the terraform version for the project for formatting

terraform/projects/app-govuk-rds/main.tf
barrucadu: These came from the rds module, I'll add some comments

terraform/projects/app-govuk-rds/main.tf
barrucadu: It looks like `name`, `value`, and `apply_method` are the only arguments a `parameter` block can take - so I'll add `apply_method` and make it default to `immediate` if not specified in the `engine_params`

terraform/projects/app-govuk-rds/main.tf
barrucadu: I did inline the `node_group` module, but could drop much of it because I just used the values that the db-admin project specified.

barrucadu:
@kevindew I'll have a go at splitting this PR up, I agree it's pretty tough to review as it is.

terraform/projects/app-govuk-rds/main.tf
kevindew: ok cool - I don't feel fully in the loop on all the logic of inlining and not inlining, but I don't think I need to be. It'd be good to capture the why though in commits.

barrucadu: I think the reason is that we can't use the modules we currently have without updating them to the new terraform syntax, and also the old version of terraform most of this repo uses doesn't support the `for_each` syntax.  We wouldn't want to create RDS instances with `count`, because we have problems currently with (eg) updating or removing something in the middle of a list affecting later entries.

And if we can't use `count` or `for_each`, we just have to repeat the resource definitions for every copy we want - so this file would be N times as long, where N is the number of different databases.

barrucadu:
@kevindew I've split this up into multiple commits — do your comments about adding comments about what things are or where they came from still apply now that that information is in the commit messages?

terraform/projects/app-govuk-rds/main.tf
richardTowers: I think we _might_ be able to use the old module in the loop (modern terraform has module foreach, not sure if it works with modules using the old syntax though).

Personally, I still prefer things to be inlined - it makes it more obvious to me all the things we're creating in this deployment. Because modules like `node_group` get used for lots of different things (db_admin machines, app machines, rabbitmq machines etc.) the module ends up sprouting dozens of conditionals, and it can be hard to tell what it's actually doing. Inlining the resources means we can keep an eye on what we're actually doing, and we can address things like "should we be creating one load balancer per EC2 instance" which the module might otherwise obfuscate.

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,688 @@
+variable "aws_region" {
```
richardTowers, line 1: This file ended up pretty big. I wonder if we should split it up into:

- `main.tf` - contains only the `terraform` and `provider` blocks
- `variables.tf` - contains all the variables
- `rds.tf` - contains all the stuff that's to do with the RDS instance (maybe including cloudwatch alarms and subscriptions?)
- `db_admin.tf` - contains all the stuff that's to do with DB Admin

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,688 @@
...
+    [{ key = "aws_migration", value = "${replace(each.value.name, "-", "_")}_db_admin", propagate_at_launch = true }],
+    [{ key = "aws_hostname", value = "${each.value.name}-db-admin-1", propagate_at_launch = true }]
+  )
+
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+resource "aws_elb" "node" {
```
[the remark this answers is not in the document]
richardTowers: Or maybe... do we even want a load balancer in the first place?

Looks like we've got two listeners - 22 lets ssh traffic to the instance, 6432 lets pgbouncer traffic through (which we don't need).

Could we just let traffic go straight from the jumpbox to the db-admin instances, and skip the load balancer entirely?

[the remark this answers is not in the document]
richardTowers: I'm probably straying too far from established patterns here. And I'm not an AWS expert by any means.

But I think the objectively simplest thing would be to:

* Create an `aws_instance` (i.e. without all the autoscaling group stuff)
* Put it in a security group
* Change the `db_admin_service_record` Route53 record so it points at the [private_dns](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#private_dns) name for the instance, instead of the ELB
* Make sure there's a security group rule that lets traffic flow from the jumpbox to the db_admin instance.

... I might be oversimplifying though. It might not be worth shaving this yak right now.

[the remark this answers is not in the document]
richardTowers: I've fully talked myself out of this now. Resolving the comment thread. Apologies for the noise.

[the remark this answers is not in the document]
richardTowers: Although still maybe one NLB per app would be better than one Classic ELB per app...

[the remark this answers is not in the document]
richardTowers: ... sorry, still flogging this horse ...

We don't actually use the load balancers for ssh:

```
gds govuk connect -e integration ssh db_admin
There is one machine to connect to

Running command: ssh -J richardtowers@jumpbox.integration.publishing.service.gov.uk richardtowers@ip-10-1-6-102.eu-west-1.compute.internal
```

`ip-10-1-6-102.eu-west-1.compute.internal` is the private DNS name of the EC2 instance, not the load balancer.

So I guess all this load balancer is doing for us is the healthcheck for the autoscaling group 🤔

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,708 @@
...
+  backup_window           = var.backup_window
+  copy_tags_to_snapshot   = true
+  snapshot_identifier     = "${each.value.name}-snapshot"
+  monitoring_interval     = 60
+  monitoring_role_arn     = data.terraform_remote_state.infra_monitoring.outputs.rds_enhanced_monitoring_role_arn
+
+  vpc_security_group_ids = [
+    data.terraform_remote_state.infra_security_groups.outputs.sg_mysql-primary_id,
+    data.terraform_remote_state.infra_security_groups.outputs.sg_postgresql-primary_id,
+  ]
```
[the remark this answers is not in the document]
barrucadu: > Have this terraform create one security group per database, output the IDs, and then have app-ckan and co read the security group ids from this project's remote state and set up rules that say "ckan can talk to ckan db" etc.

Would that work? As in, could we use the `data.terraform_remote_state.app_govuk_rds.output_name["database_name"]` syntax in an older terraform project?

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,688 @@
...
+  triggers = {
+    snippet = file("../../userdata/${local.user_data_snippets[count.index]}")
+  }
+}
+
+resource "aws_launch_configuration" "node" {
+  name_prefix   = "${var.stackname}-govuk-rds-"
+  image_id      = data.aws_ami.node_ami_ubuntu.id
+  instance_type = "t2.medium"
+  user_data     = join("\n\n", ["#!/usr/bin/env bash", join("\n", null_resource.user_data.*.triggers.snippet)])
```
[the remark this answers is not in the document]
barrucadu:
```
michaelswalker@ec2-staging-blue-db_admin-ip-10-12-6-111:~$ which bash
/bin/bash
```

🤷 

That's much better, I'll change it.

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,688 @@
...
+    lb_port           = 22
+    lb_protocol       = "tcp"
+  }
+
+  listener {
+    instance_port     = 6432
+    instance_protocol = "tcp"
+    lb_port           = 6432
+    lb_protocol       = "tcp"
+  }
```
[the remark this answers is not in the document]
barrucadu: Great, I wondered if that port was needed

terraform/projects/app-govuk-rds/main.tf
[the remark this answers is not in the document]
barrucadu: > Create an `aws_instance` (i.e. without all the autoscaling group stuff)

I started by doing that, then went back to the ASG approach as I think that would mean we have to re-run terraform if the instance gets terminated (see message of e5423502ae2b22197c1df355daa9edcd014d9c0b).

If you don't think the ELB is needed, we could drop it entirely.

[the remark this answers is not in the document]
barrucadu: Resolved in slack: the ELB gives us the port 22 healthcheck, so it is needed for the ASG.

terraform/projects/app-govuk-rds/main.tf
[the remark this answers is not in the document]
richardTowers: Yeah, I think if we're on terraform pre-0.12.30 we probably can't access the remote state written by this terraform version (source: https://github.com/hashicorp/terraform/blob/v0.12.30/CHANGELOG.md#01230-january-06-2021)

We could still get the IDs of the security groups by looking them up by name using a [security_group `data` thing](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/security_group) though.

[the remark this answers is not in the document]
barrucadu: Ok, I think my preference would be to keep things as they currently are in the PR, so we can get the new databases up and running sooner rather than later, and then to introduce new security groups to tighten up security later, as we'll have to make changes to our other Terraform projects to do that.

Access to our new databases will be slightly more relaxed than access to our old ones (as we're putting them in both the postgres and mysql groups), but it's still all in a VPC so things aren't wide open, and access is still somewhat restricted.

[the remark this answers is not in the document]
barrucadu: There's now a unique security group for each RDS instance

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,688 @@
...
+}
+
+resource "aws_launch_configuration" "node" {
+  name_prefix   = "${var.stackname}-govuk-rds-"
+  image_id      = data.aws_ami.node_ami_ubuntu.id
+  instance_type = "t2.medium"
+  user_data     = join("\n\n", ["#!/usr/bin/env bash", join("\n", null_resource.user_data.*.triggers.snippet)])
+
+  security_groups = [
+    data.terraform_remote_state.infra_security_groups.outputs.sg_db-admin_id,
```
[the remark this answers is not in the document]
barrucadu: And a unique security group for each db-admin

terraform/projects/app-govuk-rds/main.tf
barrucadu: I'll split this up in one commit at the end, after we're happy with everything else here.

terraform/projects/app-whitehall-backend/main.tf
```diff
@@ -177,6 +177,20 @@ resource "aws_iam_role_policy_attachment" "whitehall_csvs_attach" {
...
+  name = "${var.stackname}_whitehall_rds_access"
+}
+
+resource "aws_security_group_rule" "whitehall-rds_ingress_whitehall-backend_mysql" {
+  type      = "ingress"
+  from_port = 3306
+  to_port   = 3306
+  protocol  = "tcp"
+
+  security_group_id        = "${aws_security_group.signon-rds.id}"
```
richardTowers, line 190: This one looks like an oopsy
```suggestion
  security_group_id        = "${aws_security_group.whitehall-rds.id}"
```

terraform/projects/app-whitehall-frontend/main.tf
```diff
@@ -154,6 +154,20 @@ module "alarms-elb-whitehall-frontend-internal" {
...
+  name = "${var.stackname}_whitehall_rds_access"
+}
+
+resource "aws_security_group_rule" "whitehall-rds_ingress_whitehall-frontend_mysql" {
+  type      = "ingress"
+  from_port = 3306
+  to_port   = 3306
+  protocol  = "tcp"
+
+  security_group_id        = "${aws_security_group.signon-rds.id}"
```
richardTowers, line 167:
```suggestion
  security_group_id        = "${aws_security_group.whitehall-rds.id}"
```

terraform/projects/app-govuk-rds/main.tf
```diff
@@ -0,0 +1,692 @@
...
+
+  name        = "${var.stackname}_${each.value.name}_db-admin_access"
+  vpc_id      = data.terraform_remote_state.infra_vpc.outputs.vpc_id
+  description = "Access to ${each.value.name} db-admin"
+
+  tags = merge(local.tags, { Name = "${var.stackname}_${each.value.name}_db-admin_access" })
+}
+
+resource "aws_security_group_rule" "rds_ingress_db-admin_postgres" {
+  for_each = var.databases
```
richardTowers, line 154: I wonder if we should filter this on engine = "postgres".

Something like:

```suggestion
  for_each = {for name, config in var.databases : name => config if config.engine == "postgres"}
```

... and then we don't need to give all the postgres databases access to the myql ports and vice versa. Maybe that's too fancy - it's not really an issue if they have access to one more port which isn't even open.

barrucadu: No, that's good, I actually wanted to do that but didn't know you could

richardTowers (approved):
To the best of my ability to review it, this looks good. A few of the IAM policies look a little over-broad (like `resource: [*]`), but they're no worse than what we have now and perfection shouldn't stop us from reaching good enough.

There are a couple more unresolved comments (splitting up main.tf, adding some comments on the cloudwatch alarms), but I don't think they need to block approval.

I've run a plan locally against integration and it looks okay (although you never really know until it's applied)

terraform/projects/app-govuk-rds/main.tf
barrucadu: It looks like they're IDs of official Amazon things, and one of them isn't needed here: https://github.com/alphagov/govuk-aws/commit/5e89e967c76844e52c1481a3721e2e124fcefcd9 - I'll remove the deep learning one.

richardTowers (approved):

richardTowers (approved):

terraform/projects/app-govuk-rds/rds.tf
```diff
@@ -130,3 +131,13 @@ resource "aws_route53_record" "database" {
...
+
+resource "aws_route53_record" "database_internal_root_domain_name" {
+  for_each = var.databases
+
+  zone_id = data.terraform_remote_state.infra_root_dns_zones.outputs.internal_root_zone_id
+  name    = "${each.value.name}-${each.value.engine}.${data.terraform_remote_state.infra_root_dns_zones.outputs.internal_root_domain_name}"
+  type    = "CNAME"
+  ttl     = 300
+  records = [aws_route53_record.database_internal_domain_name[each.key].fqdn]
+}
```
richardTowers, line 143: It's probably a bit surprising to have this here, instead of in infra-public-services. But it would be pretty painful to do this there (given the old version of terraform etc.), and it's nice having everything in one place. So I think this is okay.

richardTowers (approved):

PR #1510, opened 2021-12-20
Allow the data science data app to get bigquery credentials

maxf:
The app needs the bigquery credentials in order to compute the "structural network" using user journey data.

### DIFF
--- a/terraform/projects/app-data-science-data/main.tf
+++ b/terraform/projects/app-data-science-data/main.tf
@@ -63,6 +63,10 @@ resource "aws_iam_instance_profile" "data-science-data_instance_profile" {
 
 data "aws_caller_identity" "current" {}
 
+data "aws_secretsmanager_secret" "secret_big_query_service_account_key" {
+  name = "related_links-BIG_QUERY_SERVICE_ACCOUNT_KEY"
+}
+
 data "aws_iam_policy_document" "data-science-data_read_ssm_policy_document" {
   statement {
     actions = [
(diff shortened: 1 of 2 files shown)


terraform/projects/app-data-science-data/main.tf
```diff
@@ -80,6 +80,18 @@ data "aws_iam_policy_document" "data-science-data_read_ssm_policy_document" {
   }
 }
 
+data "aws_iam_policy_document" "read_secrets_from_secrets_manager_policy_document" {
```
richardTowers, line 83: I think you need to refer to this policy document inside an `aws_iam_policy` resource too (this data block literally just creates the text of the policy, it doesn't upload it to the AWS API).

richardTowers: A bit like https://github.com/alphagov/govuk-aws/pull/1510/files#diff-a220c9a1f8e8811c6899b696c984f41224fcdfdc80f1ffd4749ebd6df1b8f6daR112-R115 for the sagemaker policy.

And then you also need to attach your new policy to a role like https://github.com/alphagov/govuk-aws/pull/1510/files#diff-a220c9a1f8e8811c6899b696c984f41224fcdfdc80f1ffd4749ebd6df1b8f6daL110-L113

maxf: ah, yes, thansk. And I nee to supply the actual name of the secret, obvs -- as found in the [related links tf config](https://github.com/alphagov/govuk-aws/blob/give-bigquery-creds-to-data-science-app/terraform/projects/app-related-links/main.tf#L54)

richardTowers (approved):

PR #1511, opened 2021-12-21
Add prefix to avoid clash with app-related-links

maxf:
`read_secrets_from_secrets_manager_policy` was already defined in `app-related-links/main.tf` so we're namespacing the policy and attachment names by prefixing them.

### DIFF
--- a/terraform/projects/app-data-science-data/main.tf
+++ b/terraform/projects/app-data-science-data/main.tf
@@ -84,7 +84,7 @@ data "aws_iam_policy_document" "data-science-data_read_ssm_policy_document" {
   }
 }
 
-data "aws_iam_policy_document" "read_secrets_from_secrets_manager_policy_document" {
+data "aws_iam_policy_document" "data-science-data_read_secrets_from_secrets_manager_policy_document" {
   statement {
     actions = [
       "secretsmanager:GetSecretValue",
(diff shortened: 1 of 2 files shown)


terraform/projects/app-data-science-data/main.tf
```diff
@@ -113,9 +113,9 @@ resource "aws_iam_policy" "data-science-data_read_ssm_policy" {
   policy = "${data.aws_iam_policy_document.data-science-data_read_ssm_policy_document.json}"
 }
 
-resource "aws_iam_policy" "read_secrets_from_secrets_manager_policy" {
-  name   = "read_secrets_from_secrets_manager_policy"
-  policy = "${data.aws_iam_policy_document.read_secrets_from_secrets_manager_policy_document.json}"
+resource "aws_iam_policy" "data-science-data_read_secrets_from_secrets_manager_policy" {
+  name   = "data-science-data_read_secrets_from_secrets_manager_policy"
```
richardTowers, line 117: I think this is the only thing that needs to be unique across the AWS region. The resource names (`"bar` in `resource "foo" "bar"`) only have to be unique to the terraform module. No great harm in having them like this though, other than more characters to type.

richardTowers (approved):

terraform/projects/app-data-science-data/main.tf
maxf: yes, the names are a bit long here, but I think it's good practice generally. Thanks.

PR #1514, opened 2022-01-06
Add scale learn to rank ASG policy to Jenkins

theseanything:
This give Jenkins IAM permissions to scale the Search API's learn to rank autoscaling group. This is need for the Jenkins job to train and deploy a new learn to rank model.


bilbof (approved):
LGTM

PR #1517, opened 2022-01-07
Add separate variables for split postgres & msyql databases

barrucadu:
The backend machine class runs apps with each type of database, but we
have only created the postgres instances in production and staging
currently.

As expected, this is a no-op on integration with the relevant govuk-aws-data branch:
https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/2008/console


kevindew (approved):
This looks good to me. I assume we'll plan to remove these variables once this work is complete? do we have that recorded anywhere?

barrucadu:
Yes, we can remove these variables when everything is upgraded.  I'll add a card to the board.

PR #1518, opened 2022-01-07
Add Jenkins ssh ingress SG rule for LTR ASG instances

theseanything:
This rule enable SSH ingress from the Jenkins (deploy) instance. This is needed as the learn to rank (LTR) Jenkins job needs SSH access into new instances.


bilbof (approved):
LGTM!

PR #1519, opened 2022-01-10
Use Ubuntu 20.04 for LTR ASG instances

theseanything:
This updates the AMI id to use Ubuntu 20.04 instead of Ubuntu 18.04 for instances used for Search API's Learn to rank process. Using a newer version of Ubuntu as it comes pre-packaged with Python 3.8 which is required by the train and deployment scripts.


terraform/projects/app-search/main.tf
```diff
@@ -460,21 +460,9 @@ resource "aws_key_pair" "learntorank-generation-key" {
...
-  filter {
-    name   = "name"
-    values = ["ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-*"]
-  }
-}
-
 resource "aws_launch_template" "learntorank-generation" {
   name          = "govuk-${var.aws_environment}-search-ltr-generation"
-  image_id      = "${data.aws_ami.ubuntu_bionic.id}"
+  image_id      = "ami-08ca3fed11864d6bb"
```
bilbof, line 465: LGTM, though it might be nice to put in a comment to say which AMI this is, e.g. `# Ubuntu Focal 20.04 (20211129), amd64, hvm-ssd, AMI for eu-west-1`.

The benefits to using `data.aws_ami` here is that this info is documented outside of the comment, and we get a valid AMI regardless of the region we are deploying to. At the moment this AMI would only be valid for the eu-west-1 region. Additionally, I think this would enable AWS to patch the image for us automatically (giving us a new AMI for our query) rather than requiring us to update the AMI manually. If you do a search for `focal 20220104 eu-west` on https://cloud-images.ubuntu.com/locator/ you can see there is a different AMI for each region (and duplicate AMIs for the same version, since _I believe_ AWS has issued a patched AMI for our region).

bilbof (commented):
Nice one 👍 I'd recommend keeping the `data.aws_ami` block though, as (TIL) it seems to have some benefits

terraform/projects/app-search/main.tf
theseanything: Ah good shout, using the AMI ID directly was a lack of my understanding about `data.aws_ami` - definitely makes sense to use it. I've updated the commit.

bilbof (approved):
Great one 👍

PR #1525, opened 2022-01-20
Remove unused var from infra-security

bilbof:
This var is not used - the list tfvar provided by aws-data is now empty - see https://github.com/alphagov/govuk-aws-data/pull/965.


fredericfran-gds (approved):
looks good, thanks for clearing up.

PR #1530, opened 2022-01-27
Set a count of 0 on mysql-primary resources

ChrisBAshton:
These are no longer required, so we'll delete them.
Once these Terraform changes have been applied, we can then delete
the mysql.tf file.

Trello: https://trello.com/c/VWpD8SvY/55-remove-mysql-primary-and-mysql-standby-infrastructure-and-configuration


rtrinque (approved):
Looks good

PR #1531, opened 2022-01-28
Bucket for Accounts backups prior to Digital Identity migration

kevindew:
Trello: https://trello.com/c/nvAK8W8x/105-move-govuk-accounts-pre-digital-identity-out-of-the-database-backups-bucket

This bucket has been creating for storing databases that were used on the
GOV.UK accounts prototype. This service has now been superseded by DI.
We however made a commitment to storing our backups for 2 years and it
contains data (the audit log in the security_activities table) which wasn't
migrated to digital identity.

This has been created so that the files can be moved from
govuk-<environment>-database-backups which will allow us to remove some
complex life-cycle rules that have been introduced just for this
scenario.

The files for this only exist in the production environment, therefore I
have only created this bucket in production.


barrucadu (approved):
Looks good to me

PR #1533, opened 2022-01-28
Create a bucket for Graphite Whisper backups

kevindew:
Trello: https://trello.com/c/t2jebInp/106-store-whisper-backups-in-its-own-s3-bucket

This bucket has been created so that the Graphite whisper backups can be
moved out of govuk-<environment>-database-backups. I chose to name this
bucket after graphite rather than whisper as that seemed easier for
someone to understand.

The reason for moving them is so that we can make the
govuk-<environment>-database-backups have a consistent purpose and a
consistent retention period. Their presence there convoluted
the purpose of that bucket as these whisper backups don't have the same
concerns: they're not shared cross-accounts; they don't contain data
we'd use as part of a disaster recovery, so we're less precious; they use
masses of space, so we want a different retention policy.

By moving to a new bucket I decided we could forgo multi-region
replication and versioning as the data just doesn't seem particularly
high value.

This change also grants access to the graphite machine to talk to this bucket.

Once these changes are applied and puppet has been changed to write to
this bucket I intend to remove the graphite configuration in the
infra-database-backups-bucket.

Example plan: https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/2219/console


sengi (approved):
LGTM - good idea!

PR #1535, opened 2022-01-28
Simplify lifecycle rules for database backups

ollietreend:
This PR simplifies the lifecycle rules on the S3 buckets which hold database backups created by env_sync.

Previously, various retention rules were needed on different paths within the bucket. This reflected the fact that the bucket was being used to store lots of different types of backup.

We made the decision to simplify the purpose of this bucket so that it's only used to store env_sync database dumps.

We now have one lifecycle rule which is applied to the entire bucket.

Given the current configuration, this means:
- Production backups will be permanently deleted after 121 days ('soft deleted' after 120 days and 'hard deleted' the next day)
- Staging & integration backups will be permanently deleted after 4 days ('soft deleted' after 3 days)

This database backups bucket (and its replica bucket) have versioning enabled. We're dubious of the value this provides – it seems to complicate the lifecycle rules, and only protects against the unlikely case of accidental deletion. However after much deliberation, we decided not to turn it off (that was one 'yak shave' too far).

We also recognise that the lifecycle rules are duplicated between buckets, and so could probably be DRY'd up. Again, this was one 'yak shave' too far. Our intention is simply to leave this configuration in a better state than it was before.

### Related PR

This PR should be merged in conjunction with alphagov/govuk-aws-data#972

---

Trello: https://trello.com/c/zrmnbNjN

### DIFF
--- a/terraform/projects/infra-database-backups-bucket/main.tf
+++ b/terraform/projects/infra-database-backups-bucket/main.tf
@@ -40,36 +40,6 @@ variable "remote_state_infra_monitoring_key_stack" {
   default     = ""
 }
 
-variable "standard_s3_storage_time" {
-  type        = "string"
-  description = "Storage time in days for Standard S3 Bucket Objects"
-  default     = "30"
-}
-
-variable "glacier_storage_time" {
-  type        = "string"
-  description = "Storage time in days for Glacier Objects"
-  default     = "90"
-}
-
-variable "expiration_time" {
-  type        = "string"
-  description = "Expiration time in days of S3 Objects"
-  default     = "120"
-}
-
-variable "expiration_time_whisper_mongo" {
-  type        = "string"
-  description = "Expiration time in days for Whisper/Mongo S3 database backups"
-  default     = "7"
-}
-
-variable "replication_setting" {
-  type        = "string"
-  description = "Whether replication is Enabled or Disabled"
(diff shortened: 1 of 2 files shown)


kevindew:
Looks great to me, do we need a pairing PR in govuk-aws-data which removes the unnecessary attributes (assuming there are some)

kevindew (approved):
Looks good to me, will try out on integration before merging.

terraform/projects/infra-database-backups-bucket/main.tf
```diff
@@ -111,200 +81,139 @@ resource "aws_s3_bucket" "database_backups" {
...
-    id      = "govuk_accounts_pre_digital_identity_lifecycle_rule"
-    prefix  = "govuk-accounts-pre-digital-identity/"
-    enabled = true
+    # Use a long retention period in production
+    id      = "long_retention_period"
+    enabled = "${var.aws_environment == "production"}"
 
+    # Ideally everything would go in the Standard (Infrequent Access) storage class when created.
+    # But newly created objects always go into Standard, and can only move into IA after at least 30 days.
+    # https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-transition-general-considerations.html
```
kevindew, line 178: Shall we remove the explanatory comments from the copied rule?

PR #1536, opened 2022-01-28
Enable ELB health checks on calc_frontend instances

bilbof:
This will mean that both EC2 and ELB checks are used to determine whether an instance of a calculators_frontend is healthy or should be removed from an ASG and terminated.

The result is that an app failing a /healthcheck/ready check on an instance will lead to the instance being removed from an AutoScalingGroup.

As far as I can see until now we have needed to terminate instances manually when application level checks had failed.

I'm starting with calculators_frontend and will roll this out to other instances in future.

Docs: https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-add-elb-healthcheck.html


fredericfran-gds (approved):
Looks good. We would be wary of adding that to all other node classes because some health checks are not passing because of misconfiguration but load balancers will keep forwarding traffic to all nodes if all nodes are deemed unhealthy.

bilbof:
Thanks @fredericfran-gds. I did some tests with `frontend` and `calculators_frontend` ASGs and hit a couple of issues:

* `frontend`: Launching a new EC2 instance failed because “One or more target groups not found. Validating load balancer configuration failed.”
* `calculators_frontend`: Instance can be started but does not pass health checks due to a dependency issue.

Rather than resolve these and continue expanding, I'm going to accept that folks will need to manually terminate instances until we roll out kubernetes to these workloads.

Closing this as it's a better use of time to focus on Kubernetes, but if others feel like taking it on they should! :)

PR #1541, opened 2022-02-03
Drop unused govuk-mongodb-backup-s3 buckets

ollietreend:
The `govuk-mongodb-backup-s3-*` buckets have long been unused, and so should be safe to remove.

This also removes the module `s3_bucket_lifecycle` which is only used by this project, so should also be safe to remove.


kevindew:
Looks good - thanks Olly. I'll apply it (it's a bit of a battle)

kevindew (approved):

PR #1542, opened 2022-02-04
Permamently delete items in aws-logging buckets

ollietreend:
The `govuk-{env}-aws-logging` buckets have a 30 day expiration policy, but also had versioning enabled. This means that files are only soft deleted after 30 days, and continue to exist in the background as old versions taking up storage space.

The straightforward approach would be to disable versioning on the bucket so that files are permanently deleted when they expire. However versioning needs to remain enabled on the bucket because there is a replica configured.

Instead, this change configures a non-current version expiry so that old versions of files are permanently deleted after 1 day.

In real terms, this means that files will be permanently deleted from the bucket after 31 days (expiring at 30 days, and being deleted 1 day later).

### DIFF
--- a/terraform/projects/infra-monitoring/main.tf
+++ b/terraform/projects/infra-monitoring/main.tf
@@ -113,16 +113,22 @@ resource "aws_s3_bucket" "aws-logging" {
     Environment = "${var.aws_environment}"
   }
 
-  # Expire everything after 30 days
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Nice catch

kevindew (approved):
Thanks for fixing the replication snag.

PR #1544, opened 2022-02-07
Create Postgres RDS instances for Locations API

MuriloDalRi:
https://trello.com/c/oINLjnKv/2834-create-postgres-rds-instances-for-locations-api-5

### DIFF
--- a/terraform/projects/app-locations-api-postgresql/README.md
+++ b/terraform/projects/app-locations-api-postgresql/README.md
@@ -1,0 +1,54 @@
+## Project: projects/app-locations-api-postgresql
+
+RDS PostgreSQL instance for the Locations API
+
+## Requirements
+
+| Name | Version |
+|------|---------|
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | = 0.11.14 |
+| <a name="requirement_aws"></a> [aws](#requirement\_aws) | 2.46.0 |
+
+## Providers
+
+| Name | Version |
+|------|---------|
+| <a name="provider_aws"></a> [aws](#provider\_aws) | 2.46.0 |
+
+## Modules
+
+| Name | Source | Version |
+|------|--------|---------|
+| <a name="module_alarms-rds-locations-api-postgresql-primary"></a> [alarms-rds-locations-api-postgresql-primary](#module\_alarms-rds-locations-api-postgresql-primary) | ../../modules/aws/alarms/rds | n/a |
+| <a name="module_locations-api-postgresql-primary_rds_instance"></a> [locations-api-postgresql-primary\_rds\_instance](#module\_locations-api-postgresql-primary\_rds\_instance) | ../../modules/aws/rds_instance | n/a |
+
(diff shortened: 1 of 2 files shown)


ChrisBAshton (commented):
Sorry for catching this so late, but the creation of `app-locations-api-postgresql` should not be necessary. We can piggy-back off of the [app-govuk-rds](https://github.com/alphagov/govuk-aws/blob/c307dac36c0f1e0d3a241941f312762f4e0e69a6/terraform/projects/app-govuk-rds/rds.tf) which defines all the other RDS instances.

Taking Local Links Manager as inspiration, I believe you only need to add a block to integration/staging/production `common.tfvars`: [example](https://github.com/alphagov/govuk-aws-data/blob/e61abc79e45934d5baabd6c2b8e37cced8bb3c02/data/app-govuk-rds/integration/common.tfvars#L172-L189). I see you've already done that here: https://github.com/alphagov/govuk-aws-data/pull/978/files#diff-211877673544fb9919003bce211ba8c78bc21742d6557687e5d5246f709a6cccR191-R208

Then to make sure the `locations_api` node can access the RDS, you'd need to make changes to app-locations-api (like [Local Links Manager does to `backend`](https://github.com/alphagov/govuk-aws/blob/7fad0ad1298c92c58eb6f668090bf8a5cd6d9734/terraform/projects/app-backend/main.tf#L269-L281)). I guess that means it makes most sense to do https://trello.com/c/1d58bNhV/2853-create-locationsapi-nodes first!

Hope that makes sense - give me a shout if not.

MuriloDalRi:
> Sorry for catching this so late, but the creation of `app-locations-api-postgresql` should not be necessary. We can piggy-back off of the [app-govuk-rds](https://github.com/alphagov/govuk-aws/blob/c307dac36c0f1e0d3a241941f312762f4e0e69a6/terraform/projects/app-govuk-rds/rds.tf) which defines all the other RDS instances.
> 
> Taking Local Links Manager as inspiration, I believe you only need to add a block to integration/staging/production `common.tfvars`: [example](https://github.com/alphagov/govuk-aws-data/blob/e61abc79e45934d5baabd6c2b8e37cced8bb3c02/data/app-govuk-rds/integration/common.tfvars#L172-L189). I see you've already done that here: https://github.com/alphagov/govuk-aws-data/pull/978/files#diff-211877673544fb9919003bce211ba8c78bc21742d6557687e5d5246f709a6cccR191-R208
> 
> Then to make sure the `locations_api` node can access the RDS, you'd need to make changes to app-locations-api (like [Local Links Manager does to `backend`](https://github.com/alphagov/govuk-aws/blob/7fad0ad1298c92c58eb6f668090bf8a5cd6d9734/terraform/projects/app-backend/main.tf#L269-L281)). I guess that means it makes most sense to do https://trello.com/c/1d58bNhV/2853-create-locationsapi-nodes first!
> 
> Hope that makes sense - give me a shout if not.

Good, I'll close this PR then!

PR #1545, opened 2022-02-08
Add Locations API node

MuriloDalRi:
https://trello.com/c/1d58bNhV/2853-create-locationsapi-nodes


terraform/projects/app-locations-api/main.tf
```diff
@@ -0,0 +1,178 @@
...
+  type    = "A"
+
+  alias {
+    name                   = "${module.locations_api_internal_alb.lb_dns_name}"
+    zone_id                = "${module.locations_api_internal_alb.lb_zone_id}"
+    evaluate_target_health = true
+  }
+}
+
+data "aws_security_group" "locations-api-rds" {
```
fredericfran-gds, line 148: I wonder the security groups/rules should not be added there: https://github.com/alphagov/govuk-aws/tree/master/terraform/projects/infra-security-groups

terraform/projects/app-locations-api/main.tf
```diff
@@ -0,0 +1,178 @@
...
+  internal                         = true
+  vpc_id                           = "${data.terraform_remote_state.infra_vpc.vpc_id}"
+  access_logs_bucket_name          = "${data.terraform_remote_state.infra_monitoring.aws_logging_bucket_id}"
+  access_logs_bucket_prefix        = "elb/${var.stackname}-locations-api-internal-alb"
+  listener_certificate_domain_name = "${var.elb_internal_certname}"
+  listener_action                  = "${map("HTTPS:443", "HTTP:80")}"
+  subnets                          = ["${data.terraform_remote_state.infra_networking.private_subnet_ids}"]
+
+  #TODO: create new security group with alb name
+  security_groups = ["${data.terraform_remote_state.infra_security_groups.sg_locations_api_elb_internal_id}"]
```
fredericfran-gds, line 114: should you add these security groups to this PR?

fredericfran-gds:
Thanks for the terraform :-). See inline comments and some generic comments is:
1. add a proper commit/pr description
2. do `terraform fmt`

terraform/projects/app-locations-api/main.tf
MuriloDalRi: I've added that, does it look ok to you?

terraform/projects/app-locations-api/main.tf
MuriloDalRi: Removed that line

terraform/projects/app-locations-api/main.tf
```diff
@@ -0,0 +1,176 @@
...
+
+variable "internal_domain_name" {
+  type        = "string"
+  description = "The domain name of the internal DNS records, it could be different from the zone name"
+}
+
+variable "instance_type" {
+  type        = "string"
+  description = "Instance type used for EC2 resources"
+  default     = "m5.2xlarge"
```
ChrisBAshton, line 58: Any reason why this is `m5.2xlarge`? Looking at [what other apps have set](https://github.com/alphagov/govuk-aws/search?q=%22m5.%22&type=code), they mostly seem to be `m5.large`, which is the [smallest & cheapest instance size available](https://aws.amazon.com/ec2/instance-types/). I suspect we'll want to do the same, given the heavily cached Locations API should need minimal resources.

terraform/projects/app-locations-api/main.tf
```diff
@@ -0,0 +1,176 @@
...
+  domain   = "${var.elb_internal_certname}"
+  statuses = ["ISSUED"]
+}
+
+module "locations-api" {
+  source                        = "../../modules/aws/node_group"
+  name                          = "${var.stackname}-locations-api"
+  default_tags                  = "${map("Project", var.stackname, "aws_stackname", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "locations_api", "aws_hostname", "locations-api-1")}"
+  instance_subnet_ids           = "${data.terraform_remote_state.infra_networking.private_subnet_ids}"
+  instance_security_group_ids   = ["${data.terraform_remote_state.infra_security_groups.sg_locations_api_id}", "${data.terraform_remote_state.infra_security_groups.sg_management_id}", "${data.terraform_remote_state.infra_security_groups.sg_aws-vpn_id}"]
```
ChrisBAshton, line 88: Looking at other examples, I think this needs to use the hyphenated name. I don't think we need the `sg_aws-vpn_id` security group either, as that only seems to be used on the `backend` machine.

```suggestion
  instance_security_group_ids   = ["${data.terraform_remote_state.infra_security_groups.sg_locations-api_id}", "${data.terraform_remote_state.infra_security_groups.sg_management_id}"]
```

terraform/projects/app-locations-api/main.tf
```diff
@@ -0,0 +1,176 @@
...
+  instance_type                 = "${var.instance_type}"
+  instance_additional_user_data = "${join("\n", null_resource.user_data.*.triggers.snippet)}"
+  instance_elb_ids_length       = 0
+  instance_elb_ids              = []
+  instance_ami_filter_name      = "${var.instance_ami_filter_name}"
+  asg_max_size                  = "${var.asg_size}"
+  asg_min_size                  = "${var.asg_size}"
+  asg_desired_capacity          = "${var.asg_size}"
+  asg_notification_topic_arn    = "${data.terraform_remote_state.infra_monitoring.sns_topic_autoscaling_group_events_arn}"
+  root_block_device_volume_size = "60"
```
ChrisBAshton, line 98: We seem to use a range of values throughout govuk-aws, but Publishing API and Email Alert API both use `50`. Mapit, which this is replacing, only uses 20, as does Router, but neither of these are Rails apps, so I wonder if we historically need the bigger volumes for our Rails apps. To be on the safe side I'd suggest setting this to 50 rather than anything lower.

terraform/projects/app-locations-api/main.tf
```diff
@@ -0,0 +1,176 @@
...
+module "locations-api" {
+  source                        = "../../modules/aws/node_group"
+  name                          = "${var.stackname}-locations-api"
+  default_tags                  = "${map("Project", var.stackname, "aws_stackname", var.stackname, "aws_environment", var.aws_environment, "aws_migration", "locations_api", "aws_hostname", "locations-api-1")}"
+  instance_subnet_ids           = "${data.terraform_remote_state.infra_networking.private_subnet_ids}"
+  instance_security_group_ids   = ["${data.terraform_remote_state.infra_security_groups.sg_locations_api_id}", "${data.terraform_remote_state.infra_security_groups.sg_management_id}", "${data.terraform_remote_state.infra_security_groups.sg_aws-vpn_id}"]
+  instance_type                 = "${var.instance_type}"
+  instance_additional_user_data = "${join("\n", null_resource.user_data.*.triggers.snippet)}"
+  instance_elb_ids_length       = 0
+  instance_elb_ids              = []
```
ChrisBAshton, line 92: Hmm, [email-alert-api sets this to 1](https://github.com/alphagov/govuk-aws/blob/7fad0ad1298c92c58eb6f668090bf8a5cd6d9734/terraform/projects/app-email-alert-api/main.tf#L134-L135) and Publishing API [ranges from 1-2](https://github.com/alphagov/govuk-aws/blob/7fad0ad1298c92c58eb6f668090bf8a5cd6d9734/terraform/projects/app-publishing-api/main.tf#L229-L230) and, as you've taken inspiration from here, [backend sets it to 0](https://github.com/alphagov/govuk-aws/blob/7fad0ad1298c92c58eb6f668090bf8a5cd6d9734/terraform/projects/app-backend/main.tf#L109-L110).

Presumably it comes down to whether the node should use an ALB or an ELB. Based on the [Account API config](https://github.com/alphagov/govuk-aws/commit/82c9dcf61b8c1c8e55e58b81b90f08107ef1f8e5), which is the most recent app we've built, we're going the ELB route, meaning this should be set to `1` and you can remove all the `locations_api_internal_alb` stuff.

That said, I'd appreciate an SRE pair of eyes on this!

ChrisBAshton (commented):
Thanks @MuriloDalRi - I've given this an outsider's review, but someone with more Terraform experience should take a look too.

terraform/projects/app-locations-api/main.tf
sengi: 50 sounds fine to me. Logs only get rotated once a day, which is why some of these were raised to from 20 to 50 (after we had some bugs which caused logspew).

terraform/projects/app-locations-api/main.tf
sengi: `instance_elb_ids` and `instance_elb_ids_length` default to `[]` and `0` respectively, so there's no need to specify them here. They are indeed only for Classic ELB and therefore obsolete; see [the module docs for `aws/node_group`](https://github.com/alphagov/govuk-aws/tree/master/terraform/modules/aws/node_group).

sengi: Oh, definitely don't use a Classic ELB for anything new. Use an ALB.

ChrisBAshton: Thanks for the info, @sengi!

terraform/projects/infra-security-groups/locations-api.tf
```diff
@@ -0,0 +1,135 @@
...
+resource "aws_security_group_rule" "locations-api-elb-internal_egress_any_any" {
+  type              = "egress"
+  from_port         = 0
+  to_port           = 0
+  protocol          = "-1"
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = "${aws_security_group.locations-api_elb_internal.id}"
+}
+
+resource "aws_security_group" "locations-api_elb_external" {
```
sengi, line 86: Is there an external load balancer for Locations API? This SG looks redundant.

sengi (approved):

terraform/projects/infra-security-groups/locations-api.tf
MuriloDalRi: I'll remove that

ChrisBAshton:
Superseded by #1553.

PR #1547, opened 2022-02-14
Upgrade Terraform version for infra-security module.

sengi:
Someone had run `terraform apply` with a much newer version of Terraform, which upgraded the state file in S3, which effectively forces us to upgrade:

```
Error refreshing state: state snapshot was created by Terraform v1.1.5,
which is newer than current v0.13.6; upgrade to Terraform v1.1.5 or
greater to work with this state
```

So I guess we'll upgrade then. This takes us up to the current version of Terraform and allows a newer (compatible) version of the AWS provider. Seemed safer to update the provider rather than continue to pin the same ancient minor release.

Also, since we're running this from Jenkins and potentially people's workstations, add the provider version lock file (with signatures for amd64 and arm64 Mac binaries and amd64 Linux binaries) to reduce the chance of Jenkins encountering version issues. Generated with `terraform providers lock -platform=darwin_amd64 -platform darwin_arm64 -platform=linux_amd64`

Tested: plans locally as expected


richardTowers (approved):
Looks good to me - the update from 0.13 is nowhere near as involved as from pre-0.12 is it.

Sorry that you had to shave this yak.

I guess at some point we should prioritise a set of govuk-aws terraform deployments which we should upgrade terraform in. Even with replatforming, some of them are likely to live for a long time.

sengi:
> I guess at some point we should prioritise a set of govuk-aws terraform deployments which we should upgrade terraform in. Even with replatforming, some of them are likely to live for a long time.

Yeah, I'd put good money on at least a handful of these modules still being here in a year's time 🙈 The upgrade was surprisingly painless (so far), mostly just our dodgy lint checks getting in the way 🦬🦬🦬

PR #1550, opened 2022-02-24
Add Shield Response Team IAM role

richardTowers:
As part of the AWS Shield Advanced offering, there is a "Shield Response
Team" who can help out during an incident.

Unlike ordinary AWS support, this team can be given permission to access
certain APIs to help out directly during the incident (rather than just
providing advice).

To make it easier to enable this during an incident, this commit adds a
role which includes the recommended permissions. This can be selected
from the AWS Console, under the option "Choose an existing role for the
SRT to access my accounts."

We'll document how to enable this in the developer docs.

Note that most of the access this role gives is List / Describe / Get
level. The more highly privileged actions are all on shield / waf / elb
/ cloudfront resources.


sihugh (approved):
🛡️

PR #1551, opened 2022-02-24
Remove duplicate bucket parameter

richardTowers:
This is set again two lines further down. It looks like the second time
wins, because the bucket in integration is called
govuk-integration-aws-waf-logs.


sihugh (approved):
I saw that yesterday, thanks for fixing 👍

PR #1554, opened 2022-03-03
Add provisioning information for a dedicated router documentdb cluster

KludgeKML:
##WHAT

Adds provisioning information for a DocumentDB cluster called router-documentdb-<environment> 

Creates a 3-instance DocumentDB Engine 3.6.0 cluster. with security groups that allow access from the router backend machines and the db_admin machines.

##WHY

We need to move Router away from Mongo 2.6 to at least 3.6 - more details in this card: 

Since router is an important layer, it seems reasonable that it would have a dedicated DocumentDB cluster rather than the shared cluster that was originally set up for Asset manager but is used by a number of apps now.

The actual move will require temporarily halting router-api (not the router itself), dumping from the old cluster and importing into the new cluster, pointing router-api at the new cluster, then when we are happy, pointing router at the new cluster. But before the move, clusters will need to be provisioned in the appropriate environments.


KludgeKML:
Closing as PG solution preferred.

PR #1555, opened 2022-03-07
Add provisioning information for a dedicated Imminence DocumentDB Cluster

KludgeKML:
## WHAT

Adds provisioning information for a DocumentDB cluster called imminence-documentdb-\<environment\> 

Creates a 3-instance DocumentDB Engine 3.6.0 cluster. with security groups that allow access from the imminence backend machines and the db_admin machines.

## WHY

We need to move Imminence away from Mongo 2.6 to at least 3.6 - more details in this card: 

As per [https://github.com/alphagov/govuk-rfcs/blob/main/rfc-143-split-database-instances.md](RFC 143) we should make this a dedicated cluster for this app, rather than adding it to the existing shared documentdb instance.

The actual move will require temporarily halting imminence (or putting it in read-only mode), dumping from the old cluster and importing into the new cluster, then pointing imminence to the new cluster. But before the move, clusters will need to be provisioned in the appropriate environments.

## ETC

This is a pre-requisite for https://github.com/alphagov/govuk-puppet/pull/11573


fredericfran-gds (approved):
Thanks. Looks good. I don't know if you want to try version 4.0 rather than 3.6 of documentdb. Thank you.

KludgeKML:
Eventually it would make sense to go to 4.0, but for the moment I know the 2.6 -> 3.6 jump seems to work okay without any modification on the client, so sticking to that for now.

fredericfran-gds:
> Eventually it would make sense to go to 4.0, but for the moment I know the 2.6 -> 3.6 jump seems to work okay without any modification on the client, so sticking to that for now.

Cool, now worries. it is just that for the previous big db migration last December, we tried to go to the latest version.

KludgeKML:
Closing this because sadly we can't use DocumentDB for Imminence (it requires 2d index support, which DocumentDB does not provide).

PR #1557, opened 2022-03-09
Knowledge graph: remove python setup code

maxf:
The Knowledge Graph instance (that we start every morning) used to do some 
data processing (like some elegibility experiments, or the brexit spreadsheet)
using Python3.7. But that version of python is no longer available on the old neo4j 
AMI we use. So we removed all python code and moved some of it to the Data Science
Data app where it belongs anyway, since that app computes most of the Knowledge Graph data already.

Since no python code is being run on the KG instance, we're removing all the python installation commands, to avoid seeing errors and keeping unnecessary code in there.

### DIFF
--- a/terraform/projects/app-knowledge-graph/userdata.tpl
+++ b/terraform/projects/app-knowledge-graph/userdata.tpl
@@ -11,18 +11,9 @@ aws elb register-instances-with-load-balancer --load-balancer-name ${elb_name} -
 
 sudo locale-gen en_GB.UTF-8
 
-# Upgrade pip
-pip install --upgrade pip
-
-# Install pip and dependencies used in this start-up script
 sudo apt install -y python-pip
 pip install awscli==1.19.112
-pip install futures
 
-# Install Python 3.7
-sudo add-apt-repository ppa:deadsnakes/ppa -y && sudo apt-get update
-sudo apt-get install -y python3.7
-sudo apt install -y python3-pip
 sudo apt install -y libpq-dev
 sudo apt install -y libicu-dev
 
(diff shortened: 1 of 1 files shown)


rory-hurley-gds (changes requested):
comment out lines 19 and 20 too?

terraform/projects/app-knowledge-graph/userdata.tpl
```diff
@@ -11,18 +11,10 @@ aws elb register-instances-with-load-balancer --load-balancer-name ${elb_name} -
 
 sudo locale-gen en_GB.UTF-8
 
-# Upgrade pip
 pip install --upgrade pip
-
-# Install pip and dependencies used in this start-up script
 sudo apt install -y python-pip
```
exfalsoquodlibet, line 15: I think you do not want line 14-15
```pip install --upgrade pip```
as you are getting pip via APT now (line 17-18)
```sudo apt install -y python-pip```

File: terraform/projects/app-knowledge-graph/userdata.tpl [hunk 2]
```diff
@@ -11,18 +11,9 @@ aws elb register-instances-with-load-balancer --load-balancer-name ${elb_name} -
 
 sudo locale-gen en_GB.UTF-8
 
-# Upgrade pip
-pip install --upgrade pip
-
-# Install pip and dependencies used in this start-up script
 sudo apt install -y python-pip
```
exfalsoquodlibet, line 14: Final thought, you may need 
```python3-pip``
in Linux/Ubuntu

https://stackoverflow.com/a/57266972

maxf: I wouldn't do that at this point. The current version is indeed 2.7 but that's ok since `awscli` is pinned to a specific version. It's currently working so I'd rather not play with it, until we need to change the `awcli` version.

exfalsoquodlibet (approved):
Thanks for the changes, Max. Looks good to me.

mammykins:
Prudent changes, thanks for tidying up!

exfalsoquodlibet:
Why is this still blocked @maxf ? Do you need anything else from me / other reviewers?

maxf:
I lost track of this one, sorry. It still needs @rory-hurley-gds to approve though

rory-hurley-gds (approved):
approved

rory-hurley-gds:
Apologies - reviewed this before, but hasnt approved. should be done now

PR #1559, opened 2022-03-15
Add rate limit to AWS WAF # 2869

BhupsAL1:
Added new rule - (Rate Limit) to AWS WAF (count only)

The rate limit rule uses a managed rate limiter provided by AWS to limit the requests per user IP to 2000 per 5 minute period. This rule has the action 'count'. This means we will log the number of requests which would have been blocked, but this rule will not block any requests. Should we see elevated requests from a small set of IPs we should be able to easily set this rule to block {}.

This rule makes use of the Forwarded IP feature new to AWS WAF, it enables us to use the True-Client-IP set in our VCL, which is a header with a value of the real user's client IP.

Links

Forwarded IP: https://docs.aws.amazon.com/waf/latest/developerguide/waf-rule-statement-forwarded-ip-address.html
RateBasedStatement: https://docs.aws.amazon.com/waf/latest/APIReference/API_RateBasedStatement.html

True-Client-IP: https://github.com/alphagov/govuk-cdn-config/blob/4a7e2bcf6fd64217631a4ef4e59877b3fe8535c9/vcl_templates/www.vcl.erb#L311

https://trello.com/c/IvZrVCL3/2869-setting-rate-limits-within-aws-waf # 2869


BhupsAL1:
The first failure is indicating that the README documentation is not up to date.

"The documentation isn't up to date. You should run ./tools/update-docs.sh and commit the results.
Error: Process completed with exit code 1.

rtrinque (approved):
Looks good to me

BhupsAL1:
PR is ready for review

PR #1567, opened 2022-03-28
Specify time zone for daily govgraph generation

maxf:
By default the time zone is GMT so during summer time the govgraph is
generated one hour later (09:30) and isn't available to users early enough.
So adding the Europe/London IANA time zone parameter, as documented at
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_schedule\#argument-reference


exfalsoquodlibet (approved):
I can't access the linked page (Page Not Found) but the changes look reasonable and correctly impact both apps that are responsible for the two building steps of the KG.

PR #1572, opened 2022-04-04
Remove obsolete public facing load balancers

rtrinque:
During the migration from Carrenza to AWS some communication was needed between pat of the infrastructure in AWS and the part that was still running in Carrenza. Hence the need for public facing load balancers for services that would normally not require them.
Now that nothing is left running in Carrenza we can get rid of this vestigial infrastructure.
See also https://github.com/alphagov/govuk-aws-data/pull/1010


BhupsAL1 (approved):
LGTM

PR #1578, opened 2022-04-26
Move from Concourse to Jenkins for govuk-repo-mirror job

ChrisBAshton:
Concourse is no longer supported, so we're moving the repo mirroring job to Jenkins. We can reuse much of the Terraform though.

The previous Terraform was overcomplicated, handling Carrenza environments and creating multiple users. We only need one user and one role, and we only need to grant `AWSCodeCommitPowerUser` permission to the role.

There was some discussion (with Roch) as to whether this should continue to exist in the `tools` AWS environment, which was introduced [about 2.5 years ago](https://github.com/alphagov/govuk-aws/commit/608cefb10b6a58fadd42aa4df588d8d2efb224a0) as a place for 'meta' production services. But the `tools` environment is not documented anywhere and thus doesn't give much reassurance that we can restore code from CodeCommit in an emergency. There is also a school of thought that such things are deemed essential, thus should live in the production environment. But we're also inconsistent on this - for example, the 'deploy terraform' job is essential but exists in Integration.

In the absence of a clear policy of environments, integration seems as good a place as any to house this job.

Depends on https://github.com/alphagov/govuk-aws-data/pull/1021.

Trello: https://trello.com/c/QlPzhn6A/2814-move-code-repository-mirroring-off-of-concourse

### DIFF
--- a/terraform/projects/infra-govuk-repo-mirror/README.md
+++ b/terraform/projects/infra-govuk-repo-mirror/README.md
@@ -1,8 +1,8 @@
 ## Module: govuk-repo-mirror
 
-Configures a user and role to allow the govuk-repo-mirror Concourse pipeline
-to push to AWS CodeCommit (the user is used by the Jenkins
-Deploy\_App job and the role is used by the Concourse mirroring job)
+Configures a user and role to allow the "Mirror GitHub repositories"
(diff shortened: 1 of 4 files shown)


ChrisBAshton:
Superseded by #1579.

PR #1579, opened 2022-04-27
Allow Jenkins to assume the govuk-codecommit-poweruser role

ChrisBAshton:
Depends on https://github.com/alphagov/govuk-aws-data/pull/1022

These are the minimum changes required to move the repository
mirroring pipeline out of Concourse and into Jenkins. Whilst we
could go down the route of creating a brand new user and role
for Jenkins, the existing pipeline is quite tightly coupled to
the Concourse user/role and existing SSH keys, so is best left
as-is:
https://docs.publishing.service.gov.uk/manual/github-unavailable.html

We can look at removing outdated config and renaming away from
'Concourse' later once the new pipeline is working.

Trello: https://trello.com/c/QlPzhn6A/2814-move-code-repository-mirroring-off-of-concourse


fredericfran-gds (approved):
looks good

PR #1582, opened 2022-05-03
Fix #1579 by conditioning some additional resources

fredericfran-gds:
In #1579, the new resources should be created only in integration.
This PR fixes where some of the resources would have been
created in staging and be superfluous.


KludgeKML (approved):
Looks good to me.

PR #1583, opened 2022-05-04
Remove legacy roles

fredericfran-gds:
In #1412, a role is created for each user per GOV.UK privilege
category. This was done as a workaround to the size limit of a
role in ARN where we can't create a single role per GOV.UK
privilege category and add in all the users of that category in.

As a temporary measure, we left the legacy roles there but now
we hit again that limit so it's time to remove these legary
roles. Developers had time to switch to their new roles now.


fredericfran-gds:
close in favour of #1583

PR #1585, opened 2022-05-04
Remove legacy roles

fredericfran-gds:
In #1412, a role is created for each user per GOV.UK privilege
category. This was done as a workaround to the size limit of a
role in ARN where we can't create a single role per GOV.UK
privilege category and add in all the users of that category in.

As a temporary measure, we left the legacy roles there but now
we hit again that limit so it's time to remove these legary
roles. Developers had time to switch to their new roles now.


sengi (approved):

sengi:
Something for when you're back - 2nd-line found a bug in this today, unfortunately:

When adding a user, the first `terraform apply` fails because somehow TF tries to update the KMS policies before it's finished creating the user (or perhaps AWS returns success prematurely to TF; I didn't have time to investigate that far). AWS then throws a `MalformedPolicyDocumentException: Policy contains a statement with one or more invalid principals.` and the apply fails. Running the apply a second time succeeds, because the user exists by then and the policy validates just fine.

fredericfran-gds:
Maybe this will fix it: https://github.com/alphagov/govuk-aws/pull/1589. It is unfortunately an upstream issue as described in the commit.

PR #1587, opened 2022-05-10
Bump infra-public-services AWS provider to 2.69.0

sihugh:
This introduces WAFv2 resources which we'll be using.


leenagupte (approved):
All of the numbers seem to have been appropriately updated 👍

PR #1590, opened 2022-05-20
Fix aws_kms_key changes dependency

fredericfran-gds:
the time_sleep dependency had effect only when
creating/destroying and not changes.

this PR fixes that

See [doc](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/sleep#triggers)


rtrinque (approved):
looks good

PR #1593, opened 2022-05-30
Output the IDs of infra-public resources

sihugh:
These IDs will be referenced from a separate project which will configure the web acls for the public load balancers. The resources we're outputting are:

- IDs of public load balancers
- the ID of the existing kinesis data firehose

The load balancer IDs will be used in the new project to attach the new web ACLs to the public load balancers.

The kinesis data firehose is used to [send the logs generated by the new WAFs to S3.](https://aws.amazon.com/premiumsupport/knowledge-center/waf-configure-comprehensive-logging/).
This happens at present with the default web ACL, but the plan is to move that configuration out of the extremely large infra-public-services project into something a bit more easily manageable. 

We'll reuse this kinesis data firehose instance for the moment, until we need to do something different.

### DIFF
--- a/terraform/projects/infra-public-services/README.md
+++ b/terraform/projects/infra-public-services/README.md
@@ -325,4 +325,21 @@ This project adds global resources for app components:
 
 ## Outputs
 
-No outputs.
+| Name | Description |
+|------|-------------|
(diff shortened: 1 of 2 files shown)


leenagupte (approved):
I have approved this because the hashes look ok.

PR #1594, opened 2022-05-31
Add boilerplate for new WAFs project

sihugh:
This boilerplate was inspired by the set up for [govuk-rds](https://github.com/alphagov/govuk-aws/tree/main/terraform/projects/app-govuk-rds) which seems to be one of the few projects to uses a Terraform version > 1. I've chosen to match that version rather than choose newest and shiniest so that we have a little consistency.

There's nothing novel here yet, it's just setting up for what comes next.

This project will allow us to remove some config from the infra-public-services project (which is quite large), but should also allow more rapid deployment of WAF changes should they be necessary.


rtrinque (approved):
Looks good

PR #1596, opened 2022-06-10
Move WAF and shield configuration to new project

sihugh:
I've split this into 3 commits to hopefully make it easier to review. 

- copy and paste the default rule we created into the new project infra-public-wafs
- move all the config applying this rule to LBs into the new project
- remove the old rule configuration

Once it's been moved, we can then add other configuration to make these more useful.

There'll be a few minutes where we don't have the default rule applied to LBs, but it's not doing anything at present anyway, so it's not a problem.

This has been tested in integration:
- [infra-public-services apply](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/3055/console)
- [infra-public-wafs apply](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/3056/console)


rtrinque (approved):
Looks good

PR #1600, opened 2022-06-15
Add VPC endpoint to route traffic to S3

sihugh:
This allows us to communicate directly with S3 without using the NAT gateway.
It should save us a lot of money, as backups, crawler worker and lots of other things send quite a lot of data that way.

Further reading:

- https://www.lastweekinaws.com/blog/the-aws-managed-nat-gateway-is-unpleasant-and-not-recommended/
- https://docs.aws.amazon.com/vpc/latest/privatelink/gateway-endpoints.html

[Successful plan on integration](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/3108/console)

https://trello.com/c/qi2ocydX/2651-use-vpc-endpoints-for-s3


rtrinque (approved):
Looks good

PR #1602, opened 2022-06-16
app-knowledge-graph: also log stderr

maxf:
So that we get the error messages when the provision_knowledge_graph
script outputs errors to stderr


exfalsoquodlibet (approved):
Thanks.

[This post](https://www.brianstorti.com/understanding-shell-script-idiom-redirect/) was very useful for me to understand what all this was about :D

PR #1604, opened 2022-06-21
Knowledge graph new image

maxf:
We want to update neo4j but there's no suitable AMI for the latest version.
We could create our own, but it's simpler (and more visible) to just put install
instructions in the startup script.

See commits for details


nacnudus (approved):
I confirmed that there is no suitable AMI in the London region, that 4.4.8 is the latest version, and that the installation is as recommended by Neo4j, with the addition of an updated Java runtime.

PR #1605, opened 2022-06-21
Allow monitoring to access the mirror bucket

sihugh:
[Smokey is currently failing](https://alert.production.govuk.digital/cgi-bin/icinga/extinfo.cgi?type=2&host=ip-10-13-5-163.eu-west-1.compute.internal&service=Smokey+loop+for+mirror) because the monitoring machine is attempting to use the [new S3 gateway](https://github.com/alphagov/govuk-aws/pull/1605) to retrieve pages from the replica mirror and failing.

Looking at the policies on the bucket, we have locked them down to Fastly, office and AWS NAT ip addresses, none of which are used in this case.

### DIFF
--- a/terraform/projects/infra-mirror-bucket/mirror-read-policy.tf
+++ b/terraform/projects/infra-mirror-bucket/mirror-read-policy.tf
@@ -227,6 +227,27 @@ data "aws_iam_policy_document" "s3_mirror_replica_read_policy_doc" {
   }
 
   statement {
+    sid     = "S3NATGatewayReadBucket"
+    actions = ["s3:GetObject"]
+
+    resources = [
+      "arn:aws:s3:::${aws_s3_bucket.govuk-mirror-replica.id}",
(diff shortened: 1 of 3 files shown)


terraform/projects/infra-mirror-bucket/mirror-read-policy.tf
```diff
@@ -226,6 +226,27 @@ data "aws_iam_policy_document" "s3_mirror_replica_read_policy_doc" {
...
+    actions = ["s3:GetObject"]
+
+    resources = [
+      "arn:aws:s3:::${aws_s3_bucket.govuk-mirror-replica.id}",
+      "arn:aws:s3:::${aws_s3_bucket.govuk-mirror-replica.id}/*",
+    ]
+
+    condition {
+      test     = "StringEquals"
+      variable = "aws:SourceVpce"
```
chao-xian, line 240: Is this meant to have an `e` at the end?

rtrinque (approved):
Looks good

PR #1607, opened 2022-06-23
Knowledge graph: add pip version of aws cli

maxf:
We'd previously removed it, assuming that the preinstalled version was enough,
but commands in the provisioning script only work with the pip version. So we're
re-adding it


maxf:
OK, I'm going to make it upgrade to AWSCLI 2. So ignore this for now.

exfalsoquodlibet (commented):
I cannot test this. Having review the linked AWS documentation, the changes makes sense.

Two minor suggestions to be over-zealously careful to consider.

terraform/projects/app-knowledge-graph/userdata.tpl
```diff
@@ -1,7 +1,23 @@
...
+which -a aws && echo "aws binary found after uninstalling. Something's wrong."
+echo "- downloaded new version"
+curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+unzip -q awscliv2.zip
+echo "- installing new version"
+sudo ./aws/install
+if [ "$(which aws)" == "/usr/local/bin/aws" ]; then
+    echo "aws found at /usr/local/bin/aws. All good."
+else
+    echo "No aws found at /usr/local/bin/aws. Something's wrong."
```
exfalsoquodlibet, line 19: This is just for my own learning :D how do you know `/usr/local/bin/aws` is the right directory here?

exfalsoquodlibet (approved):
Thanks Max! Looks good to me

terraform/projects/app-knowledge-graph/userdata.tpl
maxf: I check on the running govgraph, where I'd follow the script manually, leading to 
```
ubuntu@ip-10-1-1-105:~$ ls -l `which aws`
lrwxrwxrwx 1 root root 37 Jun 24 07:31 /usr/local/bin/aws -> /usr/local/aws-cli/v2/current/bin/aws
```

PR #1611, opened 2022-06-30
Redirect data.gov.uk to www.data.gov.uk

kentsanggds:
## What 

The reason for this is to prevent cookies set on the apex domain, data.gov.uk, from also being set to sub domains of data.gov.uk.

## Reference 

https://trello.com/c/Z3sWmLp7/948-figure-out-how-to-handle-ga-cookies-on-apex-domain


sengi (approved):
Change looks ok to me (as someone with only a passing familiarity with VCL and Varnish). How did you test it and how do you plan to roll it out?

PR #1617, opened 2022-07-08
Setup the Trusty ESM repo to be used during provisioning

rtrinque:
We modify the way the user-data script is generated to include the setup of the ESM repo so that we can apply the latest supported update for Trusty as early as the provisioning phase.
To avoid letting the secret token used in the setup in clear in the code, we use an intermediary beacon text that is then replaced by the actual secret.


ChrisBAshton (approved):
LGTM 👍

PR #1621, opened 2022-07-22
Move Gatling to blue stack

fredericfran-gds:
This is required because Gatling being on the govuk stack interferes
with the check_reboots_required [script](https://github.com/alphagov/govuk-puppet/blob/ffb61bd9bdeb176663955f6787c7db5bdd4deb63/modules/monitoring/files/usr/lib/nagios/plugins/check_reboots_required#L46)
considers Gatling to have an unknown reboot label due to it being in the
govuk stack and terminates/exits early. This gives an incorrect set of machines
to be rebooted in icinga monitoring. This doesn't affect the reboot process itself,
only the monitoring.

Note that the govuk_node_list is also geared on using only the blue stackname also:
https://github.com/alphagov/govuk-puppet/blob/ffb61bd9bdeb176663955f6787c7db5bdd4deb63/modules/govuk_scripts/templates/govuk_node_list_aws.erb#L13)

Hence, it is better to move GAtling to the Blue stack


ChrisBAshton (approved):
Nice 👍 thanks!

PR #1622, opened 2022-08-05
Allocate EIPs for future use by GOV.UK Licensing

richardTowers:
We've been trying for a long time to get Civica to update their ACLs to allow our AWS environments to make API calls to their infrastructure, on behalf of GOV.UK Licensing's local authority users.

We're starting to see some promising signs that they might finally do this, and we want to make sure that we've got a reasonably future proof set of IP addresses to give them.

Currently, API calls to Civica go through a VPN that runs to UK Cloud.

We'd like to remove this, and have API calls go through the NAT Gateways in our production and staging VPCs. There are three NAT gateways in both environments, so we have 6 IP addresses which need to be allowlisted.

In the future, there's a high chance we'll want to move Licensing into its own VPC (or possibly its own AWS account). If this happens, we'll end up with another set of IP addresses, and we'd have to do the whole dance with Civica again.

To avoid this, this commit allocates 3 EIPs in our production VPC and 3 EIPs in our staging VPC. These will initially not be assigned to anything, but we will ask Civica to add them to their allow lists. When we eventually move Licensing, we'll be able to ask AWS support to move these EIPs to the appropriate VPC / AWS account.

NOTE: I already created one of these EIPs in production, while I was testing. Unfortunately I can't remove it again, because of the DenyEipRelease IAM policy. Instead, I've imported it into the terraform state as `aws_eip.licensify_reservation[0]`. This means that the terraform plan for this PR should only attempt to add two EIPs in production (because one is already there), but three in staging. Apologies for testing directly in production.

### DIFF
--- a/terraform/projects/infra-vpc/README.md
+++ b/terraform/projects/infra-vpc/README.md
@@ -28,6 +28,7 @@ and resources to export these logs to S3
 | Name | Type |
 |------|------|
 | [aws_cloudwatch_log_group.log](https://registry.terraform.io/providers/hashicorp/aws/2.46.0/docs/resources/cloudwatch_log_group) | resource |
+| [aws_eip.licensify_reservation](https://registry.terraform.io/providers/hashicorp/aws/2.46.0/docs/resources/eip) | resource |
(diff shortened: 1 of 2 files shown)


sengi (approved):
Thanks for taking care of this!

PR #1623, opened 2022-08-16
Set up new rule for public cache LB WAF

sihugh:
The first commit is copying the default web ACL so that:
 - the smokey tests that check for x-always-block still trigger
 - we can add rules to the new web ACL that just affect the public cache LB

The second commit explicitly allows traffic from GOV.UK's infra to hit the LB which will skip subsequent rules.

The third commit allows healthchecks from our CDN to bypass subsequent rules.

This precedes adding a [blanket rate limit](https://aws.amazon.com/blogs/security/three-most-important-aws-waf-rate-based-rules/) for other traffic. At times, we have things like link-checker-api, crawler-worker and smokey that request things in bursts. The majority of the time, those requests are served from Fastly cache, but we want them to work regardless. We can look at adding rate limit tokens to these in future.

Requests from inside the GOV.UK VPC to [www.gov.uk](http://www.gov.uk/) should come via one of the NAT gateway IP addresses. We convert the originating IP to the True-Client-IP header in our CDN config. By setting this to `allow` for these addresses, [the evaluation of the web ACL stops if a match is made](https://docs.aws.amazon.com/waf/latest/developerguide/waf-rule-action.html), and the request is permitted through the WAF.

Requests from [Fastly's healthcheck have a rate limit token](https://github.com/alphagov/govuk-cdn-config/blob/main/vcl_templates/www.vcl.erb#L270) which we want to allow

Non-matching requests continue being evaluated against the other rules which will arrive soon.


terraform/projects/infra-public-wafs/cache_public_rule.tf
```diff
@@ -0,0 +1,44 @@
+# this rule matches any request that contains the header X-Always-Block: true
+# we use it as a simple sanity check / acceptance test from smokey to ensure that
+# the waf is enabled and processing requests
+#
+resource "aws_wafv2_web_acl" "cache_public" {
```
ChrisBAshton, line 1: Did the contents of this file get copied and pasted from somewhere? If so, could you link to the original source in the commit?

ChrisBAshton (commented):
Changes LGTM 👍 just needs a docs recompilation I think.

terraform/projects/infra-public-wafs/cache_public_rule.tf
sihugh: Yep, good point - it came from https://github.com/alphagov/govuk-aws/blob/main/terraform/projects/infra-public-wafs/default_rule.tf, but will amend the commit 👍

ChrisBAshton (approved):
LGTM ⭐

PR #1626, opened 2022-08-25
Add our own owner id to the list that images are filtered by.

rtrinque:
The Canonical AMI we previously depended on is not available
anymore, we've made a copy that we own so that we can continue to
use it. In order to make it findable we add our own owner id to
the search filter.


Tetrino (approved):
Good work!

PR #1629, opened 2022-09-05
Allow traffic replay to bypass rate limiting

sihugh:
In integration and staging, production traffic is replayed to some extent. The rate limiting works using the IP address of the request originator, which in the case of traffic replay is limited to a couple of addresses on our side.

This will not affect production, as traffic is not replayed there.

As before, the IP address is provided in a CIDR block format. This is specified already in the [common config](https://github.com/alphagov/govuk-aws-data/blob/main/data/common/staging/common.tfvars#L72)

Temporarily leaving the old IP set in, as it doesn't appear to delete cleanly if all done at once.

Also re-adding metrics for the rules that bypass the rate limiting, as this makes troubleshooting easier


ChrisBAshton (approved):
Nice ⭐

PR #1630, opened 2022-09-08
Allow Google Cloud Platform to read a bucket

nacnudus:
This allows a particular GCP project to read the bucket govuk-integration-database-backups, so that other GCP projects can securely and easily access GOV.UK's data.  The approach has been [discussed](https://docs.google.com/document/d/1Nw7xqXZQn6PIjkB_XsXuAFa2iyN9s6Booi9HxtpyGCE).

The GCP project uses the "Storage Transfer" service to sync a GCP bucket with the AWS bucket.  The service uses a default service account (i.e. one that Google manages).  This pull request allows that service account, identified by its "subject ID", given here as `accounts.google.com:sub`, to assume an IAM role that has permission to read the contents of the AWS bucket.  No keys are required.  See [Google's documentation of this arrangement](https://cloud.google.com/storage-transfer/docs/source-amazon-s3#federated_identity).

The corresponding [GCP config](https://github.com/alphagov/govuk-s3-mirror/blob/main/terraform/transfer.tf).  Currently it only syncs some of the objects in the bucket, because those are the ones currently being used by other projects in GCP.  See [Google's documentation of this arrangement](https://cloud.google.com/storage-transfer/docs/create-transfers#amazon-s3).  See also [Terraform's example configuration](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_transfer_job#example-usage).

There is a [precedent](https://github.com/alphagov/govuk-aws/blob/main/terraform/projects/infra-google-mirror-bucket/main.tf), which backs up the buckets govuk-production-mirror and govuk-staging-mirror.  It uses a key, which requires manual maintenance.

### DIFF
--- a/terraform/projects/infra-security/main.tf
+++ b/terraform/projects/infra-security/main.tf
@@ -400,6 +400,61 @@ resource "aws_iam_role_policy_attachment" "shield-response-team-access" {
   policy_arn = aws_iam_policy.shield-response-team-access.arn
 }
 
+data "aws_iam_policy_document" "google_s3_mirror" {
+  count = var.aws_environment == "integration" ? 1 : 0
+
+  statement {
+    sid = "GoogleReadBucket"
+
+    actions = [
+      "s3:Get*",
+      "s3:List*",
+    ]
+
+    # Need access to the top level of the tree.
+    resources = [
+      "arn:aws:s3:::govuk-integration-database-backups",
+      "arn:aws:s3:::govuk-integration-database-backups/*",
+    ]
+  }
+}
+
+resource "aws_iam_policy" "google-s3-mirror" {
+  count       = var.aws_environment == "integration" ? 1 : 0
+  name        = "google-s3-mirror"
+  description = "Allows a Google Cloud Platform project to mirror S3 buckets."
+  policy      = data.aws_iam_policy_document.google_s3_mirror[0].json
+}
+
+resource "aws_iam_role" "google-s3-mirror" {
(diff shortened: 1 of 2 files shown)


terraform/projects/infra-security/main.tf
```diff
@@ -400,6 +400,56 @@ resource "aws_iam_role_policy_attachment" "shield-response-team-access" {
...
+
+    actions = [
+      "s3:Get*",
+      "s3:List*",
+    ]
+
+    # Need access to the top level of the tree.
+    resources = [
+      "arn:aws:s3:::govuk-integration-database-backups",
+      "arn:aws:s3:::govuk-integration-database-backups/*",
```
richardTowers, line 415: Because these bucket names are different in the different environments, I think we'll need to either do something with variables or scope all of the resources in this file down so they only appear in integration.

(For context, the infra-security terraform deployment needs to apply in integration / staging and production)

I'd suggest putting something like:

```
count = "${var.aws_environment == "integration" ? 1 : 0}"
```

On each of the resources to stop terraform from trying to create them in non-integration environments. I'm think infra-security already has a variable for aws_environment.

richardTowers: There might be a nicer syntax to do that if infra-security is on a newer version of terraform.

nacnudus: Thanks for spotting that and suggesting the fix.  Done in ee0c629ae37c53aeb0744d8ca7b9125bb4cead7e.

PR #1631, opened 2022-09-08
updated node version to prevent jenkins failed job

joshpedroAL:


deborahchua:
@joshpedroAL has confirmed we no longer need this.

PR #1639, opened 2022-10-11
Swap cache public WAF to CloudWatch logging

sihugh:
We  currently use a kinesis firehose to send log entries to Splunk. There's a lambda that runs on each log entry to filter out things that aren't relevant.

Since the kinesis firehose was configured, AWS announced support for native [WAF log filtering](https://aws.amazon.com/about-aws/whats-new/2021/05/aws-waf-adds-support-for-log-filtering/).

If this works as expected, we should eventually be able to ditch [the kinesis/lambda set up](https://github.com/alphagov/govuk-aws/blob/main/terraform/projects/infra-public-services/waf.tf#L67-L101), which will be one less thing for us to worry about keeping up to date.

Logging to CloudWatch also gives us more ability to monitor what's happening on tools that we are more familiar with.

I expect we'll eventually ship the logs off to Splunk as well, but we can probably avoid doing that while we figure things out.

The intention for this is to log requests that trigger a COUNT or a BLOCK. Things that are allowed will be logged elsewhere in our infra, so should be easy enough to track down.

This is similar to the [existing lambda](https://github.com/alphagov/govuk-aws/blob/4e81ceee540e7f1a521a638a3a214b336289cad8/terraform/lambda/WAFLogTrimmer/lambda.js) which forwards on everything that isn't a the "Default Action". We do now though explicitly allow some traffic, and we probably don't need to log that as well.

When things are in CloudWatch we can [use the docs](https://aws.amazon.com/premiumsupport/knowledge-center/waf-analyze-logs-stored-cloudwatch-s3/) a bit more straightforwardly too. We'll tweak the retention period per
environment once we're sure the filtering is working as expected.

https://trello.com/c/NK1JYc96/2759-consolidate-waf-logging-on-cloudwatch

### DIFF
--- a/terraform/projects/infra-public-wafs/cache_public_rule.tf
+++ b/terraform/projects/infra-public-wafs/cache_public_rule.tf
@@ -153,7 +153,41 @@ resource "aws_wafv2_ip_set" "govuk_requesting_ips" {
   addresses          = concat(var.traffic_replay_ips, local.nat_gateway_ips)
 }
 
-resource "aws_wafv2_web_acl_logging_configuration" "public_cache_web_acl_logging" {
-  log_destination_configs = [data.terraform_remote_state.infra_public_services.outputs.kinesis_firehose_splunk_arn]
+resource "aws_cloudwatch_log_group" "public_cache_waf" {
+  # the name must start with aws-waf-logs
+  # https://docs.aws.amazon.com/waf/latest/developerguide/logging-cw-logs.html#logging-cw-logs-naming
+  name              = "aws-waf-logs-cache-public-${var.aws_environment}"
+  retention_in_days = var.waf_log_retention_days
+
+  tags = {
+    Project       = var.stackname
+    aws_stackname = var.stackname
+  }
+}
+
+resource "aws_wafv2_web_acl_logging_configuration" "public_cache_waf" {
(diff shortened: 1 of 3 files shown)


terraform/projects/infra-public-wafs/cache_public_rule.tf
```diff
@@ -153,7 +153,40 @@ resource "aws_wafv2_ip_set" "govuk_requesting_ips" {
   addresses          = concat(var.traffic_replay_ips, local.nat_gateway_ips)
 }
 
-resource "aws_wafv2_web_acl_logging_configuration" "public_cache_web_acl_logging" {
-  log_destination_configs = [data.terraform_remote_state.infra_public_services.outputs.kinesis_firehose_splunk_arn]
+resource "aws_cloudwatch_log_group" "public_cache_waf" {
+  # the name must start with aws-waf-logs
```
ChrisBAshton, line 157: Can you add an inline comment to a URL that describes this requirement?

ChrisBAshton (approved):
LGTM

terraform/projects/infra-public-wafs/cache_public_rule.tf
sihugh: 👍  And have added a line to the commit message quoting the actual instructions.

PR #1640, opened 2022-10-11
Allow authenticating-proxy to talk to RDS for Postgres

KludgeKML:
Adds security rules to allow AP to talk to RDS, needed for the Postgres switch.

NB: This adds elements, it does not remove any existing elements (ie it is safe to deploy even if we have to roll back, because the existing infrastructure for MongoDB is untouched)

Authenticating Proxy PR:
https://github.com/alphagov/authenticating-proxy/pull/350

Related PRs:
https://github.com/alphagov/govuk-aws-data/pull/1092
https://github.com/alphagov/govuk-puppet/pull/11967
https://github.com/alphagov/govuk-secrets/pull/1370

## Why


AgaDufrat (approved):
Tested it in integration and staging – all works well

PR #1647, opened 2022-11-07
Migrate DownloadLogsAnalytics lambda function

catalinailie:
We are planning to retire govuk-lambda-app-deployment repo. In order to do that, we need to migrate the labdas that are still used in gouvuk-aws and kill the unused ones.


catalinailie:
Opened a new pr with the relevant changes

PR #1651, opened 2022-11-18
Remove MySQL XtraBackup code

ChrisBAshton:
This harks back to an era where GOV.UK managed its own MySQL clusters in VMs spread across EC2 instances. With that infra, we had to handle our own replication and backup logic. GOV.UK moved to using a managed RDS service sometime after August 2017: https://docs.publishing.service.gov.uk/repos/govuk-aws/architecture/decisions/0018-use-rds-instead-of-provisioned-ec2-databases.html

Self-managed MySQL no longer exists in GOV.UK's infrastructure except for on the Jenkins CI agents.

Logging into Staging and Production AWS environments, we can see that `govuk-mysql-xtrabackups-${ENV}` and
`govuk-whitehall-mysql-xtrabackups-${ENV}` buckets exist in each, but they are all empty, so this code does not seem to be being used.


kevindew (approved):
Yay 🎉 I'm not super confident in this so do ask wider if you want a stronger opinion on if this is definitely fine to go - but I trust you with removing it.

There's also https://github.com/alphagov/govuk-aws-data/tree/main/data/app-mysql-xtrabackups/production

nsabri1 (approved):
OK

PR #1652, opened 2022-11-18
Upgrade nodejs versions

MuriloDalRi:
The current versions are no longer supported and we should be using 16.

This changes it to the next major version and there will be a follow up PR to get to 16.

See https://github.com/alphagov/govuk-aws/commit/af88c54a97eb9cfac19bdac7ce45408ee468da07 for more info.

https://trello.com/c/PMsAXSoT/3029-review-govuks-usage-of-node-2


terraform/projects/infra-mirror-bucket/main.tf
```diff
@@ -621,6 +621,6 @@ resource "aws_lambda_function" "url_rewrite" {
   function_name = "url_rewrite"
   role          = "${aws_iam_role.basic_lambda_role.arn}"
   handler       = "index.handler"
-  runtime       = "nodejs10.x"
+  runtime       = "nodejs12.x"
```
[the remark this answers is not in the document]
ops20: discussed with Murilo and is ok as the PR is to update each runtime to the next major version and check for issues. eventually it will be updated to nodejs16.x.

ops20 (approved):
LGTM

terraform/projects/infra-mirror-bucket/main.tf
```diff
@@ -621,6 +621,6 @@ resource "aws_lambda_function" "url_rewrite" {
   function_name = "url_rewrite"
   role          = "${aws_iam_role.basic_lambda_role.arn}"
   handler       = "index.handler"
-  runtime       = "nodejs10.x"
```
ChrisBAshton, line 624: I went down a rabbit home while trying to answer the question: "what's the best way of testing that this doesn't break"?

The [lambda code](https://github.com/alphagov/govuk-aws/blob/main/terraform/lambda/CloudfrontUrlRewrite/index.js) this refers to basically rewrites the URL to append a `.html`, so that the requested file can be found in the S3 bucket (a request for `/foo` may work on GOV.UK, but the 'mirrored version' is saved as `/foo.html`).

This duplicates logic that lives in our [VCL for Fastly](https://github.com/alphagov/govuk-cdn-config/blob/3b8471bfacb3356442201a0b47a964d723b4f9b1/vcl_templates/www.vcl.erb#L302-L332), where if Fastly doesn't have something cached, it makes a request to Origin, and if Origin times out or returns a 5xx error, it will send a request to AWS -> Cloudfront -> S3 instead, and Fastly does the hard work of adding `.html` to the request.

So where does this Lambda get used? I believe it's in the failover scenario, i.e. Fastly goes down, so we update GOV.UK's DNS and point it to AWS (Cloudfront) instead. Currently, Cloudfront serves mirrored content from the S3 bucket.

Without Fastly doing the work of converting `/foo` to `/foo.html`, then requests go through to Cloudfront directly as `/foo`, and so Cloudfront needs to do the `.html` appending, which is why this Lambda exists.

Some thoughts:

1. Why do we need URL rewriting inside both Fastly _and_ Cloudfront? We could remove it from Fastly and just leave it to Cloudfront to do. I suspect the Cloudfront lambda is already being invoked anyway, but ends up doing noop because Fastly has already ensured there's a `.html` suffix.

2. We're changing our failover so that Cloudfront routes requests to Origin, not the mirrors. In which case, we don't _want_ to add the `.html` suffix at the Fastly level, because that would break the routing at Origin. So that's another reason to remove it from the Fastly VCL. (Our config for our primary CDN shouldn't know the internals of our config for the secondary CDN).

We're in a slightly awkward situation where we're about to change our Cloudfront distribution for an improved failover plan, but we still need the existing failover & mirror fallbacks to work in the meantime.

I note that this lambda is invoked on the `Default (*)` Behaviour on Production's Cloudfront (which serves mirrored content) but we don't have that on Staging (probably removed while testing).

I think we should get Staging's Cloudfront back to how it was (serving mirrored content), update this lambda, then it should be easy enough to try viewing a `/foo` Cloudfront URL and ensuring Cloudfront serves the `foo.html` version.

May need some coordination with @rtrinque 👀 

Eventually there'll be a different setup for the mirror fallback, but we're not there yet. I imagine something involving multiple Cloudfront distributions, something like:

If Origin is down, point to mirrors:

* Fastly -> Origin Cloudfront -> ~~Origin~~
* Fastly -> Mirror Cloudfront -> S3

If Fastly is down, switch to dynamic Cloudfront:

* ~~Fastly~~ Origin Cloudfront -> Origin

If Fastly is down AND Origin is down, automatically retry a fetch from the mirrors

* ~~Fastly~~ Origin Cloudfront -> ~~Origin~~ -> Mirror Cloudfront -> S3

(In this case, we would be able to manually test the URL rewrite works by accessing the Mirror Cloudfront.)

ChrisBAshton: Thanks for confirming 👍 so my suspicion was that this lambda may not be needed once the 'dynamic origin' cloudfront has been enabled. It sounds like that's the case.
So the best course of action would be to:

1) Ignore this lambda
2) Implement dynamic origin cloudfront
3) Delete the (now unused) lambda
4) One day, in the future, perhaps create a variant of the lambda again, for a fallback 'mirror cloudfront' if needed

Does that sound about right? @rtrinque

rtrinque: So the new cloudfront will indeed replace the current one.
The new cloudfront will be able to serve from two origins, cache LB, and S3 mirrors.
So we sill still need the rewriting lambda in the scenario where we lost fastly and origin

ChrisBAshton: @MuriloDalRi @rtrinque we've discussed this offline - sounds like Roch is going to attempt to copy the lambda, update its Node version, and get it working with the new Cloudfront distribution?

Though in the meantime, I think we'd still need to update this lambda because Fastly will continue to request content from the mirrors when Origin is unavailable, and that uses this lambda.

@rtrinque can you confirm we still expect this lambda to be needed, and therefore that it should be updated?
Can you advise on how best to test it, given Staging is being used to test the new Cloudfront?

Thanks!

robinjam:
It looks like this change broke the `infra-mirror-bucket` project:

https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4649/console

> Error: aws_lambda_function.url_rewrite: expected runtime to be one of [dotnetcore1.0 dotnetcore2.0 dotnetcore2.1 go1.x java8 java11 nodejs4.3 nodejs4.3-edge nodejs6.10 nodejs8.10 nodejs10.x nodejs12.x provided python2.7 python3.6 python3.7 python3.8 ruby2.5], got nodejs16.x

I think we'll need to bump the version of the AWS provider (or terraform itself) to target Node 16

PR #1654, opened 2022-11-22
Add new Owner ID where specified in projects rather than centrally.

KludgeKML:
The AMI Owner ID that was needed to fix the central node_group module (here: https://github.com/alphagov/govuk-aws/pull/1626)  is also needed in 6 other places where AMI Owner IDs have been specified explicitly.


rtrinque (approved):
Looks good to me

PR #1655, opened 2022-11-24
Add amazonmq project [integration-only, not for production yet]

aldavidson:
As part of [this epic](https://trello.com/c/O8RFciiC/309-migrate-self-hosted-eol-rabbitmq-to-amazonmq), we're converting our tech spike ('can we replace RabbitMQ with AmazonMQ?') to a working installation on integration. We'll be learning & tweaking as we go, so this is just the first of probably many PRs. 

This PR just creates the AmazonMQ broker along with associated security groups, and a Route53 entry pointing to the broker.
It's INTEGRATION-ONLY at this stage - will need reworking for a highly-available cluster as we will have in staging/prod.

It generates all RabbitMQ user passwords on create, but they are marked as `sensitive` so they will not appear in output or logs by default. The passwords can be retrieved from the remote state with `terraform output -json`.
 
There's also an [extension to govuk-connect](https://github.com/alphagov/govuk-connect/pull/78) which will allow you to connect to the RabbitMQ web management interface via a remote port-forward.


terraform/projects/app-amazonmq/main.tf
```diff
@@ -0,0 +1,123 @@
...
+}
+
+# internal_domain_name is ${var.stackname}.${internal_root_domain_name}
+resource "aws_route53_record" "amazonmq_internal_domain_name" {
+  zone_id = data.aws_route53_zone.internal.zone_id
+  name    = "${aws_mq_broker.publishing_amazonmq.broker_name}.${var.internal_domain_name}"
+  type    = "CNAME"
+  # records = [aws_mq_broker.publishing_amazonmq.instances.0.ip_address]
+  ttl     = 300
+  records = [regex(".+://([^/:]+)", aws_mq_broker.publishing_amazonmq.instances.0.console_url)[0]]
```
sengi, line 112: Is there a reason for scraping the hostname out of the `console_url` attribute vs just using the `instances[].endpoints` attribute?

terraform/projects/app-amazonmq/outputs.tf
```diff
@@ -0,0 +1,11 @@
+output "broker_id" {
+  value = aws_mq_broker.publishing_amazonmq.id
+}
+
+output "console_url" {
+  value = aws_mq_broker.publishing_amazonmq.instances.0.console_url
+}
+
+output "endpoint" {
```
sengi, line 9: This could use a description to say what form this "endpoint" takes (e.g. as a new user to this module I don't know if this is an AMQP URL or a comma-separated list of node hostnames or something else)

sengi (approved):

terraform/projects/app-amazonmq/main.tf
sengi: Ah fair enough. If it makes sense to do this here rather than at the caller then that seems reasonable then. (Though I would have thought every user of the provider would have the same problem, which suggests there might be something weird/unexpected about our setup that could be improved on 🤷)

sengi: Oh I just realised the weird/unexpected thing about our setup is probably those clunky homegrown env vars that https://github.com/alphagov/govuk_message_queue_consumer/ expects, right? 😅 That'd explain why (for now) we want a list of hostnames rather than a more usual AMQP URL.

terraform/projects/app-amazonmq/main.tf
```diff
@@ -0,0 +1,123 @@
...
+  engine_version      = var.engine_version
+  deployment_mode     = var.deployment_mode
+  host_instance_type  = var.host_instance_type
+  publicly_accessible = var.publicly_accessible
+  security_groups     = ["${aws_security_group.amazonmq.id}"]
+  # must be a single ID if deployment_mode is SINGLE_INSTANCE
+  subnet_ids = [data.terraform_remote_state.infra_networking.outputs.private_subnet_ids[0]]
+
+  # The Terraform provider will only allow us to create a single user
+  # All other users must be added from the web UI
```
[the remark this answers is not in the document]
aldavidson: good call, thanks :+1:

terraform/projects/app-amazonmq/main.tf
```diff
@@ -0,0 +1,123 @@
...
+  security_groups     = ["${aws_security_group.amazonmq.id}"]
+  # must be a single ID if deployment_mode is SINGLE_INSTANCE
+  subnet_ids = [data.terraform_remote_state.infra_networking.outputs.private_subnet_ids[0]]
+
+  # The Terraform provider will only allow us to create a single user
+  # All other users must be added from the web UI 
+  user {
+    console_access = true
+    username       = "root"
+    password       = var.amazonmq_root_password
```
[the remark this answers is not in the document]
aldavidson: I like the idea of generating the password, will give that a go.

It will mean we can get rid of the encrypted `amazonmq_root_password` secret, which is definitely a plus.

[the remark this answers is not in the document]
aldavidson: The only drawback I can see is that we'll need to output the root password in order for whoever's running it to log in to the web admin UI, which means it will be in the console output for builds in the CI. Could that be a security issue? 
I guess that question is roughly equivalent to "are there any users who have access to our CI who we would not be comfortable having the root RabbitMQ password?"

terraform/projects/app-amazonmq/main.tf
aldavidson: The 'proper' way of doing a custom domain would be to also create a Network Load Balancer that balances between the MQ instances via their IPs, and point the route53 entry at the NLB (there's an [AWS article about it](https://aws.amazon.com/blogs/compute/creating-static-custom-domain-endpoints-with-amazon-mq-for-rabbitmq/) ). 

But frankly that's a bit of complexity that I was hoping to not have to deal with at this stage - I wanted to keep this initial `integration` implementation as simple as possible, and deal with the multi-instance issues when I'm ready to set it up on `staging`. This might well be the 'something weird about our setup' that you suspected, of course...

Are you OK if I do that as a separate PR later?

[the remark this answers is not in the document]
aldavidson: Good spot, thanks

terraform/projects/app-amazonmq/outputs.tf
```diff
@@ -0,0 +1,11 @@
...
+  value = aws_mq_broker.publishing_amazonmq.instances.0.endpoints.0
```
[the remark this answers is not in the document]
aldavidson: You're right, I think this will become a list of endpoint URLs for a HA setup. Not sure yet about the `console_url`. 

Like in the [previous comment](https://github.com/alphagov/govuk-aws/pull/1655#discussion_r1034139630) I was planning to deal with that at the point of setup on `staging`, to try and keep this initial `integration` setup as simple as possible. I'm  iterating towards the final state, essentially.

terraform/projects/app-amazonmq/README.md
```diff
@@ -0,0 +1,78 @@
+## Project: app-backend
```
kevindew, line 1: Some C+P left behind

terraform/projects/app-amazonmq/main.tf
```diff
@@ -0,0 +1,125 @@
+/**
+ * ## Project: app-backend
```
kevindew, line 2: Some C+P left behind

terraform/projects/app-amazonmq/main.tf
```diff
@@ -0,0 +1,125 @@
...
+ * 
+ * Module app-amazonmq creates an Amazon MQ instance or cluster for GOV.UK.
```
kevindew, line 4: With calling this "app-amazonmq", would the intention be to use this for future AmazonMQ usages (so publishing and the crawler)? I felt like there’s perhaps a bit of a smell that the name seems generic but the inputs are specifically publishing queue, and I wonder if life would be simpler to give it a specific name like publishing-amazonmq to not leave an impression of reuse/extensability for other apps.

terraform/projects/app-amazonmq/main.tf
```diff
@@ -0,0 +1,125 @@
...
+  }
+}
+
+provider "aws" {
+  region = var.aws_region
+}
+
+# --------------------------------------------------------------
+resource "aws_mq_broker" "publishing_amazonmq" {
+  broker_name = "PublishingMQ"
```
kevindew, line 31: I guess there isn’t a way we can grab this from the data input?

terraform/projects/app-amazonmq/variables.tf
```diff
@@ -0,0 +1,80 @@
+
```
kevindew, line 1: Is there an advantage to having variables.tf and outputs.tf distinct from main.tf - I notice there’s only a couple of projects that use them as separate files aside from this with the majority including them in main.tf. I’m guessing they’re normally only separate to collate multiple terraform files.

terraform/projects/app-amazonmq/variables.tf
```diff
@@ -0,0 +1,80 @@
...
+variable "engine_type" {
+  type        = string
+  description = "ActiveMQ or RabbitMQ"
+  default     = "RabbitMQ"
+}
+
+variable "engine_version" {
+  type        = string
+  description = "broker engine version"
+  default     = "3.9.16"
```
kevindew, line 28: Is this something that’ll have to be in sync with the `"rabbit_version": “3.6.15”` from the rabbitmq definition config? (currently in https://github.com/alphagov/govuk-aws-data/pull/1127)

kevindew (commented):
Amazing work @aldavidson - really great seeing this come to life after so many years.

I wondered a bit about the security groups so that apps can talk to RabbitMQ, am I right that they’re currently out of scope for this PR?

terraform/projects/app-amazonmq/main.tf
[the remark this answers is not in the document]
sengi: Just need to specify `sensitive = true` when you define the Terraform output for the password and it won't show in stdout/stderr. Authorised people can look up the password using Terraform if/when they need it.

terraform/projects/app-amazonmq/main.tf
sengi: Oof, classic Amazon. Sure, no probs with this being a non-prod solution for now — though don't forget to signpost that with a TODO in the module description/comments/readme or whatever, and update the PR and commit messages. (I've seen non-prod-ready stuff go into prod by mistake so many times here it's unreal 😅)

terraform/projects/app-amazonmq/outputs.tf
aldavidson: done

terraform/projects/app-amazonmq/README.md
aldavidson: oops, good spot - fixed

terraform/projects/app-amazonmq/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+Module app-amazonmq creates an Amazon MQ instance or cluster for GOV.UK.
```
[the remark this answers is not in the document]
aldavidson: That's a good call. I'd based it on the other project names in the repo (e.g. `app-rabbitmq`) but happy to change it.

terraform/projects/app-amazonmq/main.tf
aldavidson: yep, agreed, see previous comment on the README

terraform/projects/app-amazonmq/main.tf
```diff
@@ -0,0 +1,125 @@
...
+
+resource "aws_security_group_rule" "amazonmq_ingress_any_https" {
+  type      = "ingress"
+  from_port = 443
+  to_port   = 443
+  protocol  = "tcp"
+
+  # Which security group is the rule assigned to
+  security_group_id = aws_security_group.amazonmq.id
+  cidr_blocks       = ["0.0.0.0/0"]
```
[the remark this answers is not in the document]
aldavidson: Yes, I originally had it as `office_ips` - expanded it to allow HTTPS from anywhere (although it's in a private subnet so 'anywhere' is effectively 'anywhere inside the VPC or VPN') when I couldn't get the connectivity working. Since figuring out the [port-forwarding](https://github.com/alphagov/govuk-connect/pull/78) I may be able to revert back to office_ips - will have a try

terraform/projects/app-amazonmq/variables.tf
aldavidson: Just clarity really - I find once the main.tf gets over a couple of hundred lines or so, it's easier (for me) to split outputs and variables out.

terraform/projects/app-amazonmq/variables.tf
aldavidson: only to a degree - and hopefully that degree is 'major version' if they're using semver properly.
I've tried posting the 3.6.15 definition to the 3.9.16 broker and it works fine, so it looks like they are.

terraform/projects/app-amazonmq/main.tf
aldavidson: not sure which data input you mean? I can easily make this a variable though, if that helps

aldavidson:
> I wondered a bit about the security groups so that apps can talk to RabbitMQ, am I right that they’re currently out of scope for this PR?
No, the security groups are in scope for this PR.

terraform/projects/app-amazonmq/main.tf
```diff
@@ -0,0 +1,123 @@
...
+  name        = "${var.stackname}_amazonmq_access"
+  vpc_id      = data.terraform_remote_state.infra_vpc.outputs.vpc_id
+  description = "Access to Amazonmq from named SGs"
+
+  tags = {
+    Name = "${var.stackname}_amazonmq_access"
+  }
+}
+
+resource "aws_security_group_rule" "amazonmq_ingress_management_amqp" {
```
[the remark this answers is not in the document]
aldavidson: My thinking on this was to keep this as separate and isolated as possible from anything existing, for a few reasons:

* I'm iterating rapidly, making lots of mistakes, and tearing things down/recreating them constantly at this stage. Didn't want to risk affecting anything already there if I could easily avoid it.
* It feels a bit jarring to me to have an existing SG named '...rabbitmq...' give access to '...amazonmq...' as well
* At the point where we've done the live migration, it seems cleaner to destroy all the rabbitmq stuff and delete the `app-rabbitmq` project entirely rather than move and rename an SG as part of that.

That being said, this is far more your domain than mine, and I'm happy to take your steer if you still think that's the right thing to do?

terraform/projects/app-amazonmq/main.tf
aldavidson: done

terraform/projects/app-amazonmq/main.tf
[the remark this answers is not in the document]
aldavidson: What about using AWS Secrets Manager? If that's how we're managing secrets on the new K8s platform, could we use that here too?

terraform/projects/app-amazonmq/README.md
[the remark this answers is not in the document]
kevindew: Ah yes I can see the logic you followed. I think that one was given the name as that's what the machine class of puppet is named, so it corresponds with: https://github.com/alphagov/govuk-puppet/blob/main/modules/govuk/manifests/node/s_rabbitmq.pp

I think that was optimistic naming back then since that isn't the only rabbitmq cluster in the stack.

terraform/projects/app-amazonmq/main.tf
kevindew: Sure, I was wondering if this was information that exists in the JSON config of https://github.com/alphagov/govuk-aws-data/pull/1127 and something that might need to be kept in sync between them.

terraform/projects/app-amazonmq/main.tf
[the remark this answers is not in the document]
kevindew: Great stuff, yeah it just caught me eye as we had the office_ips variable but didn't seem to use it.

terraform/projects/app-amazonmq/variables.tf
kevindew: Good stuff

terraform/projects/app-amazonmq/main.tf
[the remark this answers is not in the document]
aldavidson: I've got rid of the dedicated SG and added rules to the existing SG, as you suggested

terraform/projects/app-amazonmq/main.tf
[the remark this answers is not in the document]
aldavidson: I've put this back to only allowing HTTPS ingress from anything in the management SG (effectively, all EC2 instances). This still allows connection from a dev laptop via the remote port-forward

terraform/projects/app-amazonmq/main.tf
aldavidson: No, as far as I can tell this name is purely for description

terraform/projects/app-amazonmq/README.md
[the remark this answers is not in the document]
aldavidson: yep, will change the name

terraform/projects/app-amazonmq/main.tf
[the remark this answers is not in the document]
aldavidson: .....and I've also now tested that publishing-api on integration can connect to the broker over AMQPS

terraform/projects/app-amazonmq/main.tf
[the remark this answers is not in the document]
sengi: We'd only want to use Secrets Manager if this were a secret that we had to set manually, i.e. if it can't be automated. Basically, first preference is to eliminate a secret entirely, second preference is to generate it automatically, last resort is to use Secrets Manager.

terraform/projects/app-amazonmq/main.tf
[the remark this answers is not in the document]
sengi: Awesome, thanks!

terraform/projects/app-amazonmq/outputs.tf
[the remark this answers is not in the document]
sengi: no probs 👍

terraform/projects/app-amazonmq/README.md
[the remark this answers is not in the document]
aldavidson: done :+1:

terraform/projects/app-amazonmq/main.tf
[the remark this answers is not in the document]
aldavidson: >  Authorised people can look up the password using Terraform if/when they need it.
Could you slack me at some point with details of how to do this? I'd be keen to eliminate stored secrets if I can make this work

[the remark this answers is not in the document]
sengi: Just to summarise our Slack convo just now: basically folks can do something along the lines of `tf init -backend-config integration.govuk.backend && tf state show `random_password.amazonmq_root` (or whatever we end up calling it).

jenkins.sh
```diff
@@ -50,6 +50,7 @@ git clone --single-branch --branch "$GOVUK_AWS_DATA_BRANCH" git@github.com:alpha
 case $COMMAND in
   'apply') EXTRA='-auto-approve';;
   'plan (destroy)') COMMAND='plan'; EXTRA='-detailed-exitcode -destroy';;
+  # The -force flag has been deprecated since TF v1.0, supplying it causes errors in TF v1.0+
```
ChrisBAshton, line 53: This comment seems out of date now, I'm guessing it was added prior to the change to `-auto-approve` below?

aldavidson: True, I've updated the comment

terraform/projects/app-amazonmq/main.tf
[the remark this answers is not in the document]
aldavidson: I've got this working, [#afc5b9b](https://github.com/alphagov/govuk-aws/pull/1655/commits/afc5b9bf6d20481e78edbd0f5fd37cda331d9859) removes all dependence on sops-encrypted secrets and generates all passwords. Thanks for the idea!

PR #1658, opened 2022-11-28
Add API POST of full RabbitMQ definitions to AmazonMQ creation

aldavidson:
When creating an AmazonMQ RabbitMQ broker, only one user is supported. All other users, plus all the other configuration such as exchanges, permissions and queues, must be created via the web management UI, or the [RabbitMQ REST API](https://pulse.mozilla.org/api/).

This PR adds a `null_resource` which POSTS the [complete JSON definitions](https://github.com/alphagov/govuk-aws-data/pull/1127) (exported from the existing self-hosted RabbitMQ broker) to the REST API via curl, which will recreate the entirety of our existing RabbitMQ config in AmazonMQ. 

It's probably best to review and merge PR #1655 , which creates all the resources, first


aldavidson:
Closing without merging, superceded by #1661

PR #1661, opened 2022-12-02
Post RabbitMQ config to AmazonMQ API  

aldavidson:
POST RabbitMQ config as JSON, via a plain-text template file into which user passwords from govuk-aws-data can be interpolated at runtime

Supercedes #1658 (which I will close unmerged if and when this PR works)


aldavidson:
Closing without merging, superceded by subsequent work on #1655

PR #1668, opened 2022-12-08
Create a bucket for Athena Query results

kevindew:
This creates an S3 bucket that is intended for the storage of Athena query results conducted via the AWS web UI. It is has been done to reduce the frustration of developers taking an adhoc approach and setting a new bucket each time they run an Athena query.

This bucket is to be configured as the query result destination for the default workgroup, primary. This bucket will have to be connected to that workgroup via clickops as I can't find a good way to edit that automatically created workgroup, and it seems low value in using terraform to create a new workgroup that isn't the default (as I expect every user would forget to change it).

The clickops process is to go to Athena, select workgroups from the right bar (under administration). Then go into Query result configuration and select this bucket as the Location of query results.

I've already configured this in the test environment, you can experience it if you log into the test environment and then go to: https://eu-west-1.console.aws.amazon.com/athena/home?region=eu-west-1


sengi (commented):
I would just define the bucket in the existing `infra-fastly-logs` module, since it'd be more discoverable and more likely to be kept up to date, as well as being quite a good fit with what's already in there.

kevindew:
> I would just define the bucket in the existing `infra-fastly-logs` module, since it'd be more discoverable and more likely to be kept up to date, as well as being quite a good fit with what's already in there.

Thanks for taking a look. I'd agree if infra-fastly-logs was the only usage of athena but there's more, the new csp reporter also uses it.

sengi (approved):
Ah ok. I see where you're coming from then. I'd still just pick one or the other module for this to live in for now tbh, since it's an extreme amount of maintenance overhead just to define an S3 bucket - but up to you.

sengi:
Don't worry - it's gonna need to be cleaned up within a few months at most either way. I'm not really seeing a practical (as opposed to theoretical) advantage to making a separate root module for this but it's gonna be short-lived so I'm honestly fine with either.

PR #1669, opened 2022-12-09
Filter out commonly reported blocked URIs from CSP reporter

kevindew:
Trello: https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp

This is to reduce the noise in the Content Security Policy reports. We
get a lot of reports related to things we don't have control over, such
as browser extensions or apps that inject scripts into browsers. A lot
of these are hosting providers for fonts.

This is just an effort to filter out the most prolific of the reports
and does not attempt to remove all non-reports. The theory is that if we
reduce the most common false-positives then an actual positive will be
quite evident from it's volume in comparison to other reports.


KludgeKML (approved):
LGTM

PR #1671, opened 2022-12-12
Allow GOV.UK's EKS NAT IPs to bypass rate limiting

sihugh:
We already allow this for the EC2 environments' NAT Gateway IPs. This allows the replatformed environment to behave in the same way.


MahmudH (approved):
👍🏽

PR #1674, opened 2022-12-14
Apply rate limit

sihugh:
There are three steps to this:
- add a count only rate limit to allow us to monitor what happens at lower rate limits prior to reducing the blocking limit
- configure the existing rate limit to `block` instead of count and return 429, Cache-Control and Retry-After headers
- add a basic HTML response telling people what has happened should they fall foul of this limit

In future we may wish to do something better with the error page, such as writing VCL to retrieve a rendered one from S3, but I think this is OK for now. Based on the history of the limit we have set to `count` at present, it's only robots that will be affected.

https://trello.com/c/1B7DgFvt/3038-add-custom-response-headers-for-the-cache-public-waf


sihugh:
Corresponds to https://github.com/alphagov/govuk-aws-data/pull/1138

MahmudH (approved):
👍🏽

PR #1675, opened 2022-12-15
Make our amazonmq exchanges durable

aldavidson:
The exchanges auto-created on the AmazonMQ broker are not durable by default. This means if any of the underlying EC2 instances or the RabbitMQ processes running on them are restarted, the exchange will disappear.

This PR adds the `durable: true` flag to the exchange definitions.


BeckaL (approved):
👍

PR #1676, opened 2022-12-19
Use variables for RabbitMQ auto-scaling group sizes

aldavidson:
Part of the [Trial AmazonMQ deployment process on integration](https://trello.com/c/oqsXryjz/389-update-monitoring-alerting-from-rabbitmq-amazonmq) epic. 

Nothing is publishing to or consuming from the self-hosted RabbitMQ on integration, we've [repointed all apps over to AmazonMQ](https://github.com/alphagov/govuk-puppet/pull/11930) now. However, the RabbitMQ instances are now (rightly) triggering Icinga alerts because they have no consumers connected, and other related alerts.

This PR updates the RabbitMQ auto-scaling group definition to use parameters for min, max, and desired size,, which are defined in [this govuk-aws-data PR](https://github.com/alphagov/govuk-aws-data/pull/1141)  to be 0 on integration only, and 3 everywhere else.  

We've tested (by deploying & applying this branch on integration) that this gets rid of almost all of the remaining alerts

### DIFF
--- a/terraform/projects/app-rabbitmq/README.md
+++ b/terraform/projects/app-rabbitmq/README.md
@@ -48,6 +48,9 @@ Rabbitmq cluster
 
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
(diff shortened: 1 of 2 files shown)


BeckaL (approved):
Looks sensible to me, could you squash the commits before you merge?

PR #1677, opened 2022-12-23
Update AWS Glue to include client_ja3 data for www and assets

ChrisBAshton:
See https://github.com/alphagov/govuk-cdn-config/pull/438

We're going to start logging the client JA3 fingerprint of each request.
We first need to edit the Fastly UI to start logging that, and then this PR makes the corresponding change in our AWS/S3 logging logic.

Trello: https://trello.com/c/LVGgeLG7/

### DIFF
--- a/terraform/projects/infra-fastly-logs/main.tf
+++ b/terraform/projects/infra-fastly-logs/main.tf
@@ -183,7 +183,7 @@ resource "aws_glue_catalog_table" "govuk_www" {
       name = "ser_de_name"
 
       parameters {
-        paths                 = "client_ip,request_received,request_received_offset,method,url,status,request_time,time_to_generate_response,bytes,content_type,user_agent,fastly_backend,data_centre,cache_hit,cache_response,tls_client_protocol,tls_client_cipher"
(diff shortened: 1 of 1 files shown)


terraform/projects/infra-fastly-logs/main.tf
```diff
@@ -370,7 +375,7 @@ resource "aws_glue_catalog_table" "govuk_assets" {
       name = "ser_de_name"
 
       parameters {
-        paths                 = "client_ip,request_received,request_received_offset,method,url,status,request_time,time_to_generate_response,bytes,content_type,user_agent,fastly_backend,data_centre,cache_hit,cache_response,tls_client_protocol,tls_client_cipher"
+        paths                 = "client_ip,request_received,request_received_offset,method,url,status,request_time,time_to_generate_response,bytes,content_type,user_agent,fastly_backend,data_centre,cache_hit,cache_response,tls_client_protocol"
```
kevindew, line 378: Looks like this has lost a column rather than gained one

ChrisBAshton: Argh, good spot. I got sidetracked mid-change and then must have lost my place!

kevindew (approved):
Looks good to me 👍 

I've got a feeling we've only actually updated this once before and it broke (hence the docs and an icinga check) so be good to be careful to check if anything breaks when putting this out.

PR #1678, opened 2022-12-23
Bump Terraform version

ChrisBAshton:
The `plan`/`apply` worked in Integration and Staging, but failed on Production with:

```
Error:
Terraform doesn't allow running any operations against a state
that was written by a future Terraform version. The state is
reporting it is written by Terraform '0.12.30'

A newer version of Terraform is required to make changes to the current
workspace.
```

Trello: https://trello.com/c/LVGgeLG7


ChrisBAshton:
No longer needed - terraform version issue was because someone had tried deploying a branch to production. We've reverted those changes and now everything applied fine.

PR #1679, opened 2022-12-29
Migrate lambda functions

catalinailie:
We are planning to retire "govuk-lambda-app-deployment" repo.
In order to do that, we need to migrate the lambdas that are still used, into govuk-aws.


terraform/projects/infra-fastly-logs/main.tf
```diff
@@ -918,3 +918,101 @@ data "aws_iam_policy_document" "lambda_deployment_packages_bucket_access" {
     }
   }
 }
+
+# Grant permissions for acessing the S3 bucket
```
ChrisBAshton, line 923:
```suggestion
# Grant permissions for accessing the S3 bucket
```

terraform/lambda/SendPublicAPIEventsToGA/send_public_api_events_to_ga.py
```diff
@@ -6,12 +6,13 @@
...
 import urllib
+import boto3
+
+S3 = boto3.client('s3')
 
 EVENT_REGEX = re.compile(r'^ObjectCreated:')
 PUBLIC_API_LOG_REGEX = re.compile(r'^public_api_logs/')
 
-import boto3
-S3 = boto3.client('s3')
```
ChrisBAshton: This change seems unrelated to the `Add configuration for send_public_events_to_ga function` commit?

terraform/projects/infra-fastly-logs/main.tf
```diff
@@ -954,9 +954,35 @@ resource "aws_lambda_function" "download_logs_analytics" {
...
+      BUCKET_NAME = "${data.aws_s3_bucket.govuk-analytics-logs-production.bucket}"
+    }
+  }
+}
+
+# Upload local build deployment packages in lambda_deployment_packages bucket
+resource "aws_s3_bucket_object" "send_public_events_deployment_package" {
+  bucket = "${aws_s3_bucket.lambda_deployment_packages.id}"
+  key    = "send_public_api_events.zip"
+  source = "${path.module}/../../lambda/SendPublicAPIEventsToGA/send_public_events_to_ga.zip"
```
ChrisBAshton, line 981: Worth expanding on the commit message a bit, I initially thought the `terraform/lambda/SendPublicAPIEventsToGA/send_public_events_to_ga.zip` file being committed was a mistake, but now I see it was deliberate.

EDIT: looks like it _was_ a mistake, as I can see you generate all the zips in the last commit (agree that is neater). So worth going back to the `Add configuration for send_public_events_to_ga function` commit and removing the zip file.

terraform/projects/infra-fastly-logs/main.tf
```diff
@@ -918,3 +918,101 @@ data "aws_iam_policy_document" "lambda_deployment_packages_bucket_access" {
...
+  policy = data.aws_iam_policy_document.download_logs_analytics.json
+}
+
+resource "aws_lambda_permission" "allow_download_logs_analytics" {
+  statement_id  = "AllowExecutionFromS3Bucket"
+  action        = "lambda:InvokeFunction"
+  function_name = aws_lambda_function.download_logs_analytics.function_name
+  principal     = "s3.amazonaws.com"
+  source_arn    = data.aws_s3_bucket.govuk-analytics-logs-production.arn
+}
```
ChrisBAshton, line 1069: This all gets deleted in the `Add lambda trigger when an S3 object is created` commit - feels like some commit squashing is needed?

ChrisBAshton (commented):
Hey @catalinailie - nice one persevering with this, and coming up with a clean, pragmatic solution 👍 I've given it a surface level review (not looked too closely at the actual Terraform, just trusting that this is what you've arrived at through experimentation and the end result works). Also obviously haven't looked at the lambda code itself at all, as I assume this was just copied from the other repo.

A few comments below, hopefully we're close to merging! ⭐

terraform/projects/infra-fastly-logs/main.tf
```diff
@@ -922,30 +922,30 @@ data "aws_iam_policy_document" "lambda_deployment_packages_bucket_access" {
...
 
 # Upload local build deployment packages in lambda_deployment_packages bucket
 resource "aws_s3_bucket_object" "download_logs_analytics_deployment_package" {
-  bucket = aws_s3_bucket.lambda_deployment_packages.id
+  bucket = "${aws_s3_bucket.lambda_deployment_packages.id}"
   key    = "download_logs_analytics.zip"
   source = "${path.module}/../../lambda/DownloadLogsAnalytics/download_logs_analytics.zip"
 
-  etag = filemd5("${path.module}/../../lambda/DownloadLogsAnalytics/download_logs_analytics.zip")
+  etag = "${filemd5("${path.module}/../../lambda/DownloadLogsAnalytics/download_logs_analytics.zip")}"
```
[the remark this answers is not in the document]
catalinailie: Yes, this shouldn't be here. I mismatched things a bit after a rebase.

ChrisBAshton (approved):
Great stuff - much easier to follow!

Only nitpick: you could add the executable file permissions to the `.sh` files at the commit where they're added. But not a blocker.

PR #1681, opened 2022-12-30
Include original policy in CSP logged data

kevindew:
Trello: https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp

Note, I've tried this out and it is operational. It is currently applied to integration.

---

The original policy is the CSP policy that is an effect when the
violation occurs. Initially I didn't include this as data stored as I
thought it would effectively be a constant - just the same data in every
report, as our policies have very low variances between URLs.

I've since found out that it seems to be common for ad-blockers to
rewrite policies and thus what might be in effect is not what we've set.
Therefore I actually think it could be valuable to store this
information. It may help with some of the more mysterious reports.

On initial testing I found that adding a column amongst existing columns
caused problems. The terraform plan didn't consider the column new and
instead edited existing columns until adding a new one later in the
process. Worse, when I ran the ran the terraform and queried the database
it showed existing data in the wrong column. Either running a Glue
crawler run didn't help. Therefore I concluded the only safe way to add
a new column is to add it to the end. I'm not sure exactly why this is the
case, as the ORC columnar file format maps columns to field names, so 🤷

### DIFF
--- a/terraform/lambda/CspReportsToFirehose/index.mjs
+++ b/terraform/lambda/CspReportsToFirehose/index.mjs
@@ -63,6 +63,7 @@ function buildReport (json, sourceIp, userAgent) {
     sample: normaliseInputString(cspReport['script-sample']),
     line_number: normaliseInteger(cspReport['line-number']),
     status_code: normaliseInteger(cspReport['status-code']),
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
LGTM

PR #1683, opened 2023-01-04
Update timeout for the analytics lambda functions

catalinailie:
The default timeout is 3s, but this is not enough to run the functions.
The original ones have 2 min for DownloadLogsAnalytics and 10s for SendPublicAPIEventsToGA.

SendPublicAPIEventsToGA:
Screenshot 2023-01-04 at 15 07 43

DownloadLogsAnalytics:
Screenshot 2023-01-04 at 15 06 12


ChrisBAshton (approved):
Can you link to the original commits showing the times?

catalinailie:
> Can you link to the original commits showing the times?

This was configured directly in aws console. I attached screenshots with the configurations for each.

PR #1687, opened 2023-01-06
Add a backend specific WAF

sihugh:
This is mostly copied from the cache_public WAF.

We retain the x-always-block rule. This is the easiest way for Smokey to determine that a WAF is applied.

We retain the rule allowing traffic from GOV.UK's infrastructure. Smokey is the heaviest user here.

We do not need to specifically allow Fastly health checks, as they're not running against this resource.

We don't need to allow traffic from an external set of IPs, as we don't have that requirement here yet.

We keep a warning/count rate limit rule.

We'll set the "blocking" rule to count for now while we confirm that the rest works as expected.

For the rules we retain we do not need to parse the True-Client-IP header to identify the IP of the client. We have therefore switched aggregate key types from "FORWARDED_IP" (removing some of the additional configuration that this required) to simply "IP". The different rules have slightly different ways of configuring this:

- [rate based statement](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl.html#rate-based-statement)
- [IP set reference statement](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl.html#ip-set-reference-statement)

### DIFF
--- a/terraform/projects/infra-public-wafs/README.md
+++ b/terraform/projects/infra-public-wafs/README.md
@@ -20,6 +20,7 @@ No modules.
 
 | Name | Type |
 |------|------|
+| [aws_cloudwatch_log_group.public_backend_waf](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_group) | resource |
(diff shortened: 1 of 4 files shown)


MahmudH (approved):
👍🏽

PR #1689, opened 2023-01-18
WIP: Upgrade nodejs versions

ChrisBAshton:
This was first attempted in #1652 but had to be reverted in #1688, as the Node version was incompatible with the Terraform version.

This PR is a second attempt at upgrading the NodeJS versions.


ChrisBAshton:
Turns out bump to TF 1.1 is not needed. Superseded by #1698.

PR #1690, opened 2023-01-18
Update infra-mirror-bucket Terraform version from 0.11.15 to 0.13.6

ChrisBAshton:
This is blocking #1689 - we cannot upgrade the lambda's Node version without upgrading the AWS terraform provider, and we can't do that without upgrading Terraform itself to at least v1.1, and we can't jump straight to 1.1 without doing the 'major' upgrade to 0.13 first. Hence this PR.

Trello: https://trello.com/c/PMsAXSoT/3029-review-govuks-usage-of-node-2


ChrisBAshton:
I've split this PR into two: #1691 and #1692

PR #1693, opened 2023-01-19
Reference different certificate for Cloudfront Production

ChrisBAshton:
The assets Cloudfront distribution was referencing the (expired)
cert in us-east-1, for `*.publishing.service.gov.uk` - which had
been manually created, thus not auto-renewed.
https://us-east-1.console.aws.amazon.com/acm/home?region=us-east-1#/certificates/108d0f34-cf9f-4e04-81a2-231e42824b08

Instead we should be using the certificate that's created and
managed in Amazon Certificate Manager (ACM), in the eu-west-1
region:
https://eu-west-1.console.aws.amazon.com/acm/home?region=eu-west-1#/certificates/063a18cc-d4bf-4eaf-b1c1-41e86bbdd69d

That cert covers a number of domains, including the necessary
`*.publishing.service.gov.uk`.

However, we can't simply change the [aws_cloudfront_certificate](https://github.com/alphagov/govuk-aws/blob/18f25c90e6f14208ada49a452d8e726e069ffb71/terraform/projects/infra-mirror-bucket/main.tf#L146)
from us-east-1 to eu-west-2, as we still need the _www_ Cloudfront
distribution to work, and _that_ certificate is in us-east-1!

Therefore we've had to create multiple variables, to support
the multiple AWS regions where our certificates live. NB: the
default values have been applied so that the `plan` continues to
work on Integration & Staging.

Fixing the certificate configs will allow us to apply the changes
from #1691 to Production.

Successful plan on production: https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4756/console
Successful plan on integration:
https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4758/console

Trello: https://trello.com/c/PMsAXSoT/3029-review-govuks-usage-of-node-2

### DIFF
--- a/terraform/projects/infra-mirror-bucket/README.md
+++ b/terraform/projects/infra-mirror-bucket/README.md
@@ -14,6 +14,8 @@ The primary bucket should be in London and the backup in Ireland.
 | <a name="requirement_aws"></a> [aws](#requirement\_aws) | 2.46.0 |
 | <a name="requirement_aws"></a> [aws](#requirement\_aws) | 2.46.0 |
 | <a name="requirement_aws"></a> [aws](#requirement\_aws) | 2.46.0 |
+| <a name="requirement_aws"></a> [aws](#requirement\_aws) | 2.46.0 |
+| <a name="requirement_aws"></a> [aws](#requirement\_aws) | 2.46.0 |
 | <a name="requirement_fastly"></a> [fastly](#requirement\_fastly) | ~> 0.26.0 |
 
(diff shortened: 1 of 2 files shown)


sengi (approved):

sengi:
Whoops sorry didn't spot the region there. Has to be `us-east-1` for CloudFront because that's where its control plane runs :/

PR #1694, opened 2023-01-19
Add NLB in front of integration AmazonMQ

aldavidson:
After reviewing the migration process on integration, we decided to [add a network load balancer in front of the single-instance broker in integration](https://trello.com/c/owD4R2Fv/430-include-aws-network-load-balancer-in-front-of-integration-amazonmq), and repoint the DNS entry to that balancer. 

This is the recommended architecture for the highly-available cluster deployments we'll have in staging and production (see [this AWS blog post](https://aws.amazon.com/blogs/compute/creating-static-custom-domain-endpoints-with-amazon-mq/)), and adding one in integration has allowed us to work through a number of complications that this approach introduces.

It's been [applied successfully on integration](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4752/console)


terraform/projects/app-publishing-amazonmq/main.tf
```diff
@@ -79,8 +79,30 @@ locals {
     aws_environment = var.aws_environment
   }
 }
+# --------------------------------------------------------------
+# Lookups to existing resources which we need to reference
+
+data "aws_subnet" "lb_subnets" {
```
sengi, line 85: It's unusual to have a data resource looking up a resource that elsewhere in the same module is being referred to via remote state. I'm guessing it's needed here because our `infra-networking` module is only exporting the subnet ids and not the addresses? If so then fair enough.

terraform/projects/app-publishing-amazonmq/main.tf
```diff
@@ -79,8 +79,30 @@ locals {
...
+# Look up the existing SSL certificate for internal-facing infra
+data "aws_acm_certificate" "internal_cert" {
+  domain   = var.elb_internal_certname
+  statuses = ["ISSUED"]
+}
+
+# The ip_address attributes of the AmazonMQ instances are blank in the outputs
+# (see https://stackoverflow.com/a/69221987)
+# So the only way we can get the IPs in order to add them to the 
+# Load Balancer target group, is by a DNS lookup
```
sengi, line 99: Wow 😕 Prob worth raising this with Amazon if you haven't already.

sengi (approved):

terraform/projects/app-publishing-amazonmq/main.tf
aldavidson: Agreed, it's a bit awkward, but you're right - it looks like the CIDR blocks are only defined in govuk-aws-data as input variables to the `infra-networking` module, and not exposed as outputs. That makes a lookup via ID the least-worst option I could think of for retrieving them in a different project.

terraform/projects/app-publishing-amazonmq/main.tf
aldavidson: Good idea, thanks. I'll hang on until I've tried it with a HA cluster on staging though, so I can make a better-scoped report of the issue

PR #1701, opened 2023-01-26
Configure AmazonMQ via Lambda functions

aldavidson:
The previous method of performing GOV.UK-specific setup on a newly-created AmazonMQ broker was via a HTTP(s) POST from Jenkins. This only worked when targeting the integration environment - the Jenkins instance in integration has access to the AWS APIs, but not to any resources in any other environment.

As [discussed](https://docs.google.com/document/d/1dGqsopOg_aG1YGJGhNLHiu8QG6kE4127W7P3vMpLSkU/edit), the best way round this quirk is by creating Lambda functions in the target environments, which can POST the JSON config when invoked by Jenkins.

This PR adds the Lambda function and related roles/policies/security group rules. It's been [tested on integration](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4928/console), and also [on staging](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4922/console) (via a [merge with another branch](https://github.com/alphagov/govuk-aws/tree/add-staging-amazonmq-and-lambda) which creates the staging broker)


terraform/lambda/PostConfigToAmazonMQ/build_deployment_package.sh
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
```
sengi, line 1: There's [prior art](https://github.com/alphagov/govuk-aws/blob/1bc9826/terraform/projects/infra-artefact-bucket/main.tf#L314) for building the lambda zip using Terraform archive_file.

Also you can avoid all the pyenv/pip faff in return for foregoing `requests` (which is convenient but comes at the cost of needing a whole build pipeline here) and sticking with the Python standard library, which is turns out really [isn't so bad for this](https://stackoverflow.com/a/4188709).

I can pretty much guarantee otherwise that people will update the `.py` and forget to rebuild, or remember to rebuild but fail to check in the rebuilt binary, etc. etc.

terraform/lambda/PostConfigToAmazonMQ/post_config_to_amazonmq.py
```diff
@@ -0,0 +1,21 @@
...
+
+from requests.auth import HTTPBasicAuth
+
+def lambda_handler(event, context):
+    raw_json = base64.b64decode(event['json_b64'])
+    json_str = raw_json.decode('ascii')
+
+    print('Posting json to ', event['url'])
+    headers = {'Content-type': 'application/json'}
+    r = requests.post(event['url'], data=json_str, auth=HTTPBasicAuth(event['username'], event['password']), headers=headers)
```
sengi, line 13: Worth having some restriction on the domain name in `url`, just for defence-in-depth?

sengi (approved):
Cool - I think you can potentially avoid a bunch of build-related faff here though with a few small compromises.

terraform/lambda/PostConfigToAmazonMQ/post_config_to_amazonmq.py
aldavidson: The SG rule only allows egress to destinations within the same rabbitmq_access SG, so it's already protected to some degree. I could restrict it to, say, domains ending with 'mq.(region).amazonaws.com' , if you think that would be useful

terraform/lambda/PostConfigToAmazonMQ/build_deployment_package.sh
aldavidson: Agreed, it is a bit cumbersome and feels like a release process from the 1990s.
I'll have a go with `urllib` - if it proves trickier than expected I'll move it to a separate PR.

terraform/lambda/PostConfigToAmazonMQ/post_config_to_amazonmq.py
sengi: Yeah I was assuming there'd be SG/firewall rules as well but just seems like a good idea to eliminate any chance of this ending up being used as a general proxy.

I think all it'd need is something like:

```
url = event['url']
if !re.fullmatch(r'.*\.mq\.[a-z0-9_-]+\.amazonaws\.com', url):
    raise ValueError(f'url failed validation: {url}')
```

aldavidson: I've added this in 97ad6b8

terraform/lambda/PostConfigToAmazonMQ/build_deployment_package.sh
aldavidson: Done, in [f773bec](https://github.com/alphagov/govuk-aws/pull/1701/commits/f773bec82587af970660fea3942a2dbe87735a53)

PR #1705, opened 2023-02-07
Upgrade app rabbitmq terraform to 0.12

aldavidson:
While [migrating staging RabbitMQ to AmazonMQ](https://trello.com/c/8a87xtft/428-run-amazonmq-migration-process-on-staging), we were unable to `terraform apply` the `app-rabbitmq` project to delete the RabbitMQ EC2 instances at the end of the process. 

This was because some modules which the app-rabbitmq project depends on had been upgraded to use terraform 0.12 syntax, and the app-rabbitmq project still used v0.11. 

This PR updates app-rabbitmq to be v0.12-compatible. This also meant making a few tweaks to `modules/aws/node_group` to be v0.12-compatible. I've tried to find [constructs that may look a bit awkward, but should be compatible with 0.11 and 0.12](https://github.com/alphagov/govuk-aws/pull/1705/files#diff-e41a1dcc1994073b41bd36ab7eb14662b26fb3ece78742b9ac1b335cda74c32eR379)


terraform/modules/aws/node_group/main.tf
```diff
@@ -265,7 +265,9 @@ resource "aws_launch_configuration" "node_launch_configuration" {
   instance_type = "${var.instance_type}"
   user_data     = "${join("\n\n", list(file("${path.module}/${var.instance_user_data}"), var.instance_additional_user_data))}"
 
-  security_groups = ["${var.instance_security_group_ids}"]
+  # this awkward syntax should work in both v0.11 and v0.12
```
sengi, line 268: Why does it to be compatible with both?

terraform/modules/aws/node_group/main.tf
```diff
@@ -265,7 +265,9 @@ resource "aws_launch_configuration" "node_launch_configuration" {
...
+  # (see https://stackoverflow.com/questions/57117183/terraform-0-11-list-attribute-compatible-with-terraform-0-12)
+  security_groups = "${flatten(var.instance_security_group_ids)}"
```
sengi, line 270: Can these be changed to TF 0.12+ syntax or is there something preventing that?

e.g.
```suggestion
  security_groups = flatten(var.instance_security_group_ids)
```

sengi (approved):

terraform/modules/aws/node_group/main.tf
sengi: Ah is it sharing a values file with the old 0.11 stuff?

aldavidson: Caution, really - I know there's other terraform projects which use the `node_group` module, and I want to try and avoid forcing this upgrade on others if possible.

terraform/modules/aws/node_group/main.tf
sengi: sorry ignore me, makes sense now that I've read the description properly!

aldavidson: See the [previous comment](https://github.com/alphagov/govuk-aws/pull/1705#discussion_r1101243875) - there will be other projects using `node_group` which may still be using 0.11.

sengi: Makes total sense, sorry for the noise! I'll go have another coffee 🙈

PR #1707, opened 2023-02-07
Enable backend WAF

sihugh:
This puts the WAF into blocking mode so that requests to apps behind the govuk_backend_public ALB will have a rate limit applied.

The limits are applied here: https://github.com/alphagov/govuk-aws-data/blob/2120cc8ff00012acc30123fb28bba883ca976f6c/data/infra-public-wafs/production/common.tfvars#L2 and are for the preceding 5 minutes, measured every 30 seconds.

https://trello.com/c/DQkJXVPJ/2810-set-backend-waf-rate-limit-rule-into-blocking-mode


MahmudH (approved):
👍🏽

PR #1710, opened 2023-02-14
Add rate monitoring WAF for Bouncer

sihugh:
Bouncer lives behind a CDN, so this WAF is of a similar style to the [cache WAF](https://github.com/alphagov/govuk-aws/blob/main/terraform/projects/infra-public-wafs/cache_public_rule.tf) in the way it determines the original IP address.

It's stripped back to begin with so it only has the default x-always-block rule and a rule to count breaches of the rate limit.

This helps us determine whether we need to add any other rules to effectively manage legitimate traffic.

https://trello.com/c/z65OXqJ1/2845-bouncer-rate-limit-waf


MahmudH (approved):
LGTM 👍🏽

PR #1711, opened 2023-02-16
Add database_backups_access_list variable

MahmudH:
- Used for cross account access

Trello: https://trello.com/c/42WqbBy7/3082-lock-down-access-to-production-database-dumps-investigation


rtrinque (approved):
Looks good

PR #1714, opened 2023-02-28
Add rate limit to Bouncer

sihugh:
Bouncer accrues a lot of bots because of the number of domains it serves.

It's only fair that we give it a bit of protection.

The approach here is the same as the cache ALB:

  - grab the original IP from the true-client-ip header set by the CDN
  - return a 429 status and a Retry-After header
  - return a basic page for the unlikely human that trips the rate limit

I've also reset the priority numbers on other WAFs so they are sequential. Sometimes when applying Terraform, it seems like these have been reset internally by AWS, so this minimises diffs when doing a `plan`.

https://trello.com/c/z65OXqJ1/3099-bouncer-rate-limit-waf


MahmudH (approved):
👍🏽

PR #1716, opened 2023-03-02
Fix array list for account access

MahmudH:
- The access list is an array, therefore, creating a list from the identifiers will cause an array inside an array. This will fix this behaviour

### DIFF
--- a/terraform/projects/infra-database-backups-bucket/bucket_policy.tf
+++ b/terraform/projects/infra-database-backups-bucket/bucket_policy.tf
@@ -18,9 +18,7 @@ data "aws_iam_policy_document" "database_backups_cross_account_access" {
     principals = {
       type = "AWS"
 
-      identifiers = [
-        "${var.database_backups_access_list}",
-      ]
+      identifiers = "${var.database_backups_access_list}"
     }
 
     actions = ["s3:Get*", "s3:List*"]


terraform/projects/infra-database-backups-bucket/bucket_policy.tf
```diff
@@ -18,9 +18,7 @@ data "aws_iam_policy_document" "database_backups_cross_account_access" {
     principals = {
       type = "AWS"
 
-      identifiers = [
-        "${var.database_backups_access_list}",
-      ]
+      identifiers = "${var.database_backups_access_list}"
```
ChrisBAshton, line 21: Shouldn't it just be this?

```suggestion
      identifiers = var.database_backups_access_list
```

(Might not be possible in the current version of Terraform, but worth a `plan` to see. Otherwise it looks like we're setting a string, not an array)

MahmudH: Due to the formatting of the version, the change gives a linting error

ChrisBAshton (approved):
Cool 👍 I take it you've tried applying this branch and verified it fixes the issue?

If so, please merge!

ChrisBAshton:
This wasn't the issue.

PR #1717, opened 2023-03-03
Revert "Add database_backups_access_list variable"

ChrisBAshton:
Reverts alphagov/govuk-aws#1711

See https://github.com/alphagov/govuk-puppet/pull/12018

https://trello.com/c/42WqbBy7/3082-lock-down-access-to-production-database-dumps-investigation


MahmudH (approved):
👍🏽

PR #1719, opened 2023-03-18
Add EKS egress addresses to IP ACLs.

sengi:
- Allow traffic from EKS NAT gateways to S3 mirror buckets.
- In the WAF ratelimit exemptions, take the list of EKS NAT gateways from the new var in govuk-aws-data `data/common` instead of duplicating it in the module vars. (No diff from this, just avoid duplication.)

This fixes the Smokey probe failures for mirrors when running from the production EKS cluster.

Rollout: with alphagov/govuk-aws-data#1179. Already applied in production.


theseanything (approved):

martynipratt (approved):
LGTM

PR #1706, opened 2023-02-07
Upgrade app-draft-cache Terraform from 0.11.15 to 0.12.30

AgaDufrat:
## Context

It's not possible to plan/apply infrastructure changes in this project which uses terraform version 0.11 as it's relies on  `terraform/modules/aws/alarms/elb/main.tf`, which had its syntax updated to 0.12.
Running terraform plan resulted in 
```
Error: Error parsing /var/lib/jenkins/workspace/Deploy_Terraform_GOVUK_AWS/terraform/projects/app-draft-cache/main.tf:
At 285:37: unexpected token while parsing list: IDENT
```
https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/4965/console

## What
This updates the terraform version to 0.12.30 and applies formatting changes (with `terraform fmt` not sure if I need to reformat all the projects)

Tested by running `plan`. See [console output](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/5041/console)

PR #1721, opened 2023-03-27
0.12 upgrade

rtrinque:
A [PR merged on 13th Feb](https://github.com/alphagov/govuk-aws/pull/1706) bumped the root version of Terraform from 0.11 to 0.12. This applied the new constraint to all projects, which broke app-apt, but also probably many other projects.
This was an attempt to fix the problem introduced by the upgrade of the terraform/modules/aws/alarms/elb/main.tf to 0.12 syntax.

Since those two changes break a lot of projects that are anterior to 0.12, a decision was made to fix forward and upgrade all projects to 0.12, however due to time constraints and the facts that the current replatforming of govuk will make a lot of terraform projects obsolete (mainly app ones), priority was given to infra projects, this PR is concerned with upgrading only a subset of projects:
app-elasticsearch6
app-licensify-documentdb
app-shared-documentdb
infra-accounts-pre-digital-identity-backups-bucket
infra-artefact-bucket
infra-carrenza-vpn
infra-content-data-admin
infra-content-publisher
infra-database-backups-bucket
infra-datagovuk-organogram-bucket
infra-datagovuk-static-bucket
infra-env-sync-and-backup
infra-govuk-cdn-logs-monitor-bucket
infra-graphite-backups-bucket
infra-monitoring
infra-networking
infra-root-dns-zones
infra-security-groups
infra-specialist-publisher
infra-splunk
infra-stack-dns-zones
infra-support-api
infra-vpc

The following guidelines have been used: https://developer.hashicorp.com/terraform/language/v1.1.x/upgrade-guides/0-12#upgrading-to-terraform-v0-12
Companion PR at govuk-aws-data: https://github.com/alphagov/govuk-aws-data/pull/1184


MuriloDalRi (approved):
I can't see anything obviously wrong but this is a very long and difficult to review PR so approving assuming terraform plan works on all cases.

terraform/projects/infra-security-groups/main.tf
```diff
@@ -30,7 +30,6 @@ provider "fastly" {
 }
 
 provider "github" {
-  version = "~> 4.14.0"
```
ChrisBAshton, line 33: Yep, this looks like the approach recommended in the [docs](https://registry.terraform.io/providers/integrations/github/latest/docs) 👍

terraform/projects/infra-security-groups/apt.tf
```diff
@@ -97,7 +97,7 @@ resource "aws_security_group_rule" "apt-external-elb_ingress_fastly_http" {
   protocol  = "tcp"
 
   security_group_id = "${aws_security_group.apt_external_elb.id}"
-  cidr_blocks       = ["${data.fastly_ip_ranges.fastly.cidr_blocks}"]
```
ChrisBAshton, line 100: Technically these changes should be in the "Change the way we express list of list." commit, but don't bother if the rebase is going to be too painful.

terraform/modules/aws/alarms/natgateway/main.tf
```diff
@@ -50,7 +50,7 @@ variable "packetsdropcount_threshold" {
 #--------------------------------------------------------------
 resource "aws_cloudwatch_metric_alarm" "natgateway_errorportallocation" {
   count               = "${var.nat_gateway_ids_length}"
-  alarm_name          = "${var.name_prefix}-natgateway-errorportallocation-${element(var.nat_gateway_ids, count.index)}"
+  alarm_name          = "${var.name_prefix}-natgateway-errorportallocation-${var.nat_gateway_ids[0][0][count.index]}"
```
ChrisBAshton, line 53: Hmm, the `[0][0][count.index]` looks odd here - do we really have a triple nested array?
Reading the [`element` documentation](https://developer.hashicorp.com/terraform/language/functions/element) I'd expect the equivalent `element(var.nat_gateway_ids, count.index)` to be simply `var.nat_gateway_ids[count.index]` 🤔

terraform/projects/infra-root-dns-zones/main.tf
```diff
@@ -78,7 +78,7 @@ data "terraform_remote_state" "infra_vpc" {
 }
 
 resource "aws_route53_zone" "internal_zone" {
-  count = "${var.create_internal_zone}"
+  count = "${var.create_internal_zone ? 1 : 0}"
```
ChrisBAshton, line 81: LGTM - the `create_internal_zone` is [already a boolean](https://github.com/alphagov/govuk-aws-data/search?q=create_internal_zone), so nothing to change in govuk-aws-data.

terraform/projects/infra-database-backups-bucket/bucket_policy.tf
```diff
@@ -18,7 +18,12 @@ data "aws_iam_policy_document" "database_backups_cross_account_access" {
     principals {
       type = "AWS"
 
-      identifiers = var.database_backups_access_list
+      identifiers = [
```
ChrisBAshton, line 21: Ideally you'd rebase against `main` rather than have this extra merge commit.

ChrisBAshton (commented):
LGTM - nice one persevering with this ⭐ 

A couple of comments below but nothing blocking.

richardTowers:
Does this mean that all the `app-*` projects that we haven't upgraded won't be appliable from `main` anymore?

That's maybe okay (since most of them are just awaiting decommissioning), but we should think a bit about how we'd run it if we needed to (e.g. from a non-main branch). I guess it will be convenient for us to be able to terraform destroy a lot of these projects at some point.

terraform/modules/aws/alarms/natgateway/main.tf
rtrinque: Yeah this one looks a bit weird, I'll have another look at it

rtrinque:
> Does this mean that all the `app-*` projects that we haven't upgraded won't be appliable from `main` anymore?
> 
> That's maybe okay (since most of them are just awaiting decommissioning), but we should think a bit about how we'd run it if we needed to (e.g. from a non-main branch). I guess it will be convenient for us to be able to terraform destroy a lot of these projects at some point.

That is correct, a lot of projects (most of app-) are broken atm following the syntax upgrade of terraform/modules/aws/alarms/elb/main.tf
Note that they are already currently broken, the idea was to fix forward by bringing everything to 0.12, we've just stopped short of all the app- projects since the effort seems less and less worthwhile given how close we were getting to the new platform.
Work is nearly finished for app- projects as well though, so we can resume this later if we feel the need to use the app-projects

richardTowers:
👍🏻  cool, that's fine. If the worst comes to the worst we can always find some past commit that still works.

I'm not approving because I haven't reviewed, but don't let my comments block a merge :shipit:

terraform/modules/aws/alarms/natgateway/main.tf
rtrinque: fixed in https://github.com/alphagov/govuk-aws/pull/1721/commits/46027bc790971c95e6ef5b94019799b811a94114

terraform/projects/infra-database-backups-bucket/bucket_policy.tf
rtrinque: rebase done

PR #1723, opened 2023-04-27
Upgrade nodejs to v18

MahmudH:
The current versions are nearing EOL and will no longer be supported. So this PR upgrades nodejs from v16 to v18

Trello: https://trello.com/c/SOmvaQgr/3135-upgrade-nodejs-from-v16-to-v18-3


deborahchua (approved):
LGTM 👍

PR #1724, opened 2023-05-02
Remove disused load balancers.

sengi:
These are no longer in use since the Kubernetes migration (and had been deleted, but it turns out this module is a bit more active than we thought so we need to get rid of them from here so they don't keep getting re-created).

The names that referred to these were deleted in https://github.com/alphagov/govuk-dns-tf/pull/17.

Tested: [`terraform plan` in integration](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/5199/console) (needs VPN, sorry)


theseanything (commented):
🎉

PR #1731, opened 2023-05-15
Remove disused app-rate-limit-redis module.

sengi:
This has been disused since the Replatforming launch (because we missed the configuration item that would have used it).

https://github.com/alphagov/feedback/pull/1644 reinstates the rate-limit feature without the separate Redis (saving USD 13,350 annually).

Rollout: already done (since nothing was using these Redis instances).


KludgeKML (approved):
LGTM

PR #1739, opened 2023-06-01
Replicate assets bucket from prod to non-prod envs.

sengi:


terraform/projects/infra-assets/backup.tf
```diff
@@ -11,44 +11,38 @@ resource "aws_s3_bucket_versioning" "assets_backup" {
...
-  template = file("backup_policy.tpl")
-
-  vars = {
-    src_bucket = aws_s3_bucket.assets.id
-    dst_bucket = aws_s3_bucket.assets_backup[0].id
+    actions = [
+      "s3:GetObjectVersionTagging",
+      "s3:Replicate*",
+    ]
+    resources = ["arn:aws:s3:::${aws_s3_bucket.assets_backup[0].id}/*"]
```
theseanything, line 33: Nit: v minor - might be worth referring the arn attribute directly, just for consistency with how we reference other resource types in IAM policies.

```suggestion
    resources = ["${aws_s3_bucket.assets_backup[0].arn}/*"]
```

terraform/projects/infra-assets/backup.tf
```diff
@@ -11,44 +11,38 @@ resource "aws_s3_bucket_versioning" "assets_backup" {
...
-    {
-      "Action": "sts:AssumeRole",
-      "Principal": { "Service": "s3.amazonaws.com" },
-      "Effect": "Allow",
-      "Sid": ""
+  statement {
+    sid = "S3RoleInSourceAccountReplicatesObjectsToDestinationBucket"
+    principals {
+      type        = "AWS"
+      identifiers = [local.replication_service_role_in_prod_account]
```
theseanything, line 27: Nit: Couple places you've got this - wondering if you can just reference the role arn directly instead of using a local?
```suggestion
      identifiers = [aws_iam_role.replication.arn]
```

terraform/projects/infra-assets/replication.tf
```diff
@@ -0,0 +1,174 @@
...
+    }
+    filter {}
+  }
+}
+
+resource "aws_iam_role" "replication" {
+  count              = var.aws_environment == "production" ? 1 : 0
+  name               = local.replication_role_name
+  description        = "Service role for S3 replication of govuk-assets-production to staging/integration and to govuk-assets-backup-production."
+  assume_role_policy = jsonencode(local.s3_can_assume_role)
```
theseanything, line 73: Nit: Might be easier to follow if we put the statement directly here instead of a local (which doesn't seem to be used elsewhere?)

theseanything (approved):

terraform/projects/infra-assets/backup.tf
sengi: Good catch, thanks! Done.

sengi (commented):
Thanks for the neat suggestions! 🙇 All done and checked no diff.

terraform/projects/infra-assets/replication.tf
sengi: Yeah was wondering about this one myself. (At one point I had two places that used it, otherwise wouldn't have bothered hoisting it out of there, but OTOH it perhaps adds a bit of self-documentation? Then again, maybe it just looks odd if you're used to seeing these inline.) 🤔

sengi: Tried reverting it and yeah I think the local isn't really adding much. Done!

PR #1742, opened 2023-07-18
Simplify SOPs and DocumentDB KMS key policies

theseanything:
This simplifies the key policy statements for the SOPs and DocumentDB KMS keys.  The top level statement "Enable IAM User Permission" delegates all access control to IAM policies. 
```json
{
    "Sid": "Enable IAM User Permissions",
    "Effect": "Allow",
    "Principal": {
        "AWS": "arn:aws:iam::<aws_account_id>:root"
    },
    "Action": "kms:*",
    "Resource": "*"
}
```
From [the AWS docs](https://docs.aws.amazon.com/kms/latest/developerguide/key-policy-overview.html):

>When the principal in a key policy statement is an [AWS account principal](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html#principal-accounts) expressed as `arn:aws:iam::111122223333:root`, the policy statement doesn't give permission to any IAM principal. Instead, it gives the AWS account permission to use IAM policies to delegate the permissions specified in the key policy. (A principal in `arn:aws:iam::111122223333:root` format does not represent the [AWS account root user](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_root-user.html), despite the use of "root" in the account identifier. However, the account principal represents the account and its administrators, including the account root user.)

This means the statements allowing administrator and user permissions to explicit IAM Roles had no net effect - especially as the IAM Roles already had IAM Permissions to access the keys.

This simplifies the key policy and allows us to no longer need to updated the list of IAM Roles. (Which was a pain, as we had to put in a hack to wait for the IAM Roles to propergate).

---

This also adds a statement to explicitly give permission to RDS to use the key. This statement probably also has no effect given the top "Enable IAM User Permission" statement, but added to make sure we don't accidentally remove permissions for DocumentDB to use the key.

### DIFF
--- a/terraform/projects/infra-security/main.tf
+++ b/terraform/projects/infra-security/main.tf
@@ -556,6 +556,60 @@ resource "time_sleep" "wait_30_seconds" {
   }
 }
 
+data "aws_iam_policy_document" "kms_docdb_policy" {
+  statement {
+    sid = "Delegate permissions to IAM policies"
+
+    actions = [
+      "kms:*",
+    ]
+
+    resources = ["*"]
+
+    principals {
+      type        = "AWS"
+      identifiers = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:root"]
+    }
+  }
+
+  statement {
+    sid = "Allow access through RDS for all principals in the account that are authorized to use RDS"
+
+    actions = [
+      "kms:Encrypt",
+      "kms:Decrypt",
+      "kms:ReEncrypt*",
+      "kms:GenerateDataKey*",
+      "kms:CreateGrant",
+      "kms:ListGrants",
+      "kms:DescribeKey"
+    ]
+
+    resources = ["*"]
+
+    principals {
+      type        = "AWS"
+      identifiers = ["*"]
+    }
+
+    condition {
+      test     = "StringEquals"
+      variable = "kms:CallerAccount"
+
+      values = [data.aws_caller_identity.current.account_id]
+    }
+
+    condition {
+      test     = "StringEquals"
+      variable = "kms:ViaService"
+
+      values = [
+        "rds.${var.aws_region}.amazonaws.com"
+      ]
+    }
+  }
+}
+
 resource "aws_kms_key" "sops" {
   description = "Encryption key for govuk-aws-data"
   policy      = data.aws_iam_policy_document.kms_sops_policy.json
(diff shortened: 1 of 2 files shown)


terraform/projects/infra-security/main.tf
```diff
@@ -472,87 +472,57 @@ data "aws_iam_policy_document" "kms_sops_policy" {
...
+      values = [data.aws_caller_identity.current.account_id]
+    }
 
-See issues:
-1. https://github.com/hashicorp/terraform-provider-aws/issues/245
-2. https://discuss.hashicorp.com/t/terraform-malformed-policy/11281/2
-*/
-resource "time_sleep" "wait_30_seconds" {
-  create_duration = "30s"
+    condition {
```
sengi, line 520: Presumably these two `Condition` blocks together produce something like this in the output policy?

```json
            "Condition": {
                "StringEquals": {
                    "kms:ViaService": "rds.eu-west-1.amazonaws.com",
                    "kms:CallerAccount": "<our account id>"
                }
            }
```

If so then that looks right to me, from comparing with the baked-in key policies on the AWS-managed keys for other RDS stuff and my reading of https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_multi-value-conditions.html#reference_policies_multiple-conditions-eval 😵‍💫

sengi (approved):
Awesome, thanks so much for chasing down this ugly bit of unnecessary complexity!

terraform/projects/infra-security/main.tf
```diff
@@ -472,87 +472,57 @@ data "aws_iam_policy_document" "kms_sops_policy" {
...
+      test     = "StringEquals"
+      variable = "kms:ViaService"
 
-  triggers = {
-    gds_admin_roles_and_arns = jsonencode([for role, arn in module.gds_role_admin.roles_and_arns : arn])
+      values = ["rds.${var.aws_region}.amazonaws.com"]
```
sengi, line 524: Does this (also?) need to match `docdb-elastic.*.amazonaws.com`?

theseanything: According to [this table](https://docs.aws.amazon.com/kms/latest/developerguide/conditions-kms.html#viaService_table) that's not a valid service to use with condition (but also think only applies to serverless opensearch clusters, rather than the instance based ones (which is what we use).

However the docs do say you need `es.*.amazonaws.com, aoss.*.amazonaws.com`, so will add those.

I had just used `rds.*.amazonaws.com` as that's what was in a default policy when tried creating new cluster with a new key.

PR #1749, opened 2023-08-09
Remove dead resources from `infra-public-wafs`

robinjam:
[Trello card](https://trello.com/c/IpC6eYPp/3179-create-waf-for-the-licensify-backend-public-load-balancer-3)

Extracted from #1747, to keep the changes atomic.

Removes some references to resources that no longer exist post-replatforming, which are currently blocking Terraform plans and applies for this project.


robinjam:
I'm actually not confident merging this at all - this project is totally out of sync with reality on staging.

sengi (approved):
LGTM — none of these load balancers have existed for several months now.

Whether the module will actually apply ok is another question entirely, but happy to help with that if needed :)

robinjam:
Thanks very much @sengi 🙏 

I'll pore over the apply output tomorrow and try to make sense of it, but might take you up on that offer

sengi:
Tried running plans in staging and integration and didn't see anything too surprising. (It's already mostly applied in integration btw.)

sengi:
Also feel free to just go for it in staging tomorrow morning and see if anything breaks. Literally what staging is for after all. I'll be around if you need help.

PR #1750, opened 2023-08-10
Clean up dangling ACL entries from Carrenza.

sengi:
We haven't had anything running in Carrenza/6degrees for years now.

Goes with https://github.com/alphagov/govuk-aws-data/pull/1253

Tested: applied in staging (see other PR for example plan output).


robinjam (approved):
Nice!

PR #1751, opened 2023-08-10
Fix InvalidParameterValue in security group rule.

sengi:
For Amazony reasons, the API for security groups has separate fields for v4 and v6 addresses/netblocks. This means stuff breaks when v6 addresses are (correctly) added to lists.

Thankfully the GitHub provider added fields to fetch the v4 and v6 netblocks separately in the meantime.

I've kept the version updates to a minimum here to simplify getting this in, but we should update to less ancient versions in the near future.

Tested: already applied in integration (and it's a no-op everywhere else apart from the minor version upgrades of Terraform and the GitHub provider).


robinjam (approved):
Nice one! Thank goodness GitHub is sensible at least

KludgeKML (approved):
Nice find!

ChrisBAshton (approved):
⭐

PR #1752, opened 2023-08-24
Add security.txt for data.gov.uk

sihugh:
Uses the [same approach as other gov.uk services](https://github.com/alphagov/govuk-cdn-config/pull/456/commits/aba1e7797fe2380d0e302f6d18d8fa310599a4a4)

During the vcl_recv subroutine, we check to see whether the request is for one of the usual options for security.txt. If so, we throw a custom error (805) which is then processed by the vcl_error subroutine.

vcl_error then catches the custom error code, changes it to a redirect and sets the correct headers.

The user's browser will then redirect to the Cabinet Office's vulnerability disclosure program security.txt.

https://trello.com/c/VLBjBTMe/313-put-securitytxt-on-bouncer-and-dgu

### DIFF
--- a/terraform/projects/fastly-datagovuk/datagovuk.vcl
+++ b/terraform/projects/fastly-datagovuk/datagovuk.vcl
@@ -21,6 +21,11 @@ sub vcl_recv {
     error 403 "Forbidden";
   }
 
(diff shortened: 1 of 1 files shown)


robinjam (approved):

robinjam:
This'll need a rebase but apart from that, LGTM

PR #1753, opened 2023-08-24
Improve DGU Fastly cache hit rate

sihugh:
Remove UTM query params as they have little effect on the backend.

Sort remaining query params so that the resulting URL has a greater chance of being cacheable/retrievable from cache.

Discard query params from the home page. The home page doesn't respond to query params, so there's little point in accepting them at the origin.


ChrisBAshton (approved):
LGTM 🎉

sengi (commented):
Thanks! We could probably tighten this up further if it helps.

PR #1755, opened 2023-09-11
Upgrade terraform version for Fastly DGU

deborahchua:
It's not possible to plan/apply infrastructure changes in this project as we are now using Terraform version 0.12. 

Running a Terraform `plan` threw this error:
```
Error: Unsupported Terraform Core version

This configuration does not support Terraform version 0.12.30. To proceed,
either choose another supported Terraform version or update the root module's
version constraint. Version constraints are normally set for good reason, so
updating the constraint may lead to other errors or unexpected behavior.
```

The new changes have been tested running a `plan` - see the [console output ](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/5403/console)

Trello card: https://trello.com/c/lN55wEjL/3289-finish-retiring-contracts-finder-archive-3

### DIFF
--- a/terraform/projects/fastly-datagovuk/main.tf
+++ b/terraform/projects/fastly-datagovuk/main.tf
@@ -48,7 +48,7 @@ variable "backend_domain" {
 # --------------------------------------------------------------
 terraform {
   backend "s3" {}
-  required_version = "= 0.11.15"
+  required_version = "= 0.12.30"
 }
 
 provider "fastly" {
--- a/terraform/projects/fastly-datagovuk/main.tf
+++ b/terraform/projects/fastly-datagovuk/main.tf
@@ -4,43 +4,43 @@
 * Manages the Fastly service for data.gov.uk
 */
 variable "aws_region" {
-  type        = "string"
+  type        = string
   description = "AWS region"
   default     = "eu-west-1"
 }
 
 variable "stackname" {
-  type        = "string"
+  type        = string
   description = "Stackname"
 }
 
 variable "aws_environment" {
-  type        = "string"
+  type        = string
   description = "AWS Environment"
 }
 
 variable "fastly_api_key" {
-  type        = "string"
+  type        = string
   description = "API key to authenticate with Fastly"
 }
 
 variable "logging_aws_access_key_id" {
-  type        = "string"
+  type        = string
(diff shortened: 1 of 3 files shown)


terraform/projects/fastly-datagovuk/main.tf
```diff
@@ -173,20 +173,7 @@ resource "fastly_service_v1" "datagovuk" {
...
-        "%U%q",
-        "%>s",
-        "%B",
-        "%%{tls.client.protocol}V",
-        "${fastly_info.state}V",
-        "%\"Referer\"i",
-        "%\"User-Agent\"i",
-      ]
-    )
+    format             = "%h\\t%%{%Y-%m-%d %H:%M:%S}t.%%{msec_frac}t\\t%m\\t%U%q\\t%>s\\t%B\\t%%{tls.client.protocol}V\\t%%{fastly_info.state}V\\t%%{Referer}i\\t%%{User-Agent}i"
```
ChrisBAshton, line 176: Interesting... I was stumped as to why reverting the refactor commit fixes the `A managed resource "fastly_info" "state" has not been declared in the root module.` error, as they both reference `fastly_info.state` in exactly the same way.

However, it looks like `fastly_info.state` is [provided by the Fastly Terraform provider](https://developer.fastly.com/reference/vcl/variables/miscellaneous/fastly-info-state/) and is "available in
all subroutines".

I imagine when `s3logging`'s `format` is evaluated, `fastly_info.state` is inside a subroutine, and it works. But I imagine that calling `join` requires that the value of `fastly_info.state` is known at compile time, and it's not available yet, so errors.

Reverting the refactor as you've done is fine, but it might be worth exploring whether you can [create a structure at the top of the file](https://stackoverflow.com/a/64202586) that references `fastly_info.state`, and you can then refer to that variable instead 🤔 ? See also https://saturncloud.io/blog/terraform-resolving-a-managed-resource-has-not-been-declared-in-the-root-module/

ChrisBAshton (approved):
Thanks - the VCL plan LGTM 👍  one minor comment below.

terraform/projects/fastly-datagovuk/main.tf
deborahchua: Thanks @ChrisBAshton , I'll look into this in a separate PR 👍

PR #1758, opened 2023-09-14
Block autodiscover

MahmudH:
There’s nothing at https://gov.uk/autodiscover/autodiscover.xml, therefore, requests made from Microsoft Outlook to this URI path should be blocked as it is currently done through fastly VCL. Our goal is to remove VCL code, this PR is part of that work.

Trello: https://trello.com/c/tprMplI9/3293-move-logic-for-serving-an-http-404-if-request-url-matches-autodiscover-autodiscoverxml-from-cdn-to-waf-3

### DIFF
--- a/terraform/projects/infra-public-wafs/cache_public_rule.tf
+++ b/terraform/projects/infra-public-wafs/cache_public_rule.tf
@@ -229,6 +229,38 @@ resource "aws_wafv2_web_acl" "cache_public" {
     metric_name                = "cache-public-web-acl"
     sampled_requests_enabled   = true
   }
+
+  rule {
+    name     = "block-autodiscover"
+    priority = 6
+    action {
+      block {
+        custom_response {
+          response_code = "404"
+        }
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


robinjam (approved):
LGTM!

robinjam (changes requested):
~~Just realised when reviewing [the other PR on this card](https://github.com/alphagov/govuk-fastly/pull/6) that `bouncer` uses a different load balancer and [ACL](https://github.com/alphagov/govuk-aws/blob/main/terraform/projects/infra-public-wafs/bouncer_public_rule.tf) than `www`. We'll need to add this same rule to that other ACL before we can remove the corresponding logic from Fastly.~~

~~(Sorry, should have noticed that when I wrote up the card 🤦‍♂️)~~

robinjam (approved):
Mahmud and I did some digging and discovered that the `bouncer` load balancer actually does use the `cache_public` ACL, so ignore the above.

LGTM 👍

PR #1764, opened 2023-09-27
Add licensify frontend waf rate limit

MahmudH:
This is mostly copied from the cache_public WAF.

We retain the x-always-block rule. This is the easiest way for Smokey to determine
that a WAF is applied.

We keep a warning/count rate limit rule.

We'll set the "blocking" rule to count for now while we confirm that the rest works as
expected.

For the rules we retain we do not need to parse the True-Client-IP header to
identify the IP of the client. We have therefore switched aggregate key types
from "FORWARDED_IP" (with some additional configuration) to "IP". The different
rules have slightly different ways of configuring this:

[rate based statement](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl.html#rate-based-statement)
[IP set reference statement](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl.html#ip-set-reference-statement)

Trello: https://trello.com/c/RkCyn4Ot/3178-create-waf-for-the-licensify-frontend-public-load-balancer-3

### DIFF
--- a/terraform/projects/infra-public-wafs/licensify_frontend_public_rule.tf
+++ b/terraform/projects/infra-public-wafs/licensify_frontend_public_rule.tf
@@ -1,0 +1,193 @@
+resource "aws_wafv2_web_acl" "licensify_frontend_public" {
+  name  = "licensify_frontend_public_web_acl"
+  scope = "REGIONAL"
+
+  default_action {
+    allow {}
+  }
+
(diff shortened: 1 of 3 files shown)


robinjam (approved):
LGTM 👍

Feel free to share the plan outputs before applying if you want a 2nd pair of eyes

PR #1765, opened 2023-09-28
Remove outdated warning

CristinaRO:
Quoth @aldavidson: "This is outdated and should be removed, as the TF was used in production 6 months ago".

[Trello](https://trello.com/c/lZp7YWgu/837-remove-the-do-not-use-in-production-yet-warning-from-the-amazonmq-terraform-projects-readme)


aldavidson (approved):
Thank you!

PR #1791, opened 2023-10-23
Remove old IAM policy from `infra-mirror-bucket`

robinjam:
The mirrorer has been moved to Kubernetes, and the old mirrorer app has been deleted from AWS. This is causing Terraform plans/appies for `infra-mirror-bucket` to fail, because it defines an IAM policy to allow the (now-deleted) mirrorer app to write to the mirror bucket.

This PR removes the redundant IAM policy.

The Terraform `plan` for this change is a no-op for all 3 environments.


theseanything (approved):
Thanks for catching this!

PR #1792, opened 2023-10-30
Rename office_ips to gds_egress_ips.

sengi:
This is a more accurate name as it always included networks other than just the office, for example the VPNs.

This is just a straight `fd |xargs gsed -i 's/office_ips/gds_egress_ips/g' && tools/update-docs.sh`.

Roll out together with https://github.com/alphagov/govuk-aws-data/pull/1266.


sengi:
Force-push to fix accidental inclusion of my performance hack to update-docs.sh (which doesn't work properly on CI for some reason).

samsimpson1 (approved):

PR #1801, opened 2023-11-20
Fix infra-datagovuk-organogram-bucket module.

sengi:
- Update to Terraform 1.x (via 0.13).
- Update to current versions of providers.
- Use [standard module structure](https://developer.hashicorp.com/terraform/language/modules/develop/structure).
- Use provider-default tags for consistent AWS tagging.
- Update deprecated syntax, resource config etc.

Should now be deployable from arm64.

Tested: applied in all three accounts, extra domains added to CORS config as expected.


kentsanggds (approved):
LGTM

PR #1807, opened 2023-11-30
Remove DownloadLogsAnalytics & SendPublicAPIEventsToGA Lambdas

deborahchua:
These lambda functions are still running on an old version of Python (3.7) which is now end of life.

Due to previous attempts at upgrading to Python 3.8 or newer (and failing) we have agreed to remove them because:

- the are not/under- used
- they will break when Universal Analytics sunsets in June 2024
- the Data Services team doesn't have the dev/engineer resource and AWS experience to fix and upgrade them themselves.

This is the [console output ](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Terraform_GOVUK_AWS/5523/console)for the Terraform `plan` run for `production`:

`Plan: 0 to add, 0 to change, 14 to destroy.`

Trello card: https://trello.com/c/sxjNj31E/3360-archive-sendpublicapieventstoga-and-downloadlogsanalytics-lambda-functions-2


robinjam (approved):
LGTM, nice one 👍

PR #1810, opened 2023-12-14
Rationalise ratelimit exemption rules.

sengi:
Make it clearer as to what the rule does. See:
- https://github.com/alphagov/govuk-aws-data/pull/1276


robinjam (approved):
Nice!

KludgeKML (approved):
Makes sense!

PR #1813, opened 2024-01-10
Move protocol column to the end of the list

richardTowers:
Terraform identifies these columns by their position in the list, so if we add a new one in the middle of the list it ends up renaming all the subsequent columns (so request_time gets renamed to be protocol, and so on), before adding a new column for the last one in the list.

This is not what we want, and it causes problems for Athena, which gets confused about the types of columns.

We can make our lives a lot easier by just adding the column to the end of the list, so terraform will just add the one we want.


sihugh (approved):
Curse you, Terraform! 

old man shakes fist at cloud

PR #1814, opened 2024-01-17
Add new NAT IPs addresses to data.gov.uk allow list

theseanything:
We've already replaced the NATs in staging so this removes the old IPs. For production, it adds the new IPs. The old IPs will be removed in a subsequent PR after the migration.


sengi (approved):
This ACL is basically defunct at this point, but fine to merge this anyway.

sengi:
The data.gov.uk Fastly config is [being moved to govuk-fastly](https://github.com/alphagov/govuk-fastly/pull/34) and we'll drop the ACL with that. (Cache purge requests are authenticated differently now and not based on trusting the network.)

theseanything:
Closing as doesn't matter

PR #1823, opened 2024-03-06
Remove obsolete app projects

samsimpson1:
The resources in these projects have either already been deleted or are on a list to be removed in the future.

alphagov/govuk-infrastructure#1127


terraform/projects/app-asset-master/main.tf
```diff
@@ -1,45 +0,0 @@
...
-  region  = var.aws_region
-  version = "2.46.0"
-}
-
-data "aws_route53_zone" "internal" {
-  name         = var.internal_zone_name
-  private_zone = true
-}
-
-resource "aws_efs_file_system" "assets-efs-fs" {
```
sengi, line 21: Is this the EFS filer that we're still using for temporarily storing uploaded attachments while they're scanned with Clam? Might need to remember to import it somewhere afterwards :)

terraform/projects/app-shared-documentdb/main.tf
```diff
@@ -1,166 +0,0 @@
...
-    value = var.profiler
-  }
-
-  parameter {
-    name  = "profiler_threshold_ms"
-    value = var.profiler_threshold_ms
-  }
-}
-
-resource "aws_docdb_cluster" "cluster" {
```
sengi, line 130: I think this one's still in use too so gonna need to import it somewhere at some point.

sengi (approved):
Awesome! Thanks so much for sorting through all this old cruft!

Spotted a couple of resources that I think we don't wanna delete (but still fine to delete the TF — we can just import those one or two things into govuk-infrastructure somewhere — just pointing 'em out so we don't delete the actual objects)

terraform/projects/app-asset-master/main.tf
samsimpson1: Ah yes, I thought we were using a new efs volume for this but turns out it's still using the old one. I'll put it on the list of things that need moving across

terraform/projects/app-shared-documentdb/main.tf
samsimpson1: Oh yeah, this needs adding to the list too. When I looked on the AWS dashboard earlier it wasn't showing any documentdb clusters. Odd

PR #1824, opened 2024-03-11
Remove remaining obsolete projects

samsimpson1:
https://github.com/alphagov/govuk-infrastructure/issues/1127


sengi (approved):
Awesome! Thanks so much for sorting through all this cruft!

PR #1827, opened 2024-04-22
Further cleanup in MQ module.

sengi:
- Use an up-to-date Python runtime for the config Lambda.
- Use `aws_iam_policy_document` for better ahead-of-time checking.
- Fix the HTTP error handling in the lambda that uploads the config to RabbitMQ.
- Fix a JSON syntax error in the config template which was causing RabbitMQ 3.11 to reject the config.
- Use a map for the random_password resources, since we were putting them in a map for the template anyway.
- Eliminate the `publishing_amazonmq_instance_count` parameter.
- Remove unused provider declaration.
- Remove redundant comments.

Tested: applied in the integration and staging accounts.

https://github.com/alphagov/govuk-infrastructure/issues/1267

### DIFF
--- a/terraform/lambda/PostConfigToAmazonMQ/post_config_to_amazonmq.py
+++ b/terraform/lambda/PostConfigToAmazonMQ/post_config_to_amazonmq.py
@@ -5,30 +5,34 @@ import urllib.error
 
 from urllib import request
 
+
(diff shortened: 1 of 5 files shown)


sengi:
Applied everywhere now; dropped the commit with the `moved` blocks.

PR #1830, opened 2024-05-07
Chat listen to links queue

kevindew:
Previously we didn't need to listen to this queue as chat didn't utilise any decisions based on the contents of RabbitMQ links. However the new features we're planning require this (checking history mode, parent document type) and thus I'm adding this.

This will result in a higher drastically higher quantity of content being ingested as the links queue is quite noisy.


AlexAvlonitis (approved):

kevindew:
Thanks @AgaDufrat 👍

PR #1833, opened 2024-07-02
Remove Pingdom IP rules for mirror buckets

theseanything:
We no longer probe the mirrors from Pingdom.


AgaDufrat (approved):
For posterity, has this check moved somewhere else or been replaced?

theseanything:
> For posterity, has this check moved somewhere else or been replaced?

Sorry - you it's been[ re-implemented in govuk-exporter](https://github.com/alphagov/govuk-exporter/pull/18)

PR #1837, opened 2024-07-22
Fix use of govuk_chat_retry_message-ttl variable

rtrinque:
Fix use of govuk_chat_retry_message-ttl variable


ianhowell-gds (approved):
LGTM

